PR #22, opened 2016-12-30
Sample dockerfile added

jcantosz:
Docker file using code from getting started guide
Signed-off-by: Jesse C. Antoszyk <jcantosz@us.ibm.com>

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -5,7 +5,6 @@ COPY tmp/startup.sh startup.sh
 
 #Install Packages
 RUN apt-get update -y \
-    && apt-get install -y software-properties-common python-software-properties apt-transport-https \
 
     # Install NodeJS
     && apt-get install -y curl \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,43 @@
+FROM ubuntu
+MAINTAINER Jesse Antoszyk <jesse@antoszyk.com>
+
+COPY tmp/startup.sh startup.sh
+
+#Install Packages
+RUN apt-get update -y \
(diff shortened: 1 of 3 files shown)


jcantosz:
Image builds can be automated via travis/GH/docker hub. Let me know if you need any help in this department

leominov:
Why are you using just last commit from default branch?

Alternative ways:

- reset to commit
- zip-archive of commit (no need git)
- release from release tab (no need git)

v1r3n:
How about adding Dynomite and Elaticssearch to the image?

jcantosz:
@v1r3n I was assuming running the sample server would make a large number of use cases simply testing the server. If you have a main.js that used elastic search + dynamite, I'm happy to make a docker image that uses that, otherwise it will just be bloat. 
I was going to move this to docker-compose for the conductor + DynomiteDB once I had something working on that. 

@leominov Thanks for the input. I'm not sure how this will fit into the pipeline. Ideally docker hub can be triggered to build an image on a release event (which would allow us to tag with release names as well). We could do that and have a default flow that pulls the latest release for people building locally. I need to look more into the GH APIs for that.

v1r3n:
@jcantosz I have created a server module and details on how to configure it and use it.

https://netflix.github.io/conductor/server/

It uses an embedded Jetty container for the server - you might want to update your Dockerfile with this instead of test-harness.

Server module is evolving, as I think it needs more fine grain controls (thread pool size etc.) for production readiness, feel free to suggest/send PR if you think it can be done better.

I would suggest also updating the documentation with instructions on how to use docker image (I have a section on the link above that can be updated).

Thanks

jcantosz:
Closed in favor of PR #37

PR #90, opened 2017-02-20
Added support for eureka client.

ajayaks:
Conductor server can be register as eureka client.


ajayaks:
Any reason this build got failed?

PR #133, opened 2017-03-22
Add Go Client (Worker) Implementation

cquon:


client/go/conductorworker.go
```diff
@@ -0,0 +1,79 @@
...
+
+func (c *ConductorWorker) Start(taskType string, executeFunction func(t *task.Task) (*task.TaskResult, error), wait bool) {
+	log.Println("Polling for task:", taskType, "with a:", c.PollingInterval, "(ms) polling interval with", c.ThreadCount, "goroutines for task execution, with workerid as", hostname)
+	for i := 1; i <= c.ThreadCount; i++ {
+		go c.PollAndExecute(taskType, executeFunction)
+	}
+
+	// wait infinitely while the go routines are running
+	if wait {
+		for {
```
v1r3n: It seems this might take up 100% cpu.  The blow might be better.

```go
select {}
```

Some more references:
http://blog.sgmansfield.com/2016/06/how-to-block-forever-in-go/

cquon: This is a good catch.  I verified spawning 100 goroutines for multiple task pollers with the infinite for loop and saw the 100% usage in top.  After changing to select this was no longer the case.

PR #143, opened 2017-03-30
handle uncaught throwable

raharsha:
This change is to handle unhandled exceptions thrown by the tasks. Using ListenableFuture , onFailure will update the result of the task.


client/src/main/java/com/netflix/conductor/client/task/WorkflowTaskCoordinator.java
```diff
@@ -22,16 +22,10 @@
...
-import java.util.concurrent.LinkedBlockingQueue;
-import java.util.concurrent.RejectedExecutionException;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.ThreadFactory;
-import java.util.concurrent.ThreadPoolExecutor;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicInteger;
 
+import com.google.common.util.concurrent.*;
```
tommack: It doesn't seem like you're doing anything with this that you can't do with a CompletableFuture if you don't want the Guava dependency.

client/src/main/java/com/netflix/conductor/client/task/WorkflowTaskCoordinator.java
```diff
@@ -297,32 +305,43 @@ private void pollForTask(Worker worker) {
...
-	
-	private void execute(Worker worker, Task task) {
+
+	private void handleException(Throwable t, TaskResult result, Worker worker, boolean updateTask, Task task) {
+		WorkflowTaskMetrics.executionException(worker.getTaskDefName(), t);
+		result.setStatus(TaskResult.Status.FAILED);
+		result.setReasonForIncompletion("Error while executing the task: " + t);
+		TaskExecLog execLog = result.getLog();
+		execLog.setError(t.getMessage());
+		for (StackTraceElement ste : t.getStackTrace()) {
```
tommack: You're not capturing the cause here, which is the most important part. I would just capture the stack trace as a string.

client/src/main/java/com/netflix/conductor/client/task/WorkflowTaskCoordinator.java
```diff
@@ -297,32 +305,43 @@ private void pollForTask(Worker worker) {
...
+            execLog.getErrorTrace().add(ste.toString());
+        }
+		updateWithRetry(updateRetryCount, task, result, worker);
+	}
+
+	private TaskResult execute(Worker worker, Task task) {
```
tommack: Why did you change this to return TaskResult? You're still handling the result within the body and you ignore it in the onSuccess above.

client/src/main/java/com/netflix/conductor/client/task/WorkflowTaskCoordinator.java
```diff
@@ -244,7 +240,7 @@ public Thread newThread(Runnable r) {
 				t.setName("workflow-worker-" + count.getAndIncrement());
```
tommack: If you're going to stick with a Guava dependency, see com.google.common.util.concurrent.ThreadFactoryBuilder.

client/src/main/java/com/netflix/conductor/client/task/WorkflowTaskCoordinator.java
```diff
@@ -285,8 +281,20 @@ private void pollForTask(Worker worker) {
 			List<Task> tasks = client.poll(taskType, worker.getIdentity(), worker.getPollCount(), worker.getLongPollTimeoutInMS());
 			sw.stop();
 			logger.debug("Polled {} and receivd {} tasks", worker.getTaskDefName(), tasks.size());
+
 			for(Task task : tasks) {
-				es.submit(()->execute(worker, task));	
+				ListenableFuture<TaskResult> future = es.submit(() -> execute(worker, task));
```
tommack: Since you're not doing anything with the future, this isn't any different than

```
es.submit(() -> {
    try {
        execute(worker, task);
    } catch (Exception e) {
        task.setStatus(Task.Status.FAILED);
        TaskResult result = new TaskResult(task);
        handleException(t, result, worker, true, task);
    }
});
```

client/src/main/java/com/netflix/conductor/client/task/WorkflowTaskCoordinator.java
v1r3n: `execLog.setError(t.getMessage());` should capture the cause.

tommack: No, that will return the message from the top most exception. Nothing about the cause chain will be reported.

client/src/main/java/com/netflix/conductor/client/task/WorkflowTaskCoordinator.java
v1r3n: I think the fork was done from a previous branch/release.  This is a void method in the current release.

@raharsha

client/src/main/java/com/netflix/conductor/client/task/WorkflowTaskCoordinator.java
raharsha: I didnt change this part, just refactored into a method for reuse.

Issue #178, opened 2017-04-21
Avoid aborting the entire workflow if some forked tasks are failing


picaron:
Hi, 

I am building a workflow that has a dynamic fork step of type HTTP followed by a join.

Currently, if any of the forked tasks fail, the entire workflow is aborted.

See example execution graph :

<img width="486" alt="workflow-graph-with-error" src="https://cloud.githubusercontent.com/assets/507577/25285836/9dbd10d8-2689-11e7-82b4-0bd764d2f45b.png">

I want to generate a report at the end of the workflow that will collect the result of each forked tasks.

I added a report task at the end of the workflow but if any of the forked tasks fail this last step never gets executed.

I tried marking the join step as optional but still no success.

Can you please advise on how I could implement the workflow I describe above?

Thanks and great framework by the way 👍

picaron:
Here is the workflow definition I'm using : 

```
{
  "updateTime": 1492789695762,
  "name": "cancel_all_subscriptions",
  "description": "Cancel all subscriptions for a given company",
  "version": 9,
  "tasks": [
    {
      "name": "find_active_subscriptions",
      "taskReferenceName": "find",
      "inputParameters": {
        "http_request": {
          "uri": "https://827e9f2f.ngrok.io/tasks/find_subscriptions_for_company",
          "method": "GET",
          "accept": "application/json",
          "body": {
            "companyId": "${workflow.input.companyId}"
          }
        }
      },
      "type": "HTTP",
      "startDelay": 0
    },
    {
      "name": "dynamic_fork",
      "taskReferenceName": "fork_cancel",
      "inputParameters": {
        "dynamicTasks": "${find.output.response.body.dynamicTasksJSON}",
        "dynamicTasksInput": "${find.output.response.body.dynamicTasksInputJSON}"
      },
      "type": "FORK_JOIN_DYNAMIC",
      "dynamicForkTasksParam": "dynamicTasks",
      "dynamicForkTasksInputParamName": "dynamicTasksInput",
      "startDelay": 0
    },
    {
      "name": "dynamic_join",
      "taskReferenceName": "join",
      "type": "JOIN",
      "startDelay": 0,
      "optional": true
    },
    {
      "name": "generate_report",
      "taskReferenceName": "report",
      "inputParameters": {
        "http_request": {
          "uri": "https://827e9f2f.ngrok.io/tasks/generate_report",
          "method": "POST",
          "body": {
            "workflowId": "${workflow.workflowId}"
          }
        }
      },
[...]

PR #179, opened 2017-04-21
Add optional support for HTTP tasks

picaron:
#### Description

If the HTTP task has the optional parameter set to true and the task fails, the status will be set to COMPLETED_WITH_ERRORS instead of FAILED allowing the workflow to continue.

See also issue #178 

#### Example use : 

```
  "http_request": {
    "uri": "http://host/path",
    "method": "POST",
    "optional": true,
    "body": {
      "var1": true
    }
  }
```

### DIFF
--- a/contribs/src/main/java/com/netflix/conductor/contribs/http/HttpTask.java
+++ b/contribs/src/main/java/com/netflix/conductor/contribs/http/HttpTask.java
@@ -114,6 +114,8 @@ public class HttpTask extends WorkflowSystemTask {
 			return;
 		}
 		
+		boolean isOptionalTask = input.getOptional() != null && input.getOptional();
+		
 		try {
 			
 			HttpResponse response = httpCall(input);
@@ -121,7 +123,7 @@ public class HttpTask extends WorkflowSystemTask {
 				task.setStatus(Status.COMPLETED);
 			} else {
 				task.setReasonForIncompletion(response.body.toString());
-				task.setStatus(Status.FAILED);
+				task.setStatus(isOptionalTask ? Status.COMPLETED_WITH_ERRORS : Status.FAILED);
 			}
 			if(response != null) {
 				task.getOutputData().put("response", response.asMap());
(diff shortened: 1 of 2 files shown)


v1r3n:
@picaron This may not be required.  Have you tried configuring the workflow with workflow task marked as optional?  We have this feature in a pre-release version 1.7.2+.

picaron:
@v1r3n yes I tried using the latest master but for HTTP task this doesn't work because it does not extend `WorkflowTask`

marioluan:
Wouldn't it be better to make HTTP task extend the WorflowTask instead?

v1r3n:
All the system tasks like HTTP extend from WorkflowSystemTask, which is what HTTP task extends too. 

To use the optional flag, you want to set workflow task as option in the workflow definition.  That should do.

picaron:
Well in my case the HTTP tasks are forked using a `FORK_JOIN_DYNAMIC` task and putting optional on it didn't work. I also tried putting on the `JOIN` and still I wasn't getting the desired behavior.

I had to do the changes in this PR to be able to get the desired behavior (see issue #178)

v1r3n:
@picaron seems you deleted the branch before the merges are done.  Although setting optional should work, I think this is still a good feature to have.

picaron:
@v1r3n actually I deleted the branch after you merge a recent PR because my changes were no longer compatible with the latest master, main reason being that the COMPLETED_WITH_ERRORS status was removed.

I would be happy to create a new PR.

Since this status `COMPLETED_WITH_ERRORS` no longer exists, how do to suggest to implement this?

Should I simply set the HTTP task to `COMPLETED` or is there another better approach?

v1r3n:
@picaron we recently starting implementing a feature where the tasks can be marked as optional.  There are some pending items there (like dynamic fork to support that), but once implemented, it should allow HTTP tasks to ignore failures.  ETA is sometime this week.

picaron:
ok that's perfect then, I can wait for that, thanks @v1r3n 👍

picaron:
@v1r3n any updates on supporting optional HTTP tasks?

PR #186, opened 2017-04-27
Add support for Elasticsearch 5.3.0

cquon:
Couple of Notes:

1. Technically Elasticsearch says Embedded Elasticsearch is not supported:
https://www.elastic.co/blog/elasticsearch-the-server

I went with an approach similar to
https://github.com/elastic/elasticsearch-hadoop/blob/fefcf8b191d287aca93a04144c67b803c6c81db5/mr/src/itest/java/org/elasticsearch/hadoop/EsEmbeddedServer.java
for the Embedded Elasticsearch that is used for the in memory persistence.

2. The new elasticsearch is preinstalled with X-Pack:
https://www.elastic.co/guide/en/elasticsearch/reference/current/docker.html

Therefore the transport client is having connection issues since there is authentication (default username: elastic/ default password: changeme)

For a quick and easy bringup with the docker-compose, I have elected to disable this security feature in the docker-compose.yml for now:

```
 environment:
     - xpack.security.enabled=false
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ jdk:
 - oraclejdk8
 install: true
 script: "./buildViaTravis.sh"
+git:
+  depth: 150
 addons:
   apt:
     packages:
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ com.netflix.conductor.dao.RedisESWorkflowModule
 
 * The default persistence used is [Dynomite](https://github.com/Netflix/dynomite)
 * For queues, we are relying on [dyno-queues](https://github.com/Netflix/dyno-queues)
-* The indexing backend is [Elasticsearch](https://www.elastic.co/) (2.+)
+* The indexing backend is [Elasticsearch](https://www.elastic.co/) (5.3.0)
 
 ## Other Requirements
 * JDK 1.8+
--- a/build.gradle
+++ b/build.gradle
@@ -2,6 +2,9 @@ buildscript {
 	
 	repositories {
         jcenter()
+        maven {
+            url "https://artifacts.elastic.co/maven"
+        }
     }
     
     dependencies {
@@ -33,6 +36,9 @@ subprojects {
 
     repositories {
         jcenter()
+        maven {
+            url "https://artifacts.elastic.co/maven"
+        }
     }
 
     dependencies {
--- a/build.gradle
+++ b/build.gradle
@@ -43,7 +43,7 @@ subprojects {
     group = "com.netflix.${githubProjectName}"
     
     tasks.withType(Test)  {
-        maxParallelForks = 1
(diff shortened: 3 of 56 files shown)


v1r3n:
@cquon will this change support ES 2.x?

cquon:
@v1r3n 

No I had tested with ES2.x and it throws exception:
```
java.lang.IllegalStateException: Received message from unsupported version: [5.0.0] minimal compatible version is: [2.0.0]
```

Documentation says it must be on same major version:
https://www.elastic.co/guide/en/elasticsearch/client/java-api/current/client.html

It would require a full cluster restart:
https://www.elastic.co/guide/en/elasticsearch/reference/5.3/setup-upgrade.html

v1r3n:
@cquon thanks.  This could be a problem because we have few users including ourselves using it on 2.x and are not yet ready to move to 5.x.

Is there a way to create a separate module for ES 5 and put this class in there?  That would allow folks to pick the right version.  Also I would suggest renaming the class to reflect that.

Thoughts?

cquon:
@v1r3n Yes that definitely makes sense.

When you say different module do you mean a different top level project?

The legacy support for elasticsearch 2.4 has a dependency on "org.elasticsearch:elasticsearch:2.+"

while elasticsearch 5.3.0 has a dependency on "org.elasticsearch:elasticsearch:5.3.0" and "org.elasticsearch.client:transport:5.3.0".

Do you recommend under top level conductor directory having a elasticsearch-client and elasticsearch-client5 project for example, each with their own dependencies and then the build.gradle for redis-persistence use a specific one?

cquon:
@v1r3n 
Another option is to use the new Java REST client for elasticsearch to support one of the versions and the existing Transport client for others.  It appears the REST client has more flexibility across versions, but performance may take a small hit.  I am not sure if you have a preference for one going forward it looks like the REST client is relatively new since elasticsearch 5.0.

Github:
https://github.com/elastic/elasticsearch/pull/18735

Documentation:
https://www.elastic.co/guide/en/elasticsearch/client/java-rest/current/index.html

Performance Benchmarking:
https://www.elastic.co/guide/en/elasticsearch/client/java-rest/current/index.html

v1r3n:
+1 to using the HTTP REST client.  I think performance hit even if some, might be acceptable as it is not in the critical path.

v1r3n:
@cquon any chance you can create a separate module and put this there?  The reason is we have users using it over 2.x and there are no plans to push them to 5 for a while, so it will be required to keep both versions available till then.

cquon:
Sure will look into doing that first before moving to rest client.

v1r3n:
Closing this PR - ES 5 should be available in this PR https://github.com/Netflix/conductor/pull/209

PR #219, opened 2017-06-13
Python3

guyeng0:
Added support for a python 3.x client.


v1r3n:
Thanks for the PR.  I will merge it after the current RC is certified.

PR #224, opened 2017-06-16
Add OAuth 1.0 authentication support to HTTP tasks

picaron:
This small PR adds support for calling HTTP endpoints secured with OAuth 1.0

If the OAuth parameters (consumer key and consumer secret) are present in the input parameters for the HTTP tasks, those will be used for authentication against the endpoint.

We tested this internally successfully with our OAuth endpoints.

### DIFF
--- a/contribs/build.gradle
+++ b/contribs/build.gradle
@@ -5,13 +5,13 @@ dependencies {
 
 	compile 'com.amazonaws:aws-java-sdk-sqs:latest.release'
 	compile ('com.google.inject:guice:4.1+') { force= true}
-	
+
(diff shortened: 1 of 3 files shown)


picaron:
@v1r3n could you please have a look at this PR ?

We really need HTTP tasks to support OAuth authentication

Could you consider merging this PR ? Ideally we'd like to merge this feature here to avoid maintaining a separate dependency. I think this could be useful to others as well.

v1r3n:
@picaron thanks for the the PR.  We will merge this to dev sometime next week once the current release candidate is certified.

PR #227, opened 2017-06-19
Add new system task type JSON_TRANSFORM

picaron:
# Description

This new system task type allows transforming a JSON input to another JSON structure using a query expression.

The library jq for Jackson is used for processing (https://github.com/eiiches/jackson-jq). It is a Java implementation of the excellent jq command-line processor (https://stedolan.github.io/jq/)

# Why?

Many times, we want to use the output of existing REST HTTP endpoints to launch dynamic fork tasks and we would like to avoid having to modify those existing endpoints.

By applying a JSON transformation on the output of existing services, we can then directly use that transformed json structure to the format expected by Conductor. 

# Example use :

**Input (example task within a workflow definition)**

The json data to transform is passed in the "inputJSON" attribute and the query expression is specified in the "queryExpression" attribute.

```
...
      {
        "type": "JSON_TRANSFORM",
        "name": "transform",
        "taskReferenceName": "step1",
        "inputParameters": {
          "inputJSON": [
            {
              "manufacturer": "Porsche",
              "model": "911",
              "price": 135000,
              "wiki": "http://en.wikipedia.org/wiki/Porsche_997",
              "img": "2004_Porsche_911_Carrera_type_997.jpg"
            },
            {
              "manufacturer": "Nissan",
              "model": "GT-R",
              "price": 80000,
              "wiki": "http://en.wikipedia.org/wiki/Nissan_Gt-r",
              "img": "250px-Nissan_GT-R.jpg"
            },
            {
              "manufacturer": "BMW",
              "model": "M3",
              "price": 60500,
              "wiki": "http://en.wikipedia.org/wiki/Bmw_m3",
              "img": "250px-BMW_M3_E92.jpg"
            },
            {
              "manufacturer": "Audi",
              "model": "S5",
              "price": 53000,
              "wiki": "http://en.wikipedia.org/wiki/Audi_S5#Audi_S5",
              "img": "250px-Audi_S5.jpg"
            },
            {
              "manufacturer": "Audi",
              "model": "TT",
              "price": 40000,
              "wiki": "http://en.wikipedia.org/wiki/Audi_TT",
              "img": "250px-2007_Audi_TT_Coupe.jpg"
            }
          ],
          "queryExpression": ".[] |= { (.manufacturer + \" \" + .model): {price: .price, image: .img} }"
        }
      }
...
```

**Output**

The output is a json object with a key "result" and with the result of the transformation for its value.

```
{
   "result": [
      {
         "Porsche 911": {
            "image": "2004_Porsche_911_Carrera_type_997.jpg",
            "price": "135000"
         }
      },
      {
         "Nissan GT-R": {
            "image": "250px-Nissan_GT-R.jpg",
            "price": "80000"
         }
      },
      {
         "BMW M3": {
            "image": "250px-BMW_M3_E92.jpg",
            "price": "60500"
         }
      },
      {
         "Audi S5": {
            "image": "250px-Audi_S5.jpg",
            "price": "53000"
         }
      },
      {
         "Audi TT": {
            "image": "250px-2007_Audi_TT_Coupe.jpg",
            "price": "40000"
         }
      }
   ]
}
```


picaron:
Hi @v1r3n 

Another missing key component for implementing our workflows using our existing REST endpoints was the ability to transform a json structure into another one (one that in the expected format for some conductor system tasks).

We were initially considering deploying a new service within our infra just for that but we thought it might be more useful and efficient as a system task within Conductor and we thought that other might benefit from this capability as well.

So I am submitting this PR for your approval, all your comments are very welcome.

Thanks in advance 😃

v1r3n:
@picaron Thanks for the PR.  We actually had this in the plans - so thanks for adding the feature.  Appreciate it.  I will merge this PR once the current release is certified.

tommack:
This is close to what I've been using. I'd suggest the following changes:

1. This shouldn't be called a `JsonTransform`. There are lots of ways of transforming JSON, this is very particular to jq.
2. I would just use the whole inputData as the input. Additionally, I made an optional argument to instead use the entire task as the input to the transform, thus allowing access to things like the task ID.
3. Cache the compiled queries (e.g. a Guava cache). 
4. Looks like a NullPointerException: `e.getCause().getMessage()`
5. A list of one is still a list. The conversion of the result makes the output challenging to consume if you don't know the exact size ahead of time. I'd suggest something like the following.

```
            if (result == null) {
                outputDataMap.put("result", null);
                outputDataMap.put("resultList", null);
            } else if (result.isEmpty()) {
                outputDataMap.put("result", null);
                outputDataMap.put("resultList", result);
            } else {
                outputDataMap.put("result", result.get(0));
                outputDataMap.put("resultList", result);
            }

```

picaron:
Thanks for the pertinent comments @tommack 

Regarding your suggestions: 

1. You are right, maybe we could call this a JsonJqTransform ?
2. I'm not sure about exposing the entire input data. That requires more complex processing in the query expression. A dedicated attribute gives some control of the data that gets passed to the processor.
3. Good idea, I'll do that
4. I'll fix that
5. The issue I had is that using the JQ library, the result is always wrapped in a list even if the result of the query does not return a list. I think you need to know the type of result your query expression will be generating anyhow. We could implement something like you suggest but I feel the output will be more convoluted

v1r3n:
+1 for JsonJqTransform

tommack:
1. Sounds good.
2. You can simply change `JsonNode inputJson = om.valueToTree(inputData);` to `JsonNode inputJson = om.valueToTree(task.getInputData()).
3. 
4. 
5. Like I said, a list of one is still a list. If I write an expression that returns a list, but can't guarantee that it will never be of size one, I wouldn't want to have to handle that on the consumer side where I'd have to write code that handled a list and a value. With the code I wrote, if I know my expression will always return a single result: great, I can use the `result` parameter. However, if I know my expression returns a list, then I can use the `resultList` parameter and be assured that I'll always get a list when I receive it.

picaron:
2. Maybe I'm missing something but doing that, wouldn't the processor also receive the queryExpression parameter? I think its preferable to have some control on the data interpreted by the query no?

5. The issue is that if you write some expression that does not return a list, somehow the Java jq lib still always wraps it in a list. For example, I was debugging my expression using the online tool (https://jqplay.org/) and it returns me the object I expect. But when the jq java library processes the result, it always wrap the result in a list. I think it's because you can optionally pass a list of transformations to the lib. That's why I need to "unwrap" that result. If you write an expression that returns a list, it will stay a list (because the jq lib will put that list in a list 😱 )

tommack:
Regarding the second point, to me it doesn't matter. The first part of your expression can always be `.someFieldThatHasWhatIWant`.

Regarding the fifth point, it isn't accurate to say that JQ is _wrapping_ everything in a list. The output of any JQ expression has a 0-n arity. A List just happens to be a very good Java type for representing this.

Consider an input that is an array and the following three JQ expressions.

### `.`
The Java output will be a List with one element, where that element is an array.

### `.[]`
The Java output will be a List with an element per item that was in the input array.

### `.[0]`
The Java output will be a List with one element that is the first element from the input.

In the first and third case, having the result List "unwrapped" is very convenient because I know based on the expression that only one response is expected. But in the second case, I'm actually expecting multiple results. If you try to unwrap this, you will corrupt it in the size=1 case. This is why the code I offered originally works well: you get the best of both worlds with no compromises.

picaron:
Mmm you are right for the 5th point, I didn't consider the second case which will corrupt the result. I guess the code you suggested is the best approach 😄 

For the second point, I guess you are right as well. 

Thanks for taking the time to explain, I didn't see all the whys and wherefores.

tommack:
Glad I could help.

patrickbadley:
@picaron Can you provide an example usage of this task?  I tried the example from your initial issue, but the task stays in SCHEDULED status indefinitely.  Thanks!  I apologize if this is documented elsewhere.

patrickbadley:
I worked through utilizing this task type and thought I'd share my example for reference:
`{
  "name": "Compensate_Place_Order_Workflow",
  "description": "Compensate_Place_Order_Workflow",
  "version": 3,
  "tasks": [
    {
      "name": "Get_Workflow_Details",
      "taskReferenceName": "Get_Workflow_Details",
      "inputParameters": {
        "http_request": {
          "uri": "http://localhost:8080/api/workflow/${workflow.input.workflowId}",
          "method": "GET"
        }
      },
      "type": "HTTP",
      "startDelay": 0
    },
    {
      "type": "JSON_JQ_TRANSFORM",
      "name": "transform",
      "taskReferenceName": "CompensatingRequestsTasks",
      "inputParameters": {
        "requests": "${Get_Workflow_Details.output.response.body.tasks[?(@.status == 'COMPLETED' && @.taskType == 'HTTP')]}",
        "queryExpression": ".requests[] |= { name: \"Compensate_HTTP_Task\", taskReferenceName: .referenceTaskName + \"_Compensate\", type: .taskType, key: .referenceTaskName + \"_Compensate\", value: { http_request: { uri: .inputData.http_request.uri + \"/compensate\", method: .inputData.http_request.method, body: .inputData.http_request.body } } }"
      }
    },
    {
      "type": "JSON_JQ_TRANSFORM",
      "name": "transform",
      "taskReferenceName": "CompensatingRequestsInputParameters",
      "inputParameters": {
        "requests": "${CompensatingRequestsTasks.output.result.requests}",
        "queryExpression": ".requests |= from_entries"
      }
    },
    {
      "inputParameters": {
        "dynamicTasks": "${CompensatingRequestsTasks.output.result.requests}",
        "dynamicTasksInput": "${CompensatingRequestsInputParameters.output.result.requests}"
      },
      "type": "FORK_JOIN_DYNAMIC",
      "dynamicForkTasksParam": "dynamicTasks",
      "dynamicForkTasksInputParamName": "dynamicTasksInput",
      "taskReferenceName": "dynamicJoin"
    },
    {
      "name": "join",
      "taskReferenceName": "join",
      "type": "JOIN"
    }
  ],
  "outputParameters": {
    "orderId": "${workflow.input.OrderId}",
    "joinOutput": "${dynamicJoin.output}"
  },
  "restartable": true,
  "schemaVersion": 2
}`

wknd-deevpos:
This is a nice feature... is there any documentation for the usage? One of our API returns following JSON
```
{
    "hits": [
        {
            "_index": "mes$i$v2$lid_infor.daf.daf$mds_file$1.0$1561444341392",
            "_type": "mds_file",
            "_source": {
                "LASTCHANGEDTS": "2018-11-06T10:01:10Z"
            },
            "_id": "3_ax1",
            "_score": 4.491742
        },
        {
            "_index": "mes$i$v2$lid_infor.daf.daf$mds_file$1.0$1567159125621",
            "_type": "mds_file",
            "_source": {
                "LASTCHANGEDTS": "2019-04-25T07:07:14.223Z"
            },
            "_id": "3_ax2",
            "_score": 2.2275105
        }
    ]
}
```
I used following task definition, but the workflow results in error. I tested JQ on jqplay.org and they work as expected.
```
{
    "type": "JSON_JQ_TRANSFORM",
    "name": "transform",
    "taskReferenceName": "TransformSerchResult",
    "inputParameters": {
        "hits": "${search.output.response.body}",
        "queryExpression": ".hits[] |=  { id: ._id, index: ._index, source: {created:._source.LASTCHANGEDTS} }"
    }
}
```
Any suggestions ?

PR #246, opened 2017-06-27
adds information about how to run it on docker compose

leandromoreira:
I'm reading/following the quickstart documentation (more specifically the intro) and I saw that it was possible to use `docker-compose` and I thought it'd be good to have this information there.

About the `clone` change, I was following the quickstart guide and then it raised a silly error.

> $ git clone git@github.com/Netflix/conductor.git
> fatal: repository 'git@github.com/Netflix/conductor.git' does not exist


leandromoreira:
:( I don't see how or why this change would break the CI and I also can't rerun the Travis.

PR #248, opened 2017-06-28
simplifies docker-compose usage

leandromoreira:


leandromoreira:
It'd be nice if we could pass environment variable mostly for all the url instead of hard coding them at `properties` file, for instance, the elastic search url api seems to be provided by `workflow.elasticsearch.url`.

PR #265, opened 2017-07-13
HTTP task : add HTTP reason phrase in the response metadata

picaron:
When debugging failing HTTP tasks, it's helpful to include the HTTP reason phrase instead of exposing only the status code. This is specially useful for non common status codes.

Here is an example output for an HTTP task returning a 424 (Failed Dependency) response code:

```
{
   "response": {
      "headers": {
         ...
      },
      "reasonPhrase": "Failed Dependency",
      "body": {
         "code": "EXTERNAL_ERROR",
         "message": "An error occurred while creating or modifying your account"
      },
      "statusCode": 424
   }
}
```


picaron:
@v1r3n can you please review this very small change?

it would be helpful for debugging and also would allow to do some processing based on the reason phrase.

PR #279, opened 2017-07-20
Option to specify worker name prefix and support for AWS Elasticache Redis

clari-akhilesh:


server/src/main/java/com/netflix/conductor/server/ConductorServer.java
```diff
@@ -167,6 +170,15 @@ public HostToken getTokenForHost(Host host, Set<Host> activeHosts) {
 			}
 			logger.info("Starting conductor server using in memory data store");
 			break;
+
+		case redis_cluster:
```
v1r3n: @clari-akhilesh can you also update the documentation on how to use the redis cluster and maybe samples?

Issue #280, opened 2017-07-21
Migrate docker images to Alpine Linux


toddschilling:
"Alpine Linux is a Linux distribution built around musl libc and BusyBox. The image is only 5 MB in size and has access to a package repository that is much more complete than other BusyBox based images. This makes Alpine Linux a great image base for utilities and even production applications. Read more about Alpine Linux here and you can see how their mantra fits in right at home with Docker images."

- [Alpine - Docker Hub](https://hub.docker.com/_/alpine/)

Both Java and Node.js have base image variants for Alpine Linux which result in final runtime images which are 49%-85% smaller than the current conductor ui and server images respectively.

- [Java 8 Alpine - GitHub](https://github.com/docker-library/openjdk/blob/9a0822673dffd3e5ba66f18a8547aa60faed6d08/8-jre/alpine/Dockerfile)
- [Node.js Alpine - GitHub](https://github.com/nodejs/docker-node/blob/4f2354e558a0cd93ae53278f83f5f0a35af23be0/8.2/alpine/Dockerfile)

Objectives
-----------
* Migrate server build to a 2-step process where the server is built using the java:8-jdk and is then copied to a java:8-jre-alpine runtime image.
* Migrate ui build to a node:alpine image.
* Change startup scripts to run ui and server processes in the foreground.
* Change startup scripts to remove the logging redirect.  Docker expects log messages to be sent directly to stdout and stderr.
* Update server configuration code to always allow configuration settings to be overridden by environment variables.

Notes
------
* Leave ServerAndUI image as-is for local execution only.  Consider deprecating.
* Consolidate readme documentation to the single readme file in the docker folder.

toddschilling:
I have this working on my personal fork.  The new image sizes are as follows:
```
                        before      after
                        -------     -----
conductor:server        995 MB      153 MB
conductor:ui            914 MB      463 MB
```

PR #281, opened 2017-07-21
Migrate docker images to Alpine Linux

toddschilling:
Pull Request for Issue #280 

Objectives
-----------
* Migrate server build to a 2-step process where the server is built using the java:8-jdk and is then copied to a java:8-jre-alpine runtime image.
* Migrate ui build to a node:alpine image.
* Run ui and api in the foreground
* Allow environment variables to be override configuration file settings.


toddschilling:
Server and UI containers have been verified locally using docker-compose and on AWS using HashiCorp Nomad.

core/src/main/java/com/netflix/conductor/core/execution/ParametersUtils.java
```diff
@@ -223,9 +223,9 @@ private String getSystemParametersValue(String sysParam, String taskId){
 		if("CPEWF_TASK_ID".equals(sysParam)) {
 			return taskId;
 		}
-		String value = System.getProperty(sysParam);
```
toddschilling: This is enforcing the order of precedence as 1. environment variable and then 2. configuration file.  This establishes the configuration file as defaults and environment variables as overrides.

docker/README.md
```diff
@@ -1,35 +1,47 @@
 # Docker
```
toddschilling: I was thinking that the individual readme files at docker/server and docker/ui could be consolidated into the readme under docker.

docker/README.md
```diff
@@ -1,35 +1,47 @@
...
-To ensure images are stopped do:
- - `docker-compose down`
+`docker build --no-cache -t conductor:ui -f docker/ui/Dockerfile .`
 
-## Running in Interactive Mode
-In interactive mode the default startup script for the container do not run
- - `docker run -p 8080:8080 -t -i conductor:server -`
- - `docker run -p 5000:5000 -t -i conductor:ui -`
+## Running Conductor
+
```
toddschilling: I'm thinking that the appropriate way to run Conductor locally is to spin up the entire product service stack using compose.  As users become more savvy with Docker, they can always mimic the behavior of compose with individual commands, but compose is still an easier starting point - especially since you provided the compose file.

docker/server/Dockerfile
```diff
@@ -1,42 +1,20 @@
...
+COPY ./server/build/libs/conductor-server-*-all.jar /app/libs
 
+# Copy the files for the server into the app folders
+RUN chmod +x /app/startup.sh
 
 EXPOSE 8080
-
-CMD ["/app/startup.sh"]
-ENTRYPOINT ["/bin/bash"]
+CMD [ "/app/startup.sh" ]
```
toddschilling: `startup.sh` will run the service process in the foreground now.

docker/server/README.md
```diff
@@ -1,10 +0,0 @@
-# Docker
```
toddschilling: Consolidated this file into `docker/README.md`.

docker/server/bin/startup.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
```
toddschilling: Alpine Linux doesn't come with bash.  Easy enough to rectify.

docker/server/bin/startup.sh
```diff
@@ -11,11 +11,13 @@ export config_file=
...
-    echo "Using '$CONFIG_PROP'";
     export config_file=/app/config/$CONFIG_PROP
 fi
 
-nohup java -jar conductor-server-*-all.jar $config_file 1>&2 > /app/logs/server.log
+# Log the configuration settings as defaults
+echo "Starting conductor server with the following defaults: $(cat $config_file | grep = | grep -v '#' | sed ':a;N;$!ba;s/\n/ /g')"
+
+# Run java in the foreground and stream messages directly to stdout
+java -jar conductor-server-*-all.jar $config_file
```
toddschilling: Run in the foreground by removing `nohup`.  Remove redirect so messages go to stdout and stderr.

docker/serverAndUI/Dockerfile
```diff
@@ -51,4 +51,4 @@ RUN git clone https://github.com/Netflix/conductor.git \
 EXPOSE 5000 8080
```
toddschilling: I didn't touch `docker/serverAndUI`.  I think you could probably deprecate it in favor of asking noobs to just use docker-compose to run the stack locally.

docker/ui/Dockerfile
```diff
@@ -1,38 +1,41 @@
...
+  build-base \
+  libstdc++ \
+  gcc \
+  abuild \
+  binutils \
+  nasm \
+  libpng \
+  libpng-dev \
+  libjpeg-turbo \
+  libjpeg-turbo-dev
```
toddschilling: Some of these might be redundant.  We might be able to make this image smaller yet by installing only what is necessary to run the ui.

docker/ui/README.md
```diff
@@ -1,10 +0,0 @@
-# Docker
```
toddschilling: Consolidated this file into `docker/README.md`.

docker/ui/bin/startup.sh
```diff
@@ -12,4 +12,5 @@ if [ -z "$WF_SERVER" ];
     echo "using Conductor API server from '$WF_SERVER'"
 fi
 
-nohup node server.js 1>&2 > /app/logs/ui.log
\ No newline at end of file
+# Run server.js in the foreground, sending messages to stdout
+node server.js
```
toddschilling: Same as server.  Removed `nohup` to run in the foreground and removed redirects to keep messages going to stdout and stderr.

server/src/main/java/com/netflix/conductor/server/ConductorConfig.java
```diff
@@ -95,24 +95,37 @@ public String getAvailabilityZone() {
...
-		int value  = defaultValue;
+		String val = getProperty(key, Integer.toString(defaultValue));
 		try{
-			value = Integer.parseInt(val);
-		}catch(Exception e){}
-		return value;
+			defaultValue = Integer.parseInt(val);
+		}catch(NumberFormatException e){}
+		return defaultValue;
 	}
```
toddschilling: There might be a cleaner way to do this.  Java is not my first language.

server/src/main/java/com/netflix/conductor/server/ConductorConfig.java
```diff
@@ -95,24 +95,37 @@ public String getAvailabilityZone() {
...
 
 	@Override
 	public String getProperty(String key, String defaultValue) {
-		return Optional.ofNullable(System.getProperty(key)).orElse(defaultValue);
+
+		String val = null;
+		try{
+			val = System.getenv(key.replace('.','_'));
```
toddschilling: Some docker orchestrators will replace `.` with `_`.  Nomad for example.   Maybe we could check both?  For now the assumption is that config values look like `component.subsection.key` and environment variables take the form `component_subsection_key`.

server/src/main/java/com/netflix/conductor/server/ConductorConfig.java
```diff
@@ -95,24 +95,37 @@ public String getAvailabilityZone() {
...
 
 	@Override
 	public Map<String, Object> getAll() {
 		Map<String, Object> map = new HashMap<>();
 		Properties props = System.getProperties();
-		props.entrySet().forEach(entry -> map.put(entry.getKey().toString(), entry.getValue()));
+		props.entrySet().forEach(entry -> {
+			String key = entry.getKey().toString();
+			String val = entry.getValue().toString();
+			map.put(key, getProperty(key, val));
```
toddschilling: Just go through our `getProperty` method so that the environment variable precedence is used.

server/src/main/java/com/netflix/conductor/server/ConductorServer.java
```diff
@@ -156,10 +156,10 @@ public HostToken getTokenForHost(Host host, Set<Host> activeHosts) {
 			jedis = new JedisMock();
 			try {
 				EmbeddedElasticSearch.start();
-				if(System.getProperty("workflow.elasticsearch.url") == null) {
+				if(cc.getProperty("workflow.elasticsearch.url", null) == null) {
```
toddschilling: Making sure that we're going through ConductorConfig to get configuration values so that environment variables are considered.

euforia:
👍

docker/docker-compose.yml
```diff
@@ -3,7 +3,16 @@ version: '2'
...
-      - CONFIG_PROP=config.properties
+      - db=dynomite
+      - workflow_dynomite_cluster_hosts=dyno1:8102:us-east-1c
+      - workflow_dynomite_cluster_name=dyno1
+      - workflow_namespace_prefix=conductor
+      - workflow_namespace_queue_prefix=conductor_queues
+      - queues_dynomite_threads=10
+      - queues_dynomite_nonQuorum_port=22122
+      - workflow_elasticsearch_url=es:9300
+      - workflow_elasticsearch_index_name=conductor
```
neojin: From my understanding these are `ConductorConfig` properties that are now being exposed as environment variables. This might be out of scope for this PR, but it seems unusual that env vars are lowecased. I have almost always seen them in caps.

I wonder if a better pattern for this is:

1. Expose the necessary configuration here as ENV vars
2. Pass and set these vars as ConductorConfig properties. For example - the env var `WORKFLOW_ELASTICSEARCH_INDEX_NAME` would map to CC's `workflow.elasticsearch.index.name`

Java is not my native language so I may be totally out of line here.

docker/serverAndUI/Dockerfile
```diff
@@ -51,4 +51,4 @@ RUN git clone https://github.com/Netflix/conductor.git \
...
 
 CMD ["/app/startup.sh"]
-ENTRYPOINT ["/bin/bash"]
+ENTRYPOINT ["/bin/bash"]
```
neojin: super nitpick: there should be a newline at the end of this line (hence github adding the symbol at the end)

toddschilling: super nitpicks are welcome.  my intent was to rollback this change completely since i didn't bother to migrate the server-and-ui image to alpine.  i'll clean that up.

docker/docker-compose.yml
toddschilling: I'm not sure I understand.  I get the UPPERCASE part.  Are you saying that, rather than checking the environment variable first, we simply overwrite the values in the actual configuration file at startup?  I think some of the Elastic.co products behave that way.  For example, if you run `kibana:5.2` and set the `ELASTICSEARCH_URL` environment variable, I'm pretty sure that it actually writes that value into the kibana.yml as `elasticsearch.url`.

neojin: >  if you run kibana:5.2 and set the ELASTICSEARCH_URL environment variable, I'm pretty sure that it actually writes that value into the kibana.yml as elasticsearch.url.

Yep this is exactly what I mean. I think, put simpler is:

`UPPERCASE_AND_UNDERSCORED_ENV_VAR` should write to `some.internally.used.property`

If this is the case here, feel free to ignore. I might be misunderstanding something with docker-compose and `environment`

core/src/main/java/com/netflix/conductor/core/execution/ParametersUtils.java
```diff
@@ -223,9 +223,9 @@ private String getSystemParametersValue(String sysParam, String taskId){
...
+		String value = System.getenv(sysParam);
 		if(value == null) {
-			value = System.getenv(sysParam);
+			value = System.getProperty(sysParam);
 		}
```
toddschilling: This is enforcing the order of precedence as 1. environment variable and then 2. configuration file. This establishes the configuration file as defaults and environment variables as overrides.

toddschilling:
I'm going to resubmit this with fewer changes.

PR #309, opened 2017-08-22
Fix : Cannot override system properties when launching the conductor server using the gradle command-line

picaron:
Currently, if you try to override system properties when launching the conductor server using the gradle command-line, those are not taken into consideration.

For example, if I run the following command (from the server subdirectory) : 

`../gradlew server -DloadSample=false -Dport=8888`

The server starts on port 8080 instead of the specified one and the loadSample=false is ignored as well.

This PR adds a missing line in the gradle build to allow overrides from the command-line

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -76,6 +76,7 @@ build.dependsOn('shadowJar')
 task server(type: JavaExec) {
   systemProperty 'workflow.elasticsearch.url', 'localhost:9300'
   systemProperty 'loadSample', 'true'
+  systemProperties System.properties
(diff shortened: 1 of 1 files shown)


picaron:
Hi @v1r3n

Could you please review this very small PR ?

it is a bit inconvenient having to update the `build.gradle` all the time when switching branches.

v1r3n:
@picaron can you send this against the `dev` branch?

picaron:
oh yes, my mistake

here is the new PR : https://github.com/Netflix/conductor/pull/318

PR #312, opened 2017-08-29
polling tasks only when worker is not fully occupied

dophinlife:
Without this PR, this worker keeps polling tasks even though it is currently fully occupied, the consequence is that:
                a). even if we start another instance (B) of worker to help, it may not be able to get any tasks, it would be idle even though A has tasks on the queue waiting to be executed
                b). if the max size of the queue is reached, tasks would be dropped as exception happens

### DIFF
--- a/client/src/main/java/com/netflix/conductor/client/task/WorkflowTaskCoordinator.java
+++ b/client/src/main/java/com/netflix/conductor/client/task/WorkflowTaskCoordinator.java
@@ -23,6 +23,7 @@ import java.util.List;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicInteger;
 
+import com.google.common.collect.Lists;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


v1r3n:
How about not polling for tasks if the workers are busy?  So when the queue is full, the polling should be paused.

dophinlife:
Do you mean to poll the task until all the threads in the thread pool are occupied? That's better.

dophinlife:
@v1r3n  any comments?

client/src/main/java/com/netflix/conductor/client/task/WorkflowTaskCoordinator.java
```diff
@@ -275,11 +269,17 @@ private void pollForTask(Worker worker) {
 		try{
 			
 			String taskType = worker.getTaskDefName();
-			Stopwatch sw = WorkflowTaskMetrics.pollTimer(worker.getTaskDefName());
-			List<Task> tasks = client.poll(taskType, domain, worker.getIdentity(), worker.getPollCount(), worker.getLongPollTimeoutInMS());
-			sw.stop();
-			logger.debug("Polled {}, for domain {} and receivd {} tasks", worker.getTaskDefName(), domain, tasks.size());
-			for(Task task : tasks) {
+            List<Task> tasks = Lists.newArrayList();
```
v1r3n: How about using logic similar to https://github.com/Netflix/conductor/blob/master/core/src/main/java/com/netflix/conductor/core/execution/tasks/SystemTaskWorkerCoordinator.java#L137 ?

Currently, the internal queue is not stored as a class variable, but it can be and then its size can be used to find out if the queue is full and if so, do not poll.

dophinlife:
Thanks for your suggestion, I have made the modification accordingly, please help me review again, thanks. @v1r3n

Issue #307, opened 2017-08-22
Add persistence support for a MySQL backend


picaron:
Hi !

We require using our existing MySQL infrastructure as the backend for our conductor instance.

We also would like to avoid to fork our own branch of the project so we are considering submitting a PR in the `contribs` package.

I would just like to make sure that we will have your approval on getting this feature merged so we don't waste our efforts for nothing.

v1r3n:
@picaron I would suggest create a separate module similar to es5-persistence so it can be added in the package build based on the requirements.

picaron:
ok understood, a separate `mysql-persistence` makes more sense indeed

v1r3n:
how do you plan on adding the support for queues?  currently, the queues are implemented using the redis based queue recipe - will you implement the equivalent using mysql?

picaron:
I see. We are trying to figure out how to do this.

We want to use only one backend and re-use our existing MySQL cluster so yes we are considering implementing everything on top of MySQL (except maybe we would reuse the embedded Elasticsearch if the index can be reconstructed every time the app starts). Do you see any major issues in trying to do this?

I'm guessing we are not the only ones who would benefit from a lighter setup using only MySQL ...

Another alternative would be to use RabbitMQ for the queuing, which is already deployed in our environment but ideally we'd like to keep it as simple as possible and use only one backend.

picaron:
Also, I was looking at the `ConductorServer` bootstrap, it seems we will need to add some logic in there to support the new mysql backend right?

v1r3n:
You will have to change the server module to bind the interfaces to your implementation.
For the queueing either you can build a lightweight queuing support using MySQL by implementing [QueueDAO.java](https://github.com/Netflix/conductor/blob/master/core/src/main/java/com/netflix/conductor/dao/QueueDAO.java) or supporting RabbitMQ is also an option - but then I would go with a separate module for that providing flexibility on queues.

+1 to supporting MySQL

PR #315, opened 2017-09-01
Add MySQL persistence support

picaron:
As discussed in https://github.com/Netflix/conductor/issues/307,  this PR is to add support for MySQL as a persistence backend.

This is a simple implementation for the `MetadataDAO`, `ExecutionDAO` and `QueueDAO` to use the MySQL database. 

Instead of relying on Hibernate or JPA, a small and fast library called [Sql2o](https://www.sql2o.org) is used for the persistence to get optimal [performance](https://github.com/aaberg/sql2o#performance)

The schema generation is managed by the excellent [Flyway](https://flywaydb.org/) library.

For the `IndexDAO`, I am still unsure if it's worth creating a MySQL DAO. For the moment the existing embedded `ElasticSearchDAO` is reused.

For testing the new DAOs, I copied the unit tests from the Redis persistence module. All tests are passing for the new MySQL DAOs.

The unit tests are running on an embedded MySQL database on which the Flyway migration are first executed. This ensures that the DAOs are tested with a real MySQL database.


picaron:
Hi @v1r3n 

Can you please review the PR for the new proposed `mysql-persistence` module? 

The sooner the better since we are gonna have to go live very soon with this implementation so if it's not merged we will probably have to go live with our own fork temporarily.

Thanks in advance :)

mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLQueueDAO.java
```diff
@@ -0,0 +1,241 @@
...
+		if (count < 1) return Collections.emptyList();
+		return getWithTransaction(tx -> tx.createQuery(PEEK_MESSAGES)
+				.addParameter("queueName", queueName)
+				.addParameter("count", count)
+				.executeScalarList(String.class));
+	}
+
+	private List<String> popMessages(Connection connection, String queueName, List<String> messageIds) {
+		if (messageIds.isEmpty()) return messageIds;
+		List<String> stringIds = messageIds.stream().map(id -> format("'%s'", id)).collect(Collectors.toList());
```
tommack: This seems a little dangerous. Assuming it has been verified that all of the messageId strings are clean, a comment should be added here explaining why this is safe.

But, considering the proliferation of IN clauses and sql2o's [lack of support](https://github.com/aaberg/sql2o/issues/146), I'm not sure how good a fit that library is.

mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLExecutionDAO.java
```diff
@@ -0,0 +1,761 @@
...
+import com.netflix.conductor.dao.IndexDAO;
+import com.netflix.conductor.dao.MetadataDAO;
+import com.netflix.conductor.metrics.Monitors;
+
+class MySQLExecutionDAO extends MySQLBaseDAO implements ExecutionDAO {
+
+	private static final String ARCHIVED_FIELD = "archived";
+	private static final String RAW_JSON_FIELD = "rawJSON";
+
+	private static final String INSERT_SCHEDULED_TASK = "INSERT INTO task_scheduled (workflow_id, task_key, task_id) VALUES (:workflowId, :taskKey, :taskId)";
```
tommack: Why preference would be to place these constants within the methods in which they are used. When reading the code it is a lot easier to see the SQL and the placeholders together in one place.

mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLExecutionDAO.java
```diff
@@ -0,0 +1,761 @@
...
+	@Inject
+	MySQLExecutionDAO(IndexDAO indexer, MetadataDAO metadata, ObjectMapper om, Sql2o sql2o) {
+		super(om, sql2o);
+		this.indexer = indexer;
+		this.metadata = metadata;
+	}
+
+	@Override
+	public List<Task> getPendingTasksByWorkflow(String taskDefName, String workflowId) {
+		return getPendingTasksForTaskType(taskDefName).stream()
```
tommack: Why doesn't this include the `workflowId` in the query?

mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLExecutionDAO.java
```diff
@@ -0,0 +1,761 @@
...
+				.filter(task -> task.getWorkflowInstanceId().equals(workflowId))
+				.collect(Collectors.toList());
+	}
+
+	@Override
+	public List<Task> getTasks(String taskDefName, String startKey, int count) {
+		List<Task> tasks = new LinkedList<>();
```
tommack: Since you have the `count`, wouldn't an `ArrayList` be simpler?

mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLExecutionDAO.java
```diff
@@ -0,0 +1,761 @@
...
+
+		List<Task> pendingTasks = getPendingTasksForTaskType(taskDefName);
```
tommack: What is the typical ratio of `count` to `pendingTasks.size()`?

mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLExecutionDAO.java
```diff
@@ -0,0 +1,761 @@
...
+	public List<Task> getTasks(List<String> taskIds) {
+		if (taskIds.isEmpty()) return Lists.newArrayList();
+
+		List<String> taskIdParameters = taskIds.stream().map(taskId -> format("'%s'", taskId)).collect(Collectors.toList());
+
+		return getWithTransaction(c -> c.createQuery(format(GET_TASKS, join(taskIdParameters, ","))).executeScalarList(String.class))
+				.stream()
+				.filter(Objects::nonNull)
+				.map(taskJsonStr -> readValue(taskJsonStr, Task.class))
+				.collect(Collectors.toList());
```
tommack: Elsewhere I've noticed that the result of this method is sorted. It would be nice to document that the return type must be mutable *and* avoid the `toList()` method which doesn't guarantee mutability.

mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLExecutionDAO.java
```diff
@@ -0,0 +1,761 @@
...
+		return getTasks(taskIds);
+	}
+
+	@Override
+	public List<Task> getTasksForWorkflow(String workflowId) {
+		List<String> taskIds = getWithTransaction(c -> c.createQuery(GET_TASKS_FOR_WORKFLOW)
+				.addParameter("workflowId", workflowId)
+				.executeScalarList(String.class));
+
+		return getTasks(taskIds);
```
tommack: Why does this have to be done in two transactions?

mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLExecutionDAO.java
```diff
@@ -0,0 +1,761 @@
...
+				removeWorkflow(connection, workflowId);
+				removePendingWorkflow(connection, wf.getWorkflowType(), workflowId);
+			});
+
+			for(Task task : wf.getTasks()) {
+				removeTask(task.getTaskId());
+			}
+
+		} catch(Exception e) {
+			throw new ApplicationException(e.getMessage(), e);
```
tommack: No reason to copy the message of the cause. There's probably another string like `"Unable to remove workflow " + workflowId` that could be used here.

mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLExecutionDAO.java
```diff
@@ -0,0 +1,761 @@
...
+					.executeScalarList(String.class);
+
+			workflowIds.forEach(workflowId -> {
+				try {
+					Workflow wf = getWorkflow(workflowId);
+					if (wf.getCreateTime() >= startTime && wf.getCreateTime() <= endTime) {
+						workflows.add(wf);
+					}
+				} catch(Exception e) {
+					logger.error(e.getMessage(), e);
```
tommack: No need to log the message of the Exception twice. Add more context.

```
logger.error("Unable to something something on {}, workflowName, e);
```

mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLExecutionDAO.java
```diff
@@ -0,0 +1,761 @@
...
+		return taskJsonStr != null ? readValue(taskJsonStr, Task.class) : null;
+	}
+
+	@Override
+	public List<Task> getTasks(List<String> taskIds) {
+		if (taskIds.isEmpty()) return Lists.newArrayList();
+
+		List<String> taskIdParameters = taskIds.stream().map(taskId -> format("'%s'", taskId)).collect(Collectors.toList());
+
+		return getWithTransaction(c -> c.createQuery(format(GET_TASKS, join(taskIdParameters, ","))).executeScalarList(String.class))
```
tommack: I don't know anything about sql2o, but it I see a `ResultSetHandler` interface. Is building the `List<String>` more efficient than pulling the data directly off of the `ResultSet` (i.e. calling `readValue` from within the `ResultSetHandler` callback)?

mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLExecutionDAO.java
```diff
@@ -0,0 +1,761 @@
...
+	@Override
+	public List<String> getRunningWorkflowIds(String workflowName) {
+		Preconditions.checkNotNull(workflowName, "workflowName cannot be null");
+		return getWithTransaction(tx -> tx.createQuery(GET_PENDING_WORKFLOW_IDS).addParameter("workflowType", workflowName).executeScalarList(String.class));
+	}
+
+	@Override
+	public List<Workflow> getPendingWorkflowsByType(String workflowName) {
+		Preconditions.checkNotNull(workflowName, "workflowName cannot be null");
+		return getRunningWorkflowIds(workflowName).stream().map(this::getWorkflow).collect(Collectors.toList());
```
tommack: I don't have a good sense of how many `Workflow`s this might return, but it seems like adding a DAO method that directly supported returning a `List<Workflow>` would be more efficient than making an individual call for each item in the stream.

mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLQueueDAO.java
picaron: Good point @tommack, I am a bit uneasy with the lack of `IN` support as well. The thing is I didn't find that many libraries over JDBC that provide a convenient programming model and that are performant.

I don't think we want to bring in Hibernate/JPA for this right? Seems a bit heavy.

Do you know any well performing library that could replace Sql2o. I could consider switching the lib ... I want to have the best possible implementation.

Other alternative would be to add input sanitization logic and comments as you suggest.

mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLExecutionDAO.java
picaron: This is because the underlying table simply holds a mapping of task def name to task ids (similar to what is done in the Redis version)

mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLExecutionDAO.java
picaron: agree, most catch handlers I have copy-pasted them from the the RedisExecutionDAO and haven't paid much attention

I will review all try/catch blocks to make sure they are consistent

mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLExecutionDAO.java
picaron: not a very good reason, mostly because `getTasks` is a public DAO method and I cant change the signature

I will create an internal getTasks that takes a connection and reuse it in both places

picaron:
thanks for the pertinent comments @tommack 

I will look into addressing them

mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLQueueDAO.java
picaron: @tommack I found a better to do this still with Sql2o. The Query api has a withParams method that supports an array and expects the query to have a list of p1, p2, pN placeholders

So instead I will just generate a dynamic list of placeholders of the query and then use that API. That will ensure that the PreparedStatement can properly escape any dangerous SQL

luisnrkl:
Hi @picaron 

I'm also interested in a SQL implementation for Conductor, so i was looking into your code.

Did you try it with the Client provided?

On my tests it doesn't work, because it will always fail to Ack the task, due to the setUnackTimeout method always returning false.

Thanks,

picaron:
Hi @luisnrkl 

Thanks for your input.

So far, I've only tested with the same unit tests that existed for the Redis persistence modules.

No tests involve the `setUnackTimeout` and `processUnacks` methods on the DAO so that's probably why this wasn't caught.

Obviously the PR still has some rough edges, I will look into add tests for these two APIs and implementing the methods. I will also address the comments raised by @tommack

picaron:
Hi @luisnrkl 

I have added the missing implementation for `setUnackTimeout` and `processUnacks`

Can you perform your tests again?

mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLExecutionDAO.java
picaron: yes I agree ... but however there is special logic in the `getWorkflow` method that would be a bit heavy to replicate ... the `RedisExecutionDAO` is constructed the same way btw

picaron:
Hi @tommack 

I addressed all your comments and made a few improvements, can you please have another look when you get a chance?

picaron:
@v1r3n @tommack can you please review the PR again?

We addressed all comments raised by @tommack

v1r3n:
@picaron Thanks for addressing the comments.  Will get it merged this week.

vikramsingh2104:
@picaron The build after merge failed https://travis-ci.org/Netflix/conductor/builds/286218218 because QueueDAO interface has a new method setOffsetTime(), you will have to implement that in your MySQLQueueDAO class.

picaron:
thanks @vikramsingh2104 
will you revert or I will create another PR ?

picaron:
I suppose it will be simpler to create a PR to fix this small issue, I'll submit it shortly

picaron:
Here is the fix @vikramsingh2104 : https://github.com/Netflix/conductor/pull/343

Please review and merge

Issue #290, opened 2017-08-01
Problem with conductor docker image using dynomite db


eder-santos:
Hi, I'm getting the following error with the conductor docker image. It appears to be a problem with the `DynoJedisClient`, this error only occurs using the default `config.properties` (dynomite db) but it will not appear if I swith to the  `config-local.properties` file. Should I report this bug on the [Netflix/dyno-queues](https://github.com/Netflix/dyno-queues) project?

Thank you.

```
46854 [qtp1747721875-46] ERROR com.netflix.conductor.server.resources.GenericExceptionMapper  - java.lang.UnsupportedOperationException: not yet implemented
java.lang.RuntimeException: java.lang.UnsupportedOperationException: not yet implemented
        at com.netflix.dyno.queues.redis.RedisDynoQueue.pop(RedisDynoQueue.java:224)
        at com.netflix.conductor.dao.dynomite.queue.DynoQueueDAO.pop(DynoQueueDAO.java:150)
        at com.netflix.conductor.service.ExecutionService.poll(ExecutionService.java:109)
        at com.netflix.conductor.server.resources.TaskResource.batchPoll(TaskResource.java:105)
        at sun.reflect.GeneratedMethodAccessor46.invoke(Unknown Source)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at com.sun.jersey.spi.container.JavaMethodInvokerFactory$1.invoke(JavaMethodInvokerFactory.java:60)
        at com.sun.jersey.server.impl.model.method.dispatch.AbstractResourceMethodDispatchProvider$TypeOutInvoker._dispatch(AbstractResourceMethodDispatchProvider.java:185)
        at com.sun.jersey.server.impl.model.method.dispatch.ResourceJavaMethodDispatcher.dispatch(ResourceJavaMethodDispatcher.java:75)
        at com.sun.jersey.server.impl.uri.rules.HttpMethodRule.accept(HttpMethodRule.java:302)
        at com.sun.jersey.server.impl.uri.rules.RightHandPathRule.accept(RightHandPathRule.java:147)
        at com.sun.jersey.server.impl.uri.rules.ResourceClassRule.accept(ResourceClassRule.java:108)
        at com.sun.jersey.server.impl.uri.rules.RightHandPathRule.accept(RightHandPathRule.java:147)
[...]

PR #319, opened 2017-09-03
Fix incorrect dyno-jedis dependency version

picaron:
As mentioned in https://github.com/Netflix/conductor/issues/290, the dependency for `dyno-jedis` needs to be `1.5.6+`

Recent upgrade of `dyno-queues-redis` from `1.0.6` to `1.0.7` has downgraded `dyno-jedis` to 1.4.7 causing the issue #290.

Explicitly adding the dependency in `redis-persistence/build.gradle` to the latest (`1.5.9`, same as `dyno-core`) fixes the issue


picaron:
thanks @v1r3n 

one more thing: the 1.8.0 release is currently unusable because of this issue, is it possible to make a new 1.8.1 release soon?

PR #336, opened 2017-10-04
[CORE][ExecutionService] - get task by task type

math29:
Hello,

We would need the ability to get all tasks filtered by taskType without any count from ExecutionService (today on only can get SCHEDULED and IN_PROGRESS tasks with a mandatory count).

Thanks,

Math


math29:
Hello @vikramsingh2104, do you have any update about this review? :)

math29:
Hello @v1r3n @vikramsingh2104, do you have any news about this PR? :)

v1r3n:
This could lead to huge reads - its quite possible that number of tasks over the period grows to large size (millions).

How about using indexing system in elasticsearch to get the tasks?

math29:
We don't wanna use elasticsearch on our side for now.

@v1r3n what about using SSCAN for getTasks methods (adding 2 parameters to this method for pagination)

v1r3n:
@math29 smembers is going to be an expensive operation if there are many tasks by type.  I recommend switching to SSCAN with start and count (optional) for pagination.  That should work.

math29:
Ok, I'll change my PR to use SSCAN, thank you!

PR #339, opened 2017-10-06
Fix makefile

toddschilling:


toddschilling:
poop!  wrong source branch.

PR #355, opened 2017-10-30
Fix tasks getting executed multiple times with MySQL storage backend

picaron:
With the current MySQL Queue DAO implementation, tasks sometimes get executed multiple times even if the first attempt ends in success.

This PR is to fix that issue


picaron:
@v1r3n @vikramsingh2104 please review this small fix for the MySQL persistence implementation

picaron:
thanks @v1r3n are you going to create another 1.8.2 release candidate?

v1r3n:
@picaron yes, we have some pending changes in dev that we are testing.  Should create an RC soon.

picaron:
awesome, thanks 👍

PR #359, opened 2017-11-06
Downgrade Flyway version because of incompatibility with Percona 5.7 in strict mode

picaron:
We found out that Flyway versions greater than 4.1.x are not compatible with Percona XtraDB Cluster (see open issue on Flyway project: https://github.com/flyway/flyway/issues/1556)

So as a temporary workaround, I'm downgrading to the latest Flyway version that doesn't use the locking mode incompatible with Percona which is **4.0.3**


picaron:
@v1r3n @vikramsingh2104 can you review/merge this one-liner PR?

we found out the latest Flyway version is not compatible with Percona XtraDB Cluster (a MySQL HA solution) in strict mode.

so the only fix currently is to downgrade the Flyway version. I've tested with the downgrade version and didn't have any issues.

Issue #346, opened 2017-10-13
When changing to es5 persistence,  Local Server cannot start 


cquon:
I have followed the steps here to change to es5 module:
https://github.com/Netflix/conductor/blob/master/es5-persistence/README.md 

Then when trying to run the server locally, exception is thrown.

[cquon@localhost server]$ ../gradlew server
Inferred project: conductor, version: 1.8.2-SNAPSHOT
Publication nebula not found in project :.
Publication named 'nebula' does not exist for project ':' in task ':artifactoryPublish'.
:conductor-common:compileJava UP-TO-DATE
:conductor-common:processResources UP-TO-DATE
:conductor-common:classes UP-TO-DATE
:conductor-common:writeManifestProperties UP-TO-DATE
:conductor-common:jar UP-TO-DATE
:conductor-core:compileJava UP-TO-DATE
:conductor-core:processResources UP-TO-DATE
:conductor-core:classes UP-TO-DATE
:conductor-core:writeManifestProperties UP-TO-DATE
:conductor-core:jar UP-TO-DATE
:conductor-contribs:compileJava UP-TO-DATE
:conductor-contribs:processResources UP-TO-DATE
:conductor-contribs:classes UP-TO-DATE
:conductor-contribs:writeManifestProperties UP-TO-DATE
:conductor-contribs:jar UP-TO-DATE
:conductor-es5-persistence:compileJava UP-TO-DATE
:conductor-es5-persistence:processResources UP-TO-DATE
:conductor-es5-persistence:classes UP-TO-DATE
:conductor-es5-persistence:writeManifestProperties UP-TO-DATE
:conductor-es5-persistence:jar UP-TO-DATE
:conductor-jersey:compileJava UP-TO-DATE
:conductor-jersey:processResources UP-TO-DATE
:conductor-jersey:classes UP-TO-DATE
:conductor-jersey:writeManifestProperties UP-TO-DATE
:conductor-jersey:jar UP-TO-DATE
:conductor-mysql-persistence:compileJava UP-TO-DATE
:conductor-mysql-persistence:processResources UP-TO-DATE
:conductor-mysql-persistence:classes UP-TO-DATE
:conductor-mysql-persistence:writeManifestProperties UP-TO-DATE
:conductor-mysql-persistence:jar UP-TO-DATE
:conductor-redis-persistence:compileJava UP-TO-DATE
:conductor-redis-persistence:processResources UP-TO-DATE
[...]

PR #369, opened 2017-11-18
Fix log4j version that is causing ClassNotFoundException

cquon:
There is an exception being thrown with specific log4j version.

This resolves issue:
https://github.com/Netflix/conductor/issues/346


picaron:
@v1r3n @vikramsingh2104 any news on releasing a new RC soon?

PR #372, opened 2017-11-23
Added redis queue persistence Artifact

grfeng:
1. New Artifact redis queue persistence, Just a Queue persistence
1. es5-persistence check logs size.
1. Fix mysql-persistence taskKey naming rule.


grfeng:
Some redis testing use
```
JedisPool pool = new JedisPool(config, "localhost", 6379);
```
CI build will be failed.

v1r3n:
question on the 1) above - I am trying to understand the need for a separate redis-queue-persistence module.  How is it different from the queues already being used?

grfeng:
Hi , @v1r3n  
1. https://github.com/Netflix/dyno/issues/54, Because the Redis Cluster of our company needs a password
1. Dyno Is Redis Cluster supported？
1. Just using Redis to implement QueueDAO, Mysql implements ExecutionDAO and MetadataDAO, ES implements IndexDAO

v1r3n:
@grfeng 
We have not tested with redis cluster but it should work with minimal changes required to accept password.  The current version does work with AWS ElasticCache and a regular redis server.

You should be able to use the existing persistence module that supports redis queues for just queues and use mysql for other needs rather than creating a new module.

grfeng:
Ok, thanks for your feedback, I will close PR, but we will keep it in our production project, because there is still some inapplicable, such as sharding, etc.

PR #374, opened 2017-11-29
Minimal adjust, Don't to index an empty task exec log and more things

grfeng:
Some bug fixed ,e.g. Don't to index an empty task exec log

### DIFF
--- a/jersey/src/main/java/com/netflix/conductor/server/resources/MetadataResource.java
+++ b/jersey/src/main/java/com/netflix/conductor/server/resources/MetadataResource.java
@@ -65,7 +65,7 @@ public class MetadataResource {
 	}
 	
 	@PUT
-	@Path("/workflow")
+	@Path("/workflows")
 	@ApiOperation("Create or update multi  workflow definition")
 	public void update(List<WorkflowDef> defs) throws Exception{
 		service.updateWorkflowDef(defs);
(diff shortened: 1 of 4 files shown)


jersey/src/main/java/com/netflix/conductor/server/resources/MetadataResource.java
```diff
@@ -65,12 +65,19 @@ public void create(WorkflowDef def) throws Exception{
 	}
 	
 	@PUT
-	@Path("/workflow")
-	@ApiOperation("Create or update workflow definition")
+	@Path("/workflows")
```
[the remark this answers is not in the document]
grfeng: ok!

mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLExecutionDAO.java
```diff
@@ -174,7 +174,7 @@ public void removeTask(String taskId) {
 			return;
 		}
 
-		String taskKey = task.getReferenceTaskName() + "" + task.getRetryCount();
+		String taskKey = task.getReferenceTaskName() + "_" + task.getRetryCount();
```
v1r3n: @grfeng seems this will also break existing data stored in the database.

grfeng: @v1r3n Look this https://github.com/grfeng/conductor/blob/a3194665493238d6e118e34ca5b1df702942e53c/mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLExecutionDAO.java#L101

PR #381, opened 2017-12-08
Make archiving optional in deleteWorkflow API & Run all indexing except workflow indexing in a separate threadpool

clari-akhilesh:
@v1r3n , @vikramsingh2104 
Added an optional parameter `archiveWorkflow` in removeWorkflow API whose default value is always `true`. If `archiveWorkflow=false` we also delete the workflowId from the index as this workflow no longer exists. This change does not break any clients who currently do not pass this parameter and the behavior is unchanged, i.e. archives the workflow, if the parameter is not passed. 
The second commit is the one we had discussed previously to make all indexing except workflow indexing run in a separate thread pool.
It would be great if these are included in the next release. Thanks!


redis-persistence/src/main/java/com/netflix/conductor/dao/index/ElasticSearchDAO.java
```diff
@@ -118,9 +121,16 @@
 	private static final TimeZone gmt = TimeZone.getTimeZone("GMT");
 	    
     private static final SimpleDateFormat sdf = new SimpleDateFormat("yyyyMMWW");
+
```
pctreddy: `ElasticSearchDAO` is a @singleton, whats the need of the static initialization.

clari-akhilesh: I overlooked the fact that `ElasticSearchDAO` was a singleton class. I'll make this a member variable. Thanks!

clari-akhilesh: @pctreddy Done.

pctreddy: Initializing instance variable `executorService` in a static initialization block is not a good idea, can you move the initialization of `executorService` into the constructor

clari-akhilesh: @pctreddy , initialization is in the constructor.

pctreddy: Missed it in the collapsed github pull request view.

clari-akhilesh:
@v1r3n @vikramsingh2104 @pctreddy @cyzhao , any idea when this will be merged? Our production release depends on this. Thanks!

PR #382, opened 2017-12-10
Don't try to index an empty task exec log (es5)

cquon:
When Task Exec Logs are empty, trying to execute a request will lead to exceptions, so if they are empty do not try to perform this action.


cquon:
Saw this PR addresses this:

https://github.com/Netflix/conductor/pull/374

PR #384, opened 2017-12-12
MWI-5 Added failedTaskId variable in Workflow

pctreddy:


common/src/main/java/com/netflix/conductor/common/run/Workflow.java
```diff
@@ -142,13 +146,24 @@ public void setWorkflowId(String workflowId) {
...
+	 * {@link #failedTaskId} in case of any task that has a status of {@link com.netflix.conductor.common.metadata.tasks.Task.Status#FAILED}
+	 *
+	 * @param tasks: The list of the tasks that belong to the workflow
 	 */
 	public void setTasks(List<Task> tasks) {
 		this.tasks = tasks;
+		if (Objects.nonNull(tasks)) {
+			this.failedTaskId = tasks.stream()
+					.filter(task -> Task.Status.FAILED.equals(task.getStatus()))
+					.findFirst()
```
cyzhao: We may want to return CSV of failed tasks in case there are multiple failed tasks.

common/src/main/java/com/netflix/conductor/common/run/Workflow.java
```diff
@@ -142,13 +146,24 @@ public void setWorkflowId(String workflowId) {
...
 	/**
-	 * @param tasks the tasks to set
+	 * This method sets the list of tasks that belong to the workflow and also updates the
+	 * {@link #failedTaskId} in case of any task that has a status of {@link com.netflix.conductor.common.metadata.tasks.Task.Status#FAILED}
+	 *
+	 * @param tasks: The list of the tasks that belong to the workflow
 	 */
 	public void setTasks(List<Task> tasks) {
 		this.tasks = tasks;
+		if (Objects.nonNull(tasks)) {
```
cyzhao: tasks != null is more intuitive,  Objects.nonNull exists to be used as a predicate in stream filter method calls.

cyzhao:
ui also needs to be updated

common/src/main/java/com/netflix/conductor/common/run/Workflow.java
```diff
@@ -72,6 +74,8 @@ public boolean isSuccessful(){
 	private String reRunFromWorkflowId;
 	
 	private String reasonForIncompletion;
+
+	private String failedTaskId;
```
cyzhao: Capture referenceTaskName instead of taskId.

common/src/main/java/com/netflix/conductor/common/run/Workflow.java
```diff
@@ -142,13 +146,24 @@ public void setWorkflowId(String workflowId) {
...
+	 * This method sets the list of tasks that belong to the workflow and also updates the
+	 * {@link #failedTaskId} in case of any task that has a status of {@link com.netflix.conductor.common.metadata.tasks.Task.Status#FAILED}
+	 *
+	 * @param tasks: The list of the tasks that belong to the workflow
 	 */
 	public void setTasks(List<Task> tasks) {
 		this.tasks = tasks;
+		if (Objects.nonNull(tasks)) {
+			this.failedTaskId = tasks.stream()
+					.filter(task -> Task.Status.FAILED.equals(task.getStatus()))
```
cyzhao: Do we need to worry about other statuses: CANCELED, COMPLETED_WITH_ERRORS, TIMED_OUT?

common/src/main/java/com/netflix/conductor/common/run/Workflow.java
pctreddy: Changed this to a `refereceTaskName`

common/src/main/java/com/netflix/conductor/common/run/Workflow.java
pctreddy: Changed from `Objects.nonNull(tasks)` ->  `tasks !=null` to be more intuitive.

common/src/main/java/com/netflix/conductor/common/run/Workflow.java
pctreddy: Changed this to a CSV.

common/src/main/java/com/netflix/conductor/common/run/Workflow.java
pctreddy: @vikramsingh2104 Do we need to worry about states other than `Task.Status.FAILED`

PR #387, opened 2017-12-15
Upgrade gradlew to 4.4 and display the failed tasks in Workflow summary

pctreddy:


core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -434,8 +434,16 @@ public void updateTask(TaskResult result) throws Exception {
...
 			task.setEndTime(System.currentTimeMillis());
 		}
+
 		edao.updateTask(task);
 
+		//If the task has failed update the failed task reference name in the workflow.
+		//This gives the ability to look at workflow and see what tasks have failed at a high level.
+		if(Status.FAILED.equals(task.getStatus())) {
+			wf.getFailedReferenceTaskNames().add(task.getReferenceTaskName());
+			edao.updateWorkflow(wf);
```
cyzhao: We need to change WorkflowSummary as well, Workflow is converted to WorkflowSummary first, which is then indexed, failedReferenceTaskNames need to be added to WorkflowSummary and copied from Workflow in its constructor.

vikramsingh2104:
Changes look good.

v1r3n:
@pctreddy can you verify that you can create releases with upgraded nebula4?

Issue #298, opened 2017-08-10
HTTP Request ContentType urlencoded not supported.


heajuenhwang:
We are trying to call api with urlencoded contentType. 
From the api doc, urlencoded is not supported.
Is there a workaround way, we can make urlencoded Content-Type working?
The following is the task definition we are trying to call.
```
{
"name": "task_4",
"taskReferenceName": "tesk_4",
"inputParameters": {
"http_request": {
"uri": "url",
"method": "POST",
"contentType":"application/x-www-form-urlencoded",
"body":
{
"key1":"value1",
"key2":"value2"
}
}
},
"type": "HTTP",
      "startDelay": 0
    }
```

v1r3n:
@heajuenhwang content type should be `application/json` .

heajuenhwang:
Any plan to support different Content Type than application/json?

v1r3n:
@heajuenhwang all the inputs and outputs are json, so it will not make sense to support other content types.

nsiitk:
Oauth specification 
https://tools.ietf.org/html/rfc6749#section-4.1.3 talks about Access token request using content-type application/x-www-form-urlencoded. I have an authentication service which which requires this content-type. If a workflow has http task based on authentication it would need content-type flexibility.

gansa1986:
Do we have this enhancement?

apanicker-nflx:
@gansa1986 This is currently not implemented. However, feel free to submit a PR for this enhancement and we would be happy to work with you towards adding this.

PR #389, opened 2017-12-18
issue #298 HTTP Request ContentType urlencoded not supported

nsiitk:
used http header content-type to overrider application/json 
   	 "http_request": {
	      	  	"uri": "http://localhost:8000/oauth/token",
	          	"method": "POST",        	"body":"username=${workflow.input.user}&password=${workflow.input.pwd}&client_id=trusted-app&grant_type=password",
	        	"contentType": "application/x-www-form-urlencoded",
	        	"accept":"application/json",
				"headers":{"content-type":"application/x-www-form-urlencoded"}
	        	}
	      	}


v1r3n:
@nsiitk can you send the PR against the `dev` branch?

nsiitk:
@v1r3n changed PR to dev. Don't understand why 2 test cases failing as my change don't impact those test cases.

contribs/src/main/java/com/netflix/conductor/contribs/http/HttpTask.java
```diff
@@ -159,13 +159,14 @@ protected HttpResponse httpCall(Input input) throws Exception {
 			OAuthSecrets secrets = new OAuthSecrets().consumerSecret(input.oauthConsumerSecret);
 			client.addFilter(new OAuthClientFilter(client.getProviders(), params, secrets));
 		}
-
-		Builder builder = client.resource(input.uri).type(MediaType.APPLICATION_JSON);
+		final String contentType = "content-type";
```
v1r3n: What about different variations?  e.g. `Content-Type` ?

nsiitk: http headers are case insensitive so header hashmap should have all keys in same case.
If I change the setter of header in Input class all header will be lower case.
public void setHeaders(Map<String, Object> headers) {
			 for (String key : headers.keySet()) {
		            this.headers.put(key.toLowerCase(),headers.get(key));
		        }
		}

picaron: @nsiitk @v1r3n according to the Conductor [docs](https://netflix.github.io/conductor/metadata/systask/#http), there is a distinct parameter `contentType` that should be used for setting the content type of the request. 

I created a similar PR to use that parameter, see PR #391

v1r3n: @nsiitk as @picaron mentioned, `contentType` is a distinct parameter and should be used to override the contentType.

Issue #356, opened 2017-10-30
ElasticSearch Exception in Conductor Server logs


clari-akhilesh:
When looking up a workflow in the UI i.e. when I click a particular workflow instance id in the UI, I see the following exception in the Conductor Server logs:
```
999411 [qtp277630005-73] ERROR com.netflix.conductor.dao.index.ElasticSearchDAO  - all shards failed
[...]

PR #408, opened 2018-01-23
Nsiitk patch issue 356

nsiitk:
in absence of createTime search query fails. ignore the field for the missing record will not raise exception.


v1r3n:
Addressing the issue with #356 in the following commit
https://github.com/Netflix/conductor/commit/6e71f0bb07c1003f7dccbde6584ded4681d1f613

Issue #424, opened 2018-02-11
[QUESTION] conductor-mysql-persistence Sql2o dependency


mashurex:
@v1r3n and @picaron specifically, I was wondering about the  [Sql2o](https://www.sql2o.org/)  dependency that the MySQL persistence library has. 

It does not seem to be a very popular query library and has some (in my opinion) short comings compared to other popular frameworks. Additionally, questions are being raised about it's maintenance as it has been at an RC with some open issues and no activity for months. 

It seems that it is used out of convenience for the **conductor-mysql-persistence** library, which I completely understand, but it doesn't seem critical to the functionality of the DAOs. On my team, and many of the teams I've been on before, bringing in _another_ SQL framework is usually a pretty big deal.

Since the DAO implementations generally hide all the transaction management, query building, etc. would it make the **conductor-mysql-persistence** library more portable to just use lower level `java.sql.*` classes?

I have re-written them for my own use in such a fashion, but wouldn't want to submit a PR if there were reasons I'm not aware of as to their own implementation.

picaron:
@mashurex correct, I used Sql2o for convenience and mostly because it's lightweight, it fast to develop the DAOs with it and also it masks away a lot of boilerplate code that we usually see we using raw JDBC. Porting the MySQL DAOs to pure JDBC would indeed remove that dependency. My only concern is the readability/maintainability of the DAO. I will it harder to maintain a codebase with lot of low-level code. I am certainly interested to look at your implementation.

I would certainly be in favor of your proposed approach if the business logic in the DAOs is easy to read without lower-level JDBC boilerplate "noise". If you can somehow abstract the low-level code and keep the DAOs business methods clean then that'd be perfect.

picaron:
your thoughts on the topic @v1r3n ?

mashurex:
@picaron, I created PR #426 to show my thoughts on this for the `MetadataDAO` implementation.

mashurex:
Closing this for PR #426

Issue #399, opened 2018-01-04
BUG - MySqlQueueDAO::pollMessages always fails


marc-christian-schulze:
The method pollMessages of the MySQLQueueDAO always fails with an Exception.
Looking at the Implementation https://github.com/Netflix/conductor/blob/master/mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLQueueDAO.java#L72 the method first calls pop() which calls popMessages() which marks all messages returned as popped. The following readMessages() call will try to find messages with the given ids but not marked as popped which finally fails.

The only use case for this broken method is currently in the DynoQueue implementation.

v1r3n:
@marc-christian-schulze #417 should have fixed this.  Closing this, but please reopen if the issue still exists.  Will create an RC or you can use the latest snapshot to verify the changes.

PR #426, opened 2018-02-13
Sql2o removal for MySQL JDBC

mashurex:
@picaron This is a sample of the `java.sql.*` only implementation from #424 , I've included the `MetadataDAO` implementation- all the new work is in the `com.netflix.conductor.dao.jdbc` package so that it's easier to compare side by side with the existing `com.netflix.conductor.dao.mysql` package.

# Approach
Maintain the same developer flow and ergonomics that @picaron set with the usage of the Sql2o library and `MySQLBaseDAO`.

I essentially reused all existing queries, just swapping out named params for standard `PreparedStatement` bind question marks, `?`.

I created a few `@FunctionalInterface` instances for Java 8 style `Consumer` and `Function` patterns that allow for `SQLExceptions` to be handled by wrapper methods in the `JdbcBaseDAO` and `Query` classes.

## JdbcBaseDAO Class
The `jdbc.JdbcBaseDAO` mimics the `MySQLBaseDAO` conventions heavily with the `getWithTransaction` and `withTransaction` methods, but adds additional convenience wrappers as well. 

## Query Builder Class
the `jdbc.Query` class is inspired by the nice ergonomics of Sql2o, bet eschews a lot of the generic conversion and data type handling for the more simple use cases encountered in the Conductor DAOs.

**Highlights**

- Builder pattern _addParameter_ methods for quicker chaining of parameter binding.
- Primitive batch parameter addition via the `addParameters(Object...)` method
- `addJsonParameter(Object)` as a convenience method for the repeated persistence of Conductor objects as JSON strings in columns
- `SQLException` conversion to runtime `ApplicationException` exceptions to remove the massive `try/catch` boilerplate required by `PreparedStatement` and `ResultSet` interactions
- Convenience methods like `boolean exists()` and `boolean delete()` that eschew the boiler plate code of checking a `ResultSet` for affected rows or retrieving a scalar value
- Use of Java _try with resources_ for automatic management of `PreparedStatement` and `ResultSet` instances

### DIFF
--- a/mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLBaseDAO.java
+++ b/mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLBaseDAO.java
@@ -1,91 +1,185 @@
 package com.netflix.conductor.dao.mysql;
 
-import static java.sql.Connection.TRANSACTION_READ_COMMITTED;
-
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.collect.ImmutableList;
+import com.netflix.conductor.core.execution.ApplicationException;
 import java.io.IOException;
+import java.sql.Connection;
+import java.sql.SQLException;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.Arrays;
 import java.util.List;
 import java.util.function.Consumer;
-import java.util.function.Function;
-import java.util.stream.Collectors;
-import java.util.stream.IntStream;
-
+import javax.sql.DataSource;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.sql2o.Connection;
-import org.sql2o.Sql2o;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.google.common.collect.ImmutableList;
 
-abstract class MySQLBaseDAO {
+/**
+ * @author mustafa
+ */
+public abstract class MySQLBaseDAO {
+    private static final List<String> EXCLUDED_STACKTRACE_CLASS = ImmutableList.of(
+            MySQLBaseDAO.class.getName(),
+            Thread.class.getName()
+    );
 
-	private static final List<String> EXCLUDED_STACKTRACE_CLASS = ImmutableList.of("com.netflix.conductor.dao.mysql.MySQLBaseDAO", "java.lang.Thread");
+    protected final Logger logger = LoggerFactory.getLogger(getClass());
+    protected final ObjectMapper objectMapper;
+    protected final DataSource dataSource;
 
-	protected final Sql2o sql2o;
-	protected final ObjectMapper om;
-	protected final Logger logger = LoggerFactory.getLogger(getClass());
+    protected MySQLBaseDAO(ObjectMapper om, DataSource dataSource) {
+        this.objectMapper = om;
+        this.dataSource = dataSource;
+    }
 
-	protected MySQLBaseDAO(ObjectMapper om, Sql2o sql2o) {
-		this.om = om;
-		this.sql2o = sql2o;
-	}
+    protected final LazyToString getCallingMethod() {
+        return new LazyToString(() -> Arrays.stream(Thread.currentThread().getStackTrace())
+                .filter(ste -> !EXCLUDED_STACKTRACE_CLASS.contains(ste.getClassName()))
+                .findFirst()
+                .map(StackTraceElement::getMethodName)
+                .orElseThrow(() -> new NullPointerException("Cannot find Caller")));
+    }
 
-	protected String toJson(Object value) {
-		try {
-			return om.writeValueAsString(value);
-		} catch (JsonProcessingException e) {
-			throw new RuntimeException(e);
-		}
-	}
+    protected String toJson(Object value) {
+        try {
+            return objectMapper.writeValueAsString(value);
+        } catch (JsonProcessingException ex) {
+            throw new ApplicationException(ApplicationException.Code.INTERNAL_ERROR, ex);
+        }
+    }
 
-	protected <T>T readValue(String json, Class<T> clazz) {
-		try {
-			return om.readValue(json, clazz);
-		} catch (IOException e) {
-			throw new RuntimeException(e);
-		}
-	}
+    protected <T> T readValue(String json, Class<T> tClass) {
+        try {
+            return objectMapper.readValue(json, tClass);
+        } catch (IOException ex) {
+            throw new ApplicationException(ApplicationException.Code.INTERNAL_ERROR, ex);
+        }
+    }
 
-	protected <R> R getWithTransaction(Function<Connection, R> function) {
-		Instant start = Instant.now();
-		StackTraceElement caller = Arrays.stream(Thread.currentThread().getStackTrace()).filter(ste -> !EXCLUDED_STACKTRACE_CLASS.contains(ste.getClassName())).findFirst().get();
-		logger.debug("{} : starting transaction", caller.getMethodName());
-		try (Connection connection = sql2o.beginTransaction(TRANSACTION_READ_COMMITTED)) {
-			final R result = function.apply(connection);
-			connection.commit();
-			return result;
-		} finally {
-			Instant end = Instant.now();
-			logger.debug("{} : took {}ms", caller.getMethodName(), Duration.between(start, end).toMillis());
-		}
-	}
+    protected <T> T readValue(String json, TypeReference<T> typeReference) {
+        try {
+            return objectMapper.readValue(json, typeReference);
+        } catch (IOException ex) {
+            throw new ApplicationException(ApplicationException.Code.INTERNAL_ERROR, ex);
+        }
+    }
 
-	protected void withTransaction(Consumer<Connection> consumer) {
-		getWithTransaction(connection -> {
-			consumer.accept(connection);
-			return null;
-		});
-	}
+    /**
+     * Initialize a new transactional {@link Connection} from {@link #dataSource} and pass it to {@literal function}.
+     * <p/>
+     * Successful executions of {@literal function} will result in a commit and return of
+     * {@link TransactionalFunction#apply(Connection)}.
+     * <p/>
+     * If any {@link Throwable} thrown from {@code TransactionalFunction#apply(Connection)} will result in a rollback
+     * of the transaction
+     * and will be wrapped in an {@link ApplicationException} if it is not already one.
+     * <p/>
+     * Generally this is used to wrap multiple {@link #execute(Connection, String, ExecuteFunction)} or
+     * {@link #query(Connection, String, QueryFunction)} invocations that produce some expected return value.
+     *
+     * @param function The function to apply with a new transactional {@link Connection}
+     * @param <R>      The return type.
+     * @return The result of {@code TransactionalFunction#apply(Connection)}
+     * @throws ApplicationException If any errors occur.
+     */
+    protected <R> R getWithTransaction(TransactionalFunction<R> function) {
+        Instant start = Instant.now();
+        LazyToString callingMethod = getCallingMethod();
+        logger.trace("{} : starting transaction", callingMethod);
 
-	/**
-	 * This will inject a series of p1, p2, ... placeholders in the given query template so it can then be used
-	 * in conjunction with the withParams method on the Sql2o Query object.
-	 *
-	 * The withParams method in the Query class loops through each element in the given array and adds a prepared statement for each.
-	 * For each element found in the array, a pN placeholder should exists in the query.
-	 *
-	 * This is useful for generating the IN clause since Sql2o does not support passing directly a list
-	 *
-	 * @param queryTemplate a query template with a %s placeholder where the variable size parameters placeholders should be injected
-	 * @param numPlaceholders the number of placeholders to generated
-	 * @return
-	 */
-	protected String generateQueryWithParametersListPlaceholders(String queryTemplate, int numPlaceholders) {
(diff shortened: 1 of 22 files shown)


picaron:
sounds very promising @mashurex !

It would be good to get approval from @v1r3n or another of the project owners before you spend more time on the PR. If they have no objection with this change I certainly don't.

Also I forgot to mention one of the reason for choosing Sql2o over other libraries was performance but raw JDBC should be even faster 🏎 

I love the idea of removing the Sql2o dependency but keeping the same "ergonomics" as you say 👍

mashurex:
@picaron It only took about 30 seconds to guess why Sql2o would be included. Raw JDBC in Java is painfully repetitive. Getting more familiar w/ Sql2o, I was able to come up with what I feel like is a decent wrapper around the native calls without introducing a crazy amount of complexity or re-inventing the wheel and yet introduce some of the very nice things that Sql2o gives to a developer.

picaron:
I haven't used Raw JDBC in a long time but I remembered it was very painful because it required so much repetitive boilerplate code.

a decent wrapper with an API similar to Sql2o is the way to go I think because otherwise if the DAOs end-up being polluted with boilerplate then they become much harder to read and maintain ...

mashurex:
@v1r3n and @picaron this PR is code complete from a DAO interface implementation standpoint. I did not put in a `Module` as I'm waiting for your thought on this work. If taken, I'm assuming we'd move this into and replace the `mysql` package.

- ExecutionDAO implemented
- MetadataDAO implemented
- QueueDAO implemented
- All `MySQL*Test` unit tests re-implemented for the Jdbc based DAOs
- Brought in and ported the work from #399 (as well as tests)

v1r3n:
@mashurex @picaron Thanks for the PR and details on the JDBC vs. Sql2o
IMHO, given that we do not have a lot of complexity in the DAOs and also the nature of the DAOs do not evolve that frequently as otherwise in a typical app, plain JDBC with some wrappers to make things easier will be a good balance.  

Using direct JDBC also helps those who are not very familiar with Sql2o or other ORM frameworks get up to speed quickly.  Conductor is a framework in the end and given that I would prefer to keep it as light as possible.

My +1 is for JDBC.

v1r3n:
@mashurex do you want to add the rest of the stuff (module etc) and update the PR so it replaces the existing implementation?

@picaron @pctreddy

mashurex:
@v1r3n Will do, was just waiting on everyone's feedback.

mashurex:
@v1r3n and @picaron, the work is ready for review and merge, pending any feedback from you guys.

mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLBaseDAO.java
```diff
@@ -1,91 +1,202 @@
...
-	protected String generateQueryWithParametersListPlaceholders(String queryTemplate, int numPlaceholders) {
-		String paramsPlaceholders = String.join(",", IntStream.rangeClosed(1, numPlaceholders).mapToObj(paramNumber -> ":p" + paramNumber).collect(Collectors.toList()));
-		return String.format(queryTemplate, paramsPlaceholders);
-	}
+/**
+ * @author mustafa
+ */
+public abstract class MySQLBaseDAO {
+    private static final List<String> EXCLUDED_STACKTRACE_CLASS = ImmutableList.of(
+        "com.netflix.conductor.dao.jdbc.JdbcBaseDAO", "java.lang.Thread");
```
picaron: this probably needs to be updated

mashurex: Oops! Good catch!

mashurex:
@picaron I fixed that stack trace class string.

picaron:
thanks @mashurex , didn't have time to do a full review

I'll continue asap

mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLBaseDAO.java
```diff
@@ -1,91 +1,202 @@
...
-	protected String generateQueryWithParametersListPlaceholders(String queryTemplate, int numPlaceholders) {
-		String paramsPlaceholders = String.join(",", IntStream.rangeClosed(1, numPlaceholders).mapToObj(paramNumber -> ":p" + paramNumber).collect(Collectors.toList()));
-		return String.format(queryTemplate, paramsPlaceholders);
-	}
+/**
+ * @author mustafa
+ */
+public abstract class MySQLBaseDAO {
+    private static final List<String> EXCLUDED_STACKTRACE_CLASS = ImmutableList.of(
+        "com.netflix.conductor.dao.mysql.MySQLBaseDAO", "java.lang.Thread");
```
picaron: maybe you could replace the strings with the class reference, for example:

`com.netflix.conductor.dao.mysql.MySQLBaseDAO.class.getName()` and `java.lang.Thread.class.getName()`

this way if those classes are renamed will make sure we don't forget to update this line

mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLBaseDAO.java
```diff
@@ -1,91 +1,202 @@
...
+     * <p/>
+     * Generally this is used to wrap multiple {@link #execute(Connection, String, ExecuteFunction)} or
+     * {@link #query(Connection, String, QueryFunction)} invocations that produce some expected return value.
+     *
+     * @param function The function to apply with a new transactional {@link Connection}
+     * @param <R>      The return type.
+     * @return The result of {@code TransactionalFunction#apply(Connection)}
+     * @throws ApplicationException If any errors occur.
+     */
+    protected <R> R getWithTransaction(TransactionalFunction<R> function) {
```
picaron: I love what you did with the overall PR but this method I find it very hard to read

it's indeed difficult to manage all the flows of exceptions. The `Connection` object supports auto-closeable so it can be used in a try with resource block...

I tried to simplify the code a bit to improve readability, here's what I came up with:

```
protected <R> R getWithTransaction(TransactionalFunction<R> function) {
	Instant start = Instant.now();
	StackTraceElement caller = getCaller();
	logger.trace("{} : starting transaction", caller.getMethodName());

	try(Connection tx = dataSource.getConnection()) {
		boolean previousAutoCommitMode = tx.getAutoCommit();
		tx.setAutoCommit(false);
		try {
			R result = function.apply(tx);
			tx.commit();
			return result;
		} catch (SQLException ex) {
			tx.rollback();
			throw new ApplicationException(ApplicationException.Code.BACKEND_ERROR, ex);
		} finally {
			tx.setAutoCommit(previousAutoCommitMode);
		}
	} catch (SQLException ex) {
		throw new ApplicationException(ApplicationException.Code.BACKEND_ERROR, ex);
	} finally {
		logger.trace("{} : took {}ms", caller.getMethodName(), Duration.between(start, Instant.now()).toMillis());
	}
}
```

mashurex: This is much easier to read, the reason the checks for `logger.isTraceEnabled()` were being used is that this method gets called very frequently and I didn't want to burden the execution with parsing StackTraceElements if they weren't going to be used. 

I didn't spend a lot of time measuring what kind of execution cost `getCaller()` has, but I figured that if we are not logging `trace` level messages, then there is no reason to parse the `StackTraceElement`. This adds a lot of noise with if statements, but it gives us the best _likelihood_ of faster execution times.

I'm really not sure why I didn't use a try with resources on the connection, it was probably an oversight on my part. I also see that you added `previousAutoCommitMode` which is a fantastic idea.

picaron: regarding the logger, since here we are using SLF4J with parameterized logging, the logger.isTraceEnabled() is redundant.

have a look at : https://www.slf4j.org/faq.html#logging_performance

mashurex: That's correct and we're using it, except for we are executing `getCaller()` no matter if the Logger is trace enabled or not. In my experience with this there's a trade off between a little harder to read and not executing debug/trace level calculations if we don't need to. In my other projects we usually have utility classes to handle this kind of stuff to reduce the boilerplate.

Again, I haven't measured the performance impact in any real detail but I assume that wrapping the `getCaller()` in a `isTraceEnabled()` block is more of a micro-optimization than a true performance benefit. I agree that it makes it harder to follow the code, so I think it's perfectly reasonable to follow your approach of removing the guard blocks and simplifying it.

picaron: you are right @mashurex I overlooked that!

how about this: 

```
protected <R> R getWithTransaction(TransactionalFunction<R> function) {
        Instant start = Instant.now();
        LazyToString callingMethod = getLazyCallingMethod();
        logger.trace("{} : starting transaction", callingMethod);

        try(Connection tx = dataSource.getConnection()) {
            boolean previousAutoCommitMode = tx.getAutoCommit();
            tx.setAutoCommit(false);
            try {
                R result = function.apply(tx);
                tx.commit();
                return result;
            } catch (Throwable th) {
                tx.rollback();
                throw new ApplicationException(ApplicationException.Code.BACKEND_ERROR, th);
            } finally {
                tx.setAutoCommit(previousAutoCommitMode);
            }
        } catch (SQLException ex) {
            throw new ApplicationException(ApplicationException.Code.BACKEND_ERROR, ex);
        } finally {
            logger.trace("{} : took {}ms", callingMethod, Duration.between(start, Instant.now()).toMillis());
        }
    }

    protected final LazyToString getLazyCallingMethod() {
        return new LazyToString(() -> Arrays.stream(Thread.currentThread().getStackTrace())
            .filter(ste -> !EXCLUDED_STACKTRACE_CLASS.contains(ste.getClassName()))
            .findFirst()
            .map(StackTraceElement::getMethodName)
            .orElseThrow(() -> new NullPointerException("Cannot find Caller")));
    }
```

and the `LazyToString` class:

```
package com.netflix.conductor.dao.mysql;

import java.util.function.Supplier;

class LazyToString {
    private final Supplier<String> supplier;

    LazyToString(Supplier<String> supplier) {
        this.supplier = supplier;
    }

    @Override
    public String toString() {
        return supplier.get();
    }
}
```

mashurex: I really like this. I'll get it in today.

mashurex:
@picaron I've added the fix from #448 into this PR as well as the lazy stack trace retrieval.

picaron:
good work @mashurex ! thanks!

there also seems to be a merge conflict to resolve...

mashurex:
@picaron This is probably where my Github PR naïveté comes in as I wasn't exactly sure who is responsible for merging the conflict. I'll get it fixed up.

mashurex:
@picaron we’re merge ready but I’d like your opinion on the unit tests I added to the `MySQLQueueDAOTest` test class.

I was having some timing/sequence failures when the build would push to Travis and run the unit tests. I _think_ it’s because the timestamps in the DB columns don’t have MS resolution and the VM layer in Travis when calling `System.currentTimeMillis()` is only giving 10ms resolution. 

The resulting issue is that if we push X messages onto the queue in a for loop, they all have the exact same `created_on` (and `deliver_on` for a given offset). That means when we run our peek query against the DB we have no guarantee of the exact order of messages that hit the queue in the same second with the same deliver on, because they all have the same timestamp for any given 1 second window.

I changed my tests from expecting a particular order to testing for the messages getting polled being in the collection of expected messages.

mashurex:
@picaron and @v1r3n is there any chance we can get this reviewed and merged soon? Keeping it up to date and conflict free is getting a bit troublesome.

mashurex:
Sounds good. I will get the migrations fixed ASAP.

mashurex:
@picaron Schema changes moved to V2 and pushed.

mysql-persistence/src/test/resources/logback-test.xml
```diff
@@ -1,17 +1,18 @@
...
+  <root level="info">
+    <appender-ref ref="STDOUT"/>
+  </root>
 
-	<logger name="org.sql2o" level="INFO" />
-	<logger name="com.netflix.conductor.dao.mysql" level="DEBUG" />
-	<logger name="org.flywaydb" level="WARN" />
-	<logger name="com.zaxxer.hikari" level="WARN" />
-	<logger name="ch.vorburger" level="WARN" />
+  <logger name="org.sql2o" level="INFO"/>
```
picaron: this can probably be removed

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,5 @@
-#Thu Mar 29 08:18:43 PDT 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.4-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.4-bin.zip
```
picaron: maybe this file should be left as is was before?

picaron:
@mashurex is it possible to squash your commits into a single one? 

once merged into the target repo, it would keep the commits history a bit cleaner

picaron:
@v1r3n are ok with merging this?

mashurex:
@picaron I squashed the changes to make the history cleaner.

picaron:
@mashurex btw, I don't see the sql2o dependency was removed from the `build.gradle`, that should be removed

you can also remove `  <logger name="org.sql2o" level="INFO"/>` from `logback_test.xml`

picaron:
@v1r3n I'm not familiar with the `dependencies.lock` how can that be updated? is there a command or we can just remove all references to sql2o from there?

mashurex:
@picaron and @v1r3n I removed sql2o references and updated the dependencies.lock as well. Travis builds 1018 and 1019 seem to fail in an unrelated test in the same way that build 1017 did which is from PR #514- so I think there's something currently broken on the dev branch.

picaron:
seems @mashurex has address all the latest comments

good to merge @v1r3n ?

Issue #415, opened 2018-02-07
When Workflow complete， call removeWorkflow at ExecutionDAO


grfeng:
All state inquiries, including Workflow, Task can use IndexDAO(Elasticsearch), Not dependent ExecutionDAO(Redis), Keep Redis clean?

v1r3n:
we are considering this in the upcoming versions.

PR #437, opened 2018-02-27
Run-time data is deleted after the workflow is completed

grfeng:
Run-time data is deleted after the workflow is completed and history is stored in the ES
#415


core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -345,6 +345,8 @@ public void completeWorkflow(Workflow wf) throws Exception {
         }
         Monitors.recordWorkflowCompletion(workflow.getWorkflowType(), workflow.getEndTime() - workflow.getStartTime(), wf.getOwnerApp());
         queueDAO.remove(deciderQueue, workflow.getWorkflowId());    //remove from the sweep queue
+
```
pctreddy: Can this be a configuration and runtime data deleted only if the configuration is set to that value ?
Conductor Config looks like a good place for it and it does not break the functionality where the users might be still using the runtime data from some other systems to query.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -505,15 +507,15 @@ public void updateTask(TaskResult result) throws Exception {
...
-        decide(workflowId);
-
         if (task.getStatus().isTerminal()) {
             long duration = getTaskDuration(0, task);
             long lastDuration = task.getEndTime() - task.getStartTime();
             Monitors.recordTaskExecutionTime(task.getTaskDefName(), duration, true, task.getStatus());
             Monitors.recordTaskExecutionTime(task.getTaskDefName(), lastDuration, false, task.getStatus());
         }
 
+        decide(workflowId);
```
pctreddy: Is this moved here to fix the  recording of the metrics ?

redis-persistence/src/main/java/com/netflix/conductor/dao/dynomite/RedisExecutionDAO.java
```diff
@@ -378,7 +378,7 @@ public Workflow getWorkflow(String workflowId, boolean includeTasks) {
 			throw new ApplicationException(Code.NOT_FOUND, "No such workflow found by id: " + workflowId);
 		}
 		Workflow workflow = readValue(json, Workflow.class);
```
pctreddy: Same as above can be driven by a config parameter from the Conductor config.

test-harness/src/test/java/com/netflix/conductor/tests/utils/MockIndexDAO.java
```diff
@@ -18,8 +18,8 @@
  */
 package com.netflix.conductor.tests.utils;
 
-import java.util.ArrayList;
-import java.util.List;
+import java.util.*;
```
pctreddy: One of the things that will be added to the Conductor code base in near future is `FindBugs` and `PMD`, which will help catch the smaller static analysis sooner. Until then please take out the import with *

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
grfeng: Ok, However it is available from Elasticsearch.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
grfeng: Yes, `decide(workflowId)` will be remove runtime data.

grfeng:
@pctreddy Any problem？

PR #447, opened 2018-03-07
Optionally allow non successful status codes to COMPLETE a HttpTask.

dbsr:
For certain scenarios we have a use case for a HttpTask which can return for instance a 404 response. We base our decision in the workflow on this status code (eg, create something when it does not exist yet).

This PR allows us to optionally alllow certain non 2xx status codes to NOT fail the workflow.


dbsr:
@pctreddy I'm sorry for contacting you like this but since you seem to be very active here I would like to know if there is anything I can do to get feedback for my PR?

PR #451, opened 2018-03-09
Removed the ResponseTimeout check during ack time and enabled during sweep

pctreddy:
This PR has changes that involve
- ResponseTimeOutFix
- Gradle version dependency files to externalize the dependencies
- Dependency locks, to help do a rollback in case of where `+` versions are used.


core/src/main/java/com/netflix/conductor/service/ExecutionService.java
```diff
@@ -174,20 +175,18 @@ public Task getPendingTaskForWorkflow(String taskReferenceName, String workflowI
...
-		return false;
-
+	/**
+	 * This method removes the task from the un-acked Queue
+	 *
+	 * @param taskId: the taskId that needs to be updated and removed from the unacked queue
+	 * @throws Exception In case of an error while getting the Task from the executionDao
+	 * @return: True in case of successful removal of the taskId from the un-acked queue
+	 */
+	public boolean ackTaskReceived(String taskId) throws Exception {
```
apanicker-nflx: It would be useful to log the `workerid` that sends the ack here, unless this is logged somewhere else.

cyzhao: Good point.

core/src/main/java/com/netflix/conductor/service/ExecutionService.java
```diff
@@ -174,20 +175,18 @@ public Task getPendingTaskForWorkflow(String taskReferenceName, String workflowI
 		return executor.getPendingTaskByWorkflow(taskReferenceName, workflowId);
 	}
 
-	public boolean ackTaskRecieved(String taskId, String consumerId) throws Exception {
-		Task task = getTask(taskId);
-
-		if (task != null) {
-			String queueName = QueueUtils.getQueueName(task);
-			if(task.getResponseTimeoutSeconds() > 0) {
```
cyzhao: Essentially this logic is removed, context on why we can remove it?

pctreddy: @cyzhao , the un-acked queue mechanism was used to re-queue the message in case of `ResponseTimeoutSeconds`. Unacked queue is only intended be only used to when the task is given out to the worker and worker acks back saying it has received and processing work.

Now since the getter of responeTimeoutSeconds has been fixed to return the actual value instead of 0, the Sweeper will evaluate and reschedule the task by requeueing it.

core/src/main/java/com/netflix/conductor/service/ExecutionService.java
pctreddy: The TaskResource endpoint has not changed and still accepts the consumerId as a query param. 

```
@POST
	@Path("/{taskId}/ack")
	@ApiOperation("Ack Task is recieved")
	@Consumes({ MediaType.WILDCARD })
	public String ack(@PathParam("taskId") String taskId, @QueryParam("workerid") String workerId) throws Exception {
		return "" + taskService.ackTaskReceived(taskId);
	}
```

The access logs will still show the task being acked and also the consumer id being used for the ack.

PR #477, opened 2018-04-04
python client improvements

booooh:
* Added ability to do long-polling via python client (/poll/batch)
* Added ability to specify taskDomain via python client
* minor pep8 cleanup

### DIFF
--- a/client/python/conductor/ConductorWorker.py
+++ b/client/python/conductor/ConductorWorker.py
@@ -22,7 +22,7 @@ import socket
 hostname = socket.gethostname()
 
 
-class ConductorWorker(object):
+class ConductorWorker:
     def __init__(self, server_url, thread_count, polling_interval, worker_id=None):
         wfcMgr = WFClientMgr(server_url)
         self.workflowClient = wfcMgr.workflowClient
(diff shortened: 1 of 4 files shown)


client/python/conductor/ConductorWorker.py
```diff
@@ -15,26 +15,26 @@
...
 from conductor.conductor import WFClientMgr
 from threading import Thread
 import socket
 
 hostname = socket.gethostname()
 
-class ConductorWorker:
-    def __init__(self, server_url, thread_count, polling_interval):
+
+class ConductorWorker(object):
```
zsimic: You don't need to explicitly inherit `object` in python 3, this only applies to python 2
Unless this is fixing a bug, I would leave it out.
This page has a nice explanation: https://stackoverflow.com/questions/4015417/python-class-inherits-object

client/python/conductor/ConductorWorker.py
```diff
@@ -48,12 +48,13 @@ def execute(self, task, exec_function):
...
-            polled = self.taskClient.pollForTask(taskType, hostname)
-            if(polled != None):
+            polled = self.taskClient.pollForTask(taskType, self.worker_id)
+            if polled is not None:
+                self.taskClient.ackTask(polled['taskId'], self.worker_id)
                 self.execute(polled, exec_function)
 
     def start(self, taskType, exec_function, wait):
-        print('Polling for task ' + taskType + ' at a ' + str(self.polling_interval) + ' ms interval with ' + str(self.thread_count) + ' threads for task execution, with worker id as ' + hostname)
+        print('Polling for task ' + taskType + ' at a ' + str(self.polling_interval) + ' ms interval with ' + str(self.thread_count) + ' threads for task execution, with worker id as ' + self.worker_id)
```
zsimic: It's more readable to use formatting:
```
'Polling for task %s at a %s ms interval with %s threads for task execution, with worker id as %s' % (taskType, self.polling_interval, self.thread_count, self.worker_id)
```

booooh:
Thanks for the review and approval.  Will make the requested changes, and update the pull request

booooh:
@zsimic I've made the fixes you requested, and updated the ConductorWorkerClient to pass the taskDomain if requested.  To verify the functionality I've also added a small pytest unit-test.

PR #503, opened 2018-04-24
avoid repeat task

huangyiminghappy:
when i use 
fork ->wait1,wait2->join(wait1,wait2) like this:
image

image

found the workflow's tasks is repeat. i found that when the wait task  is not completed, in workflowExecutor ,deciderService.decide repeat put the wait task into the tasksToBeRequeued in workflowexecutor. the result like below.
throuth the repeat task can not occur bug,but reduce performance.see the bottom picture ,
The red area could not have been executed as many times.
image

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
@@ -567,7 +567,11 @@ public class WorkflowExecutor {
             if (!tasksToBeRequeued.isEmpty()) {
                 addTaskToQueue(tasksToBeRequeued);
             }
-            workflow.getTasks().addAll(tasksToBeScheduled);
+
+            if(!workflow.getTasks().contains(tasksToBeScheduled))
+            {
+                workflow.getTasks().addAll(tasksToBeScheduled);
+            }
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -567,7 +567,11 @@ public boolean decide(String workflowId) throws Exception {
             if (!tasksToBeRequeued.isEmpty()) {
                 addTaskToQueue(tasksToBeRequeued);
             }
-            workflow.getTasks().addAll(tasksToBeScheduled);
+
+            if(!workflow.getTasks().contains(tasksToBeScheduled))
+            {
```
apanicker-nflx: Good catch!
I think a cleaner way would be 
```java
tasksToBeScheduled.stream()
                    .filter(task -> !workflow.getTasks().contains(task))
                    .forEach(workflow.getTasks()::add);
```

cyzhao: @huangyiminghappy please make the change according to Anoop's recommendation

apanicker-nflx:
This has been fixed as part of another change.

PR #528, opened 2018-05-07
Trim some overly broad dependencies.

gorzell:
It's a bit questionable whether the client should depend on the AWS SDK at all, however since it does lets limit the scope to just the bit that it uses.  Previously it was pulling in the *entire* sdk universe, in order to use one class.

Also, the eureka-client dependency in core is not directly used and should be removed.


gorzell:
Apparently the eureka client, while not used directly by core, did have transitive dependencies that were being used.  😞  I will have to clean that up tomorrow.

PR #531, opened 2018-05-08
Clean up inconsistency with tabs and spaces.

gorzell:
Some files had tabs while others had spaces.  I replaced all of the tabs with 4 spaces to make this consistent.


picaron:
this maybe a bit too big undertaking ... but I agree with @gorzell , whitespace usage should be consistent all around

I suggest adopting tabs for indentation and maybe we should add a coding guideline to the root of the repo

_hiding behind something to avoid rocks thrown at me_

gorzell:
I have been out of the java scene for a while, but it used to be that the most common style was 4-4-8 with spaces rather than tab characters.  I believe this was also the most common practice within Netflix.

Not sure about the too big an undertaking part.  If it is fixed once and there is a standard it should be done.  Also there may even be a gradle plugin to enforce the standard pre-commit.

picaron:
I used to prefer spaces but I've been consistently been using tabs for a 2+ years now and I find myself spending much less time fixing formatting issues. I also find Indentation errors are much easier to spot.

Having to modify all files just to fix just a formatting issue seems a bit overkill to me but it would certainly make working with the codebase a bit easier. 

As for plugin, in theory there are good idea but they are always case where the auto-formatting will do a bad job and format the code in an ugly way so I'm not too fond of this forced approach.

PR #532, opened 2018-05-08
Switch from "java" to "openjdk" base docker containers.

gorzell:
The [java](https://hub.docker.com/_/java/) images have been deprecated according to the site in favor of the [openjdk](https://hub.docker.com/_/openjdk/) images.

### DIFF
--- a/docker/server/Dockerfile
+++ b/docker/server/Dockerfile
@@ -1,7 +1,7 @@
 #
 # conductor:server - Netflix conductor server
 #
-FROM java:8-jre-alpine
+FROM openjdk:8-jre-slim
 
 MAINTAINER Netflix OSS <conductor@netflix.com>
 
--- a/docker/server/Dockerfile.build
+++ b/docker/server/Dockerfile.build
@@ -1,7 +1,7 @@
 #
 # conductor:server - Netflix conductor server
 #
-FROM java:8-jdk
+FROM openjdk:8-jdk
 
 MAINTAINER Netflix OSS <conductor@netflix.com>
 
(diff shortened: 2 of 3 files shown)


docker/server/Dockerfile
```diff
@@ -1,7 +1,7 @@
 #
 # conductor:server - Netflix conductor server
 #
-FROM java:8-jre-alpine
+FROM openjdk:8-jre-slim
```
picaron: would the alpine images work with conductor? I did some test on my own and build alpine images and they seem to work fine

gorzell: You probably could, but the latest Java 8 version doesn't appear to have an apline version on docker hub.  For this particular use case, I am not even sure that it would add any value.

picaron: actually yes there is `openjdk:8-jre-alpine` and `openjdk:8-jdk-alpine` (see https://hub.docker.com/_/openjdk/) ..

the added value is the very small image size, I did some tests, the alpine image I got was around 150Mb in size while the slim one around 250Mb.

gorzell: Sorry, I wasn't clear.  The tags exist, but they are for a much older release.

```
docker run --rm -it openjdk:8-jre-alpine java -version
openjdk version "1.8.0_151"
OpenJDK Runtime Environment (IcedTea 3.6.0) (Alpine 8.151.12-r0)
OpenJDK 64-Bit Server VM (build 25.151-b12, mixed mode)
```
vs
```
docker run --rm -it openjdk:8-jre-slim java -version
openjdk version "1.8.0_171"
OpenJDK Runtime Environment (build 1.8.0_171-8u171-b11-1~deb9u1-b11)
OpenJDK 64-Bit Server VM (build 25.171-b11, mixed mode)
```

The slim version is listed as 95MB compressed.  I think that the new version is worth any additional size that there may be.

gorzell:
This is in the `2.0` branch at this point.

PR #544, opened 2018-05-20
Added FAILED_WITH_TERMINAL_ERROR support to TaskResult and Task Status

pctreddy:


core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -59,6 +59,9 @@
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
+import static com.netflix.conductor.common.metadata.tasks.Task.Status.*;
+import static com.netflix.conductor.common.metadata.tasks.Task.Status.FAILED;
```
apanicker-nflx: This is already imported in the line above.

pctreddy: Thank you, fixed my IDE settings.

PR #549, opened 2018-05-22
Ability to disable workflows based on workflow definition configuration

pctreddy:


cyzhao:
@pctreddy can you resolve the conflict so that can be merged?

PR #550, opened 2018-05-24
Refactor Modules and bootstrap code to leverage Providers. Part 1

gorzell:
I think that this code is about 1/2 of the way to done and I would love feedback on the direction.  So far I have mostly just pulled code out and/or moved it around.  The next step is to replace existing code paths with those abstractions.  It may be better to do that in a separate PR, so I will branch from this to maintain that option.

# Goal
In an effort to help pave the way for gRPC, it is important make the composition of different configurations of implementations easier than it currently is.  Guice is a great help in this area, but it doesn't appear to be consistently used or always following best practices.

 I have started to refactor the bootstrap and configuration code to be less tightly coupled and Jersey server specific.

# Affected Areas
## Module Constructors
My reading of the documentation is that Guice Modules should more or less only be configuration as code (to install other modules and create bindings) and as a result they should not generally require constructor modules.  Specifically doing so makes installing a module within another module rather problematic.

There does appear to be a pattern to help cover this case, which is creating Provider classes that can be bound and cover most cases for which the constructor arguments were being used.

## Provider Creation/Encapsulation
I have gone through and pulled as much of the code related to bootstrapping instances of certain objects out into Provider classes.  I have not yet removed the original code, but rather only duplicated it.

This will facilitate moving much of this code out of the `server` sub project and into the actual sub project that it is bootstrapping, which will align dependencies. i.e. moving redis/dyn connection code into that module/sub project.

## Configuration
I move the `Configuration` implementation into `core` from `server` because it allowed me to create a pattern of extending it for implementation specific configurations, i.e. redis/dyn.  I also clarified the naming to better reflect how the implementation functions, which should pave the way for an eventual migration to Archaius.

## Bootstrap
This code is very tightly coupled to starting a jersey/jetty server and often configures portions of that in non-Guicy ways.  The next step is to try and refactor that away so there is a more modular and abstract way to compose the aspects of a server.  Specifically, we will want to be able to run both the gRPC and REST servers in the same JVM (or independently) once the gRPC implementation is completed.


redis-persistence/build.gradle
```diff
@@ -8,7 +8,6 @@ dependencies {
  	compile "com.netflix.dyno-queues:dyno-queues-redis:${revDynoQueues}"
 	compile "org.elasticsearch:elasticsearch:${revElasticSearch2}"
 
-	//In memory redis for unit testing
- 	testCompile "org.rarefiedredis.redis:redis-java:${revRarefiedRedis}"
-
+    //In memory
+    compile "org.rarefiedredis.redis:redis-java:${revRarefiedRedis}"
```
cyzhao: what is reason of changing from testCompile to compile?

redis-persistence/src/main/java/com/netflix/conductor/dao/DynoProxyProvider.java
```diff
@@ -0,0 +1,24 @@
...
+
+
+import com.netflix.conductor.dao.dynomite.DynoProxy;
+
+import javax.inject.Inject;
+import javax.inject.Provider;
+
+import redis.clients.jedis.JedisCommands;
+
+public class DynoProxyProvider implements Provider<DynoProxy> {
```
cyzhao: +1

redis-persistence/src/main/java/com/netflix/conductor/dao/DynoProxyProvider.java
```diff
@@ -0,0 +1,24 @@
+package com.netflix.conductor.dao;
+
+
+import com.netflix.conductor.dao.dynomite.DynoProxy;
+
+import javax.inject.Inject;
+import javax.inject.Provider;
+
+import redis.clients.jedis.JedisCommands;
+
```
cyzhao: @Singleton?

redis-persistence/src/main/java/com/netflix/conductor/dao/DynoQueueDAOProvider.java
```diff
@@ -0,0 +1,34 @@
...
+import com.netflix.conductor.core.config.Configuration;
+import com.netflix.conductor.dao.dynomite.queue.DynoQueueDAO;
+import com.netflix.dyno.connectionpool.HostSupplier;
+import com.netflix.dyno.queues.redis.DynoShardSupplier;
+
+import javax.inject.Inject;
+import javax.inject.Provider;
+
+import redis.clients.jedis.JedisCommands;
+
```
cyzhao: @Singleton?

redis-persistence/src/main/java/com/netflix/conductor/jedis/ConfigurationHostSupplierProvider.java
```diff
@@ -0,0 +1,60 @@
...
+
+    private List<Host> parseHostsFromConfig(DynomiteConfiguration configuration) {
+        String hosts = configuration.getHosts();
+        if(hosts == null) {
+            // FIXME This type of validation probably doesn't belong here.
+            String message = String.format(
+                    "Missing dynomite/redis hosts.  Ensure '%s' has been set in the supplied configuration.",
+                    DynomiteConfiguration.HOSTS_PROPERTY_NAME
+            );
+            System.err.println(message);
```
cyzhao: remove this

redis-persistence/src/main/java/com/netflix/conductor/jedis/ConfigurationHostSupplierProvider.java
```diff
@@ -0,0 +1,60 @@
...
+            logger.error(message);
+            System.exit(1);
```
cyzhao: throw exception here, calling System.exit() is not a good idea.

redis-persistence/build.gradle
gorzell: The mock that is used for local development requires this at compile time.

`com.netflix.conductor.jedis.JedisMock`

You can see the corresponding removal from the server module compile dependencies here: https://github.com/Netflix/conductor/pull/550/files#diff-2af3da569e27d5fad9d8c14613c35a66L42

redis-persistence/src/main/java/com/netflix/conductor/jedis/ConfigurationHostSupplierProvider.java
gorzell: Happy to, was trying to leave things mostly intact for the first pass.

redis-persistence/src/main/java/com/netflix/conductor/jedis/ConfigurationHostSupplierProvider.java
gorzell: I agree, just need to double check how Guice will react.

redis-persistence/src/main/java/com/netflix/conductor/dao/DynoQueueDAOProvider.java
gorzell: There is an interesting question whether that makes the provider a single or the result. I am not super clear reading the Guice docs how that will play out.  I *think* it is better to mark the class the is being provided as a singleton, if at all possible.  I'll look into this some more though.

redis-persistence/src/main/java/com/netflix/conductor/dao/DynoProxyProvider.java
gorzell: I think this should be covered by that fact that `DynoProxy` itself is a singleton.  Unless you think the `Provider` itself should be a `@Singleton` for some reason.

gorzell: I take it back, since `new` is being called directly think that breaks this.  But it does make me think that these two providers may be unnecessary.

redis-persistence/src/main/java/com/netflix/conductor/dao/DynoQueueDAOProvider.java
gorzell: The current bootstrap code seems to totally dodge this constructor 
```
@Inject
public DynoQueueDAO(DiscoveryClient dc, Configuration config) {
```
Which seems sub-optimal.  Do you know if there is a reason for that?

In the short term I can make the Provider itself a singleton and then ensure there is only one instance within it.  But that seems like more of a hack that could end badly.

core/src/main/java/com/netflix/conductor/core/config/Configuration.java
```diff
@@ -1,121 +1,137 @@
...
+    String ENVIRONMENT_DEFAULT_VALUE = "test";
+    String STACK_PROPERTY_NAME = "STACK";
+    String STACK_DEFAULT_VALUE = "test";
+    String APP_ID_PROPERTY_NAME = "APP_ID";
+    String APP_ID_DEFAULT_VALUE = "conductor";
+    String REGION_PROPERTY_NAME = "EC2_REGION";
+    String REGION_DEFAULT_VALUE = "us-east-1";
+    String AVAILABILITY_ZONE_PROPERTY_NAME = "EC2_AVAILABILITY_ZONE";
+    String AVAILABILITY_ZONE_DEFAULT_VALUE = "us-east-1c";
+    String ADDITIONAL_MODULES_PROPERTY_NAME = "conductor.additional.modules";
```
gudmundur: If all of these above are meant as constants, could you mark them as `static final`? Also, is the intention to have default scope on these? Would a `private` or `public` visibility make more sense?

redis-persistence/src/main/java/com/netflix/conductor/dyno/DynoProxyDiscoveryProvider.java
```diff
@@ -0,0 +1,33 @@
...
+
+    @Inject
+    public DynoProxyDiscoveryProvider(DiscoveryClient discoveryClient, Configuration configuration) {
+        this.discoveryClient = discoveryClient;
+        this.configuration = configuration;
+    }
+
+    @Override
+    public JedisCommands get() {
+        String cluster = configuration.getProperty("workflow.dynomite.cluster", null);
```
gudmundur: **Question**: Could we inject a more specific `Configuration`, so that `Configuration#getCluster()` can be used?

core/src/main/java/com/netflix/conductor/core/config/Configuration.java
```diff
@@ -1,121 +1,137 @@
...
+     * @return User defined string property.
+     */
+    public String getProperty(String name, String defaultValue);
+
+
+    /**
+     *
+     * @return Returns all the configurations in a map.
+     */
+    public Map<String, Object> getAll();
```
gudmundur: **Suggestion**: Are the three `get`ers above meant to be public? I find exposing `getProperty` and such for general purpose configuration to be an anti-pattern. It is harder to trace what configuration properties are being used, as you can't easily count references to the method. My recommendation would be to hide them and force specific getters to be added.

**Note**: I know this isn't added by this PR.

redis-persistence/src/main/java/com/netflix/conductor/dyno/DynomiteConfiguration.java
```diff
@@ -0,0 +1,65 @@
...
+    int MAX_CONNECTIONS_PER_HOST_DEFAULT_VALUE = 10;
+
+    String ROOT_NAMESPACE_PROPERTY_NAME = "workflow.namespace.queue.prefix";
+    String ROOT_NAMESPACE_DEFAULT_VALUE = null;
+
+    String DOMAIN_PROPERTY_NAME = "workflow.dyno.keyspace.domain";
+    String DOMAIN_DEFAULT_VALUE = null;
+
+    String NON_QUORUM_PORT_PROPERTY_NAME = "queues.dynomite.nonQuorum.port";
+    int NON_QUORUM_PORT_DEFAULT_VALUE = 22122;
```
gudmundur: Same comments for the constants here as further up.

redis-persistence/src/main/java/com/netflix/conductor/dyno/DynomiteConfiguration.java
```diff
@@ -0,0 +1,65 @@
...
+
+    default String getCluster() {
```
gudmundur: **Question**: Is the use of package scope here intended? The absence of `public`, `protected` and `private` leaves these only accessible within the package.

redis-persistence/src/main/java/com/netflix/conductor/dyno/RedisQueuesDiscoveryProvider.java
```diff
@@ -0,0 +1,105 @@
...
+
+public class RedisQueuesDiscoveryProvider implements Provider<RedisQueues> {
+
+    private static final Logger logger = LoggerFactory.getLogger(RedisQueuesDiscoveryProvider.class);
+    
+    private final DiscoveryClient discoveryClient;
+    private final DynomiteConfiguration configuration;
+
+    @Inject
+    RedisQueuesDiscoveryProvider(DiscoveryClient discoveryClient, DynomiteConfiguration configuration){
```
gudmundur: **Nit**: Space between `(` and `{`.

server/src/main/java/com/netflix/conductor/server/ConductorServer.java
```diff
@@ -124,15 +126,15 @@ private void init(String dynoClusterName, List<Host> dynoHosts) {
...
 			ConnectionPoolConfigurationImpl connectionPoolConfiguration = new ConnectionPoolConfigurationImpl(dynoClusterName)
 					.withTokenSupplier(getTokenMapSupplier(dynoHosts))
-					.setLocalRack(conductorConfig.getAvailabilityZone())
-					.setLocalDataCenter(conductorConfig.getRegion())
+					.setLocalRack(systemPropertiesConfiguration.getAvailabilityZone())
+					.setLocalDataCenter(systemPropertiesConfiguration.getRegion())
 					.setSocketTimeout(0)
 					.setConnectTimeout(0)
-					.setMaxConnsPerHost(conductorConfig.getIntProperty("workflow.dynomite.connection.maxConnsPerHost", 10));
+					.setMaxConnsPerHost(systemPropertiesConfiguration.getIntProperty("workflow.dynomite.connection.maxConnsPerHost", 10));
```
gudmundur: **Question**: Couldn't the `getMaxConnectionsPerHost()` be used here?

core/src/main/java/com/netflix/conductor/core/config/Configuration.java
gorzell: I agree, but will probably save that for a later PR.  Perhaps when switching to Archaius.

redis-persistence/src/main/java/com/netflix/conductor/dyno/DynomiteConfiguration.java
gorzell: Same answer.

redis-persistence/src/main/java/com/netflix/conductor/dyno/DynomiteConfiguration.java
gorzell: You might think that, but `interface`s are special, the default is `public` which is indeed what we want.

server/src/main/java/com/netflix/conductor/server/ConductorServer.java
gorzell: Not really, because this is the parent type `Configuration`, not `DynomiteConfiguration`.  In this location forcing `DynomiteConfiguration` seems like an overly strong dependency since you might not be using Dynomite at all.

The good news is that I think this is mostly addressed in my followup PR gorzell/conductor/pull/5.

core/src/main/java/com/netflix/conductor/core/config/Configuration.java
gudmundur: Thanks for clarifying this @gorzell. Must admit that my Java chops are a little rusty. It's been a while since my Java 6 days. 🙇

PR #556, opened 2018-05-25
fix usage of retried flag, added executed flag

apanicker-nflx:


core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -259,46 +257,42 @@ public void retry(String workflowId) throws Exception {
...
+                task.setRetried(true);
+                executionDAO.updateTask(task);
             }
         });
 
         scheduleTask(workflow, rescheduledTasks);
 
         workflow.setStatus(WorkflowStatus.RUNNING);
         executionDAO.updateWorkflow(workflow);
+        executionDAO.updateTasks(workflow.getTasks());
```
pctreddy: Why update the tasks here since the update already happens for each task in the `forEach` ?

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -329,6 +323,7 @@ public void completeWorkflow(Workflow wf) throws Exception {
         workflow.setStatus(WorkflowStatus.COMPLETED);
         workflow.setOutput(wf.getOutput());
         executionDAO.updateWorkflow(workflow);
+        executionDAO.updateTasks(wf.getTasks());
```
pctreddy: Same here why do all tasks need to be updated ?

core/src/main/java/com/netflix/conductor/core/execution/WorkflowSweeper.java
```diff
@@ -14,10 +14,20 @@
  * limitations under the License.
  */
 /**
- * 
+ *
  */
 package com.netflix.conductor.core.execution;
 
+import com.netflix.conductor.core.WorkflowContext;
+import com.netflix.conductor.core.config.Configuration;
```
pctreddy: Does this file have any change related to `executed` flag ?

redis-persistence/src/main/java/com/netflix/conductor/dao/dynomite/RedisExecutionDAO.java
```diff
@@ -525,6 +529,13 @@ private String insertOrUpdateWorkflow(Workflow workflow, boolean update) {
 
 	}
 
+    private void mapTaskToWorkflow(String taskId, String workflowInstanceId) {
```
pctreddy: Missing java doc.  This is not just a map operation, the correlation between the workflow and the tasks contained are updated in dynomite, consider changing the name of the method.  Adding a unit test whenever a new method is introduced is a good practice.

apanicker-nflx: added javadoc and test

core/src/main/java/com/netflix/conductor/core/execution/WorkflowSweeper.java
apanicker-nflx: no changes, just some code cleanup

common/src/main/java/com/netflix/conductor/common/metadata/tasks/Task.java
```diff
@@ -353,7 +348,20 @@ public void setRetried(boolean retried) {
     }
 
     /**
-     *
+     * @return True if the task has been finished execution and has been updated in the DAO
```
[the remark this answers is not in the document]
apanicker-nflx: updated description of the field

core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
```diff
@@ -151,22 +155,22 @@ private DeciderOutcome decide(final WorkflowDef def, final Workflow workflow, Li
...
                     outcome.tasksToBeUpdated.add(pendingTask);
                 }
             }
 
-            if (!pendingTask.isRetried() && pendingTask.getStatus().isTerminal()) {
-                pendingTask.setRetried(true);
-                List<Task> nextTasks = getNextTask(def, workflow, pendingTask);
-                nextTasks.forEach(rt -> tasksToBeScheduled.put(rt.getReferenceTaskName(), rt));
+            if (!pendingTask.isExecuted() && !pendingTask.isRetried() && pendingTask.getStatus().isTerminal()) {
+                pendingTask.setExecuted(true);
```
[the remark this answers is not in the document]
apanicker-nflx: updated the description for this field

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
apanicker-nflx: inside the foreach, only the cancelled tasks are updated. This is applied for all tasks and is used as a data enrichment operation (specific use case is for archived workflows which are being retried).

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
apanicker-nflx: In some cases, the tasks are not updated in the DS after completion which causes workflows to be stuck in RUNNING state. This fixes that case.

core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
```diff
@@ -151,22 +155,22 @@ private DeciderOutcome decide(final WorkflowDef def, final Workflow workflow, Li
...
+                List<Task> nextTasks = getNextTask(workflowDef, workflow, pendingTask);
+                nextTasks.forEach(nextTask -> tasksToBeScheduled.putIfAbsent(nextTask.getReferenceTaskName(), nextTask));
```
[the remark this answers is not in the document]
apanicker-nflx: Found that in some cases, duplicate(ghost) tasks were scheduled now that we update the datastore after every execution.

core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
```diff
@@ -131,11 +135,11 @@ private DeciderOutcome decide(final WorkflowDef def, final Workflow workflow, Li
             tasksToBeScheduled.put(pst.getReferenceTaskName(), pst);
         });
 
-        //A new workflow does not enter this code branch
+        // A new workflow does not enter this code branch
         for (Task pendingTask : pendingTasks) {
 
             if (SystemTaskType.is(pendingTask.getTaskType()) && !pendingTask.getStatus().isTerminal()) {
-                tasksToBeScheduled.put(pendingTask.getReferenceTaskName(), pendingTask);
+                tasksToBeScheduled.putIfAbsent(pendingTask.getReferenceTaskName(), pendingTask);
```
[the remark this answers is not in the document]
cyzhao: There should be an validation in place to enforce task reference name is unique within workflow def (if it is not there then it needs to be fixed), so this change is insignificant.

core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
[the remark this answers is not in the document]
cyzhao: what is cause of duplicated tasks for the same task ref name?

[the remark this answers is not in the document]
apanicker-nflx: For a FORK task, the JOIN task is created for each branch, and this JOIN task is replaced later in this method, when getNextTask is called after the last task execution in the fork

PR #569, opened 2018-06-05
fix for fork/join with ghost join tasks

apanicker-nflx:


core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
```diff
@@ -446,7 +449,10 @@ boolean isResponseTimedOut(TaskDef taskDefinition, Task task) {
         String taskId = IDGenerator.generate();
         TaskMapperContext taskMapperContext = new TaskMapperContext(workflowDefinition, workflowInstance, taskToSchedule,
                 input, retryCount, retriedTaskId, taskId, this);
-        return taskMappers.get(taskType.name()).getMappedTasks(taskMapperContext);
+        List<Task> tasks = taskMappers.get(taskType.name()).getMappedTasks(taskMapperContext);
+
+        return tasks.stream().filter(task -> executionDAO.getPendingTasksByWorkflow(task.getTaskType(), workflowInstance.getWorkflowId()).stream().
```
cyzhao: please add a comment explaining why we need to filter tasks here

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -503,6 +503,8 @@ public void updateTask(TaskResult result) throws Exception {
 
         decide(workflowId);
 
+        workflowInstance = executionDAO.getWorkflow(workflowId);
```
cyzhao: Keep in mind this could be a expensive call as by default it also loads all tasks, why is it needed here?

core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
cyzhao: getPendingTasksByWorkflow is also a very expensive call, it loads all IN_PROGRESS tasks of all workflow executions into the memory then filter by workflowId, please refactor it to get tasks for workflowId first then filter by IN_PROGRESS tasks.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
apanicker-nflx: This was leftover from debugging, I will remove this.

core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
apanicker-nflx: Good point, will refactor.

apanicker-nflx: added comment

PR #583, opened 2018-06-08
Risk of termination of scheduled tasks 

huangyiminghappy:
the schedule method must catch the exception to Preventing scheduled tasks from terminating,we have  encountered similar problems

PR #584, opened 2018-06-08
Fix for scheduling tasks after executing system tasks

apanicker-nflx:
Since tasks in progress for the workflow instance is checked every time a task is scheduled, system tasks do not need explicit handling of the `executed` flag, they will follow the same path as user defined tasks.


huangyiminghappy:
thank you for reviewing https://github.com/Netflix/conductor/pull/583

PR #585, opened 2018-06-08
Refactor Modules and bootstrap code to leverage Providers. Part 2

gorzell:
This is a follow up to Netflix/conductor#550

This completely refactors and updates how the application is bootstrapped.
* Create more modules to represent the various datastore configurations.
* Remove Module constructor arguments in favor of using Providers from the previous PRs. This is better in line with Guice best practices and recommendations.
* Move all of the logic related to composing Modules into ConductorServer and have it happen when the Injector is created. This is better in line with Guice best practices and recommendations.
* Add a configuration option to disable Jersey (Although this won't be completely implemented until we work in gRPC).
* Start to minimize redundant code between ServletContext and ConductorServer.
* Deprecate some Dyno constructors in favor of simpler injection based models, especially in the DynoQueueDAO.

PR #591, opened 2018-06-12
Re-factor MySQL bootstrapping and configuration.

gorzell:
This brings the MySQL module more in line with the Redis module in how it is configured and bootstrapped.  It also cleans up the package structure a bit alone the way to be a bit more explicit.

* Add a DataSourceProvider for MySQL.
* Consolidate configuration property names and default values into an interface.
* Move most of things that aren't actually DAO related out of the dao package.
* Move boolean property reading code into the base configuration interface.


gorzell:
@cyzhao This is independent of PR #585, but a requirement for the upcoming gRPC PR.

core/src/main/java/com/netflix/conductor/core/config/Configuration.java
```diff
@@ -118,6 +118,12 @@
     public String getProperty(String name, String defaultValue);
 
 
+    default boolean getBoolProperty(String name, boolean defaultValue) {
+        String value = getProperty(name, null);
+        if(null == value || value.trim().length() == 0){ return defaultValue; }
```
cyzhao: fix formatting

mysql-persistence/src/main/java/com/netflix/conductor/mysql/MySQLDataSourceProvider.java
```diff
@@ -0,0 +1,54 @@
...
+
+import org.flywaydb.core.Flyway;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.inject.Inject;
+import javax.inject.Provider;
+import javax.sql.DataSource;
+
+public class MySQLDataSourceProvider implements Provider<DataSource> {
```
cyzhao: @Singleton

cyzhao:
@gorzell please resolve the conflict.

cyzhao:
@gorzell  please resolve the conflict.

mysql-persistence/src/main/java/com/netflix/conductor/mysql/MySQLDataSourceProvider.java
gorzell: I don't think that marking the Provider as a singleton is helpful in this case, and I also don't think that having it provide a singleton is what we want.

If the concern is that the application uses a singleton, we should handle that in the binding configuration, not the provider.  The module does in fact bind this a singleton.  Is there a different behavior that you would like to see?

gorzell: https://github.com/Netflix/conductor/pull/591/files#diff-a46b98ee0f51900430a24ce12d0ccc8fR23

cyzhao: we are fine here since data source is bound as singleton in guice module.

PR #605, opened 2018-06-19
fix response timeout seconds behavior

apanicker-nflx:


core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
```diff
@@ -144,12 +144,10 @@ private DeciderOutcome decide(final WorkflowDef workflowDef, final Workflow work
...
 
             TaskDef taskDefinition = metadataDAO.getTaskDef(pendingTask.getTaskDefName());
-            if (taskDefinition != null) {//QQ what happens when the task definition is null at this time ??
+            if (taskDefinition != null) {
                 checkForTimeout(taskDefinition, pendingTask);
-                // If the task has not been updated for "responseTimeout" then rescheduled it.
-                if (isResponseTimedOut(taskDefinition, pendingTask)) {
-                    outcome.tasksToBeRequeued.add(pendingTask);
-                }
+                // If the task has not been updated for "responseTimeout" then reschedule it
```
cyzhao: comment is out updated

core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
```diff
@@ -399,16 +397,16 @@ void checkForTimeout(TaskDef taskDef, Task task) {
     }
 
     @VisibleForTesting
-    boolean isResponseTimedOut(TaskDef taskDefinition, Task task) {
+    void checkResponseTimeOut(TaskDef taskDefinition, Task task) {
```
cyzhao: better to move the logic that times out the task out of this method into its own method so the code is readable, we won't be able to read above method and know that the task could be timed out without diving into checkResponseTimeOut method, for example, the following code is more readable:

```
if (isResponseTimedOut(taskDefinition, pendingTask) {
   timeoutTask(pendingTask);
}
```

PR #606, opened 2018-06-19
fixed display of the workflows when filtering by status

alexnetflix:
UI does not filter recent workflows by 'RUNNING' status correctly

### DIFF
--- a/core/src/main/java/com/netflix/conductor/service/ExecutionService.java
+++ b/core/src/main/java/com/netflix/conductor/service/ExecutionService.java
@@ -18,11 +18,13 @@ package com.netflix.conductor.service;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 import javax.inject.Inject;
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/netflix/conductor/service/ExecutionService.java
```diff
@@ -307,23 +309,31 @@ public void removeWorkflow(String workflowId, boolean archiveWorkflow) throws Ex
...
 				logger.error(e.getMessage(), e);
 				return null;
 			}
 		}).filter(Objects::nonNull).collect(Collectors.toList());
-		int missing = result.getResults().size() - workflows.size();
+
+		//Search again to filter out some workflows that might have been in flux state before.
+		//It's possible that during the first search ES had some workflows that were already completed in dynamo but
+		//still had a 'RUNNING' state in ES
+		SearchResult<String> repeatedResult = indexer.searchWorkflows(query, freeText, start, size, sortOptions);
```
cyzhao: I don't think we need to search again, we just need to filter List<WorkflowSummary> workflows by RUNNING status before return the list back.

alexnetflix: This search method is generic, meaning that the user might be searching by any status and other fields, using query or Lucene-based text search, so it's hard to extract exact search parameters to compare them with the retrieved workflow objects

cyzhao: ah, that makes sense, didn't realize the method takes in a arbitrary query string.

cyzhao: Thinking more this, search it repeated may not solve the problem. Let's revisit the approach.

PR #608, opened 2018-06-20
Conductor GRPC API

vmg:
Hi everyone! Thanks for your patience while we prepared this branch!

Here's the first version of the implementation of a GRPC API for Conductor. This initial PR provides:

- ProtoBuf models for all the types exposed through the API
- ProtoBuf RPC schemas for the `Workflow`, `Metadata` and `Task` APIs
- A server-side GRPC implementation of the `Workflow`, `Metadata` and `Task` APIs that can be run concurrently with the Jersey-based HTTP/REST server.
- Java clients for the `Workflow`, `Metadata` and `Task` APIs
- End2End tests for these clients and the GRPC implementation
- An idiomatic Go GRPC client implementation for the `Worker` API

As you can see, there's a few outstanding things to implement (most notably a Java GRPC worker client), but we wanted to open the PR early to start gathering feedback.

### On ProtoBuf generation

Generating the ProtoBuf models for all the exposed POJOs has been quite a nightmare; there are many objects exposed through the API and they're all very complex. 😅

We ended up settling on a codegen based approach: all the relevant fields for POJOs in the `common` module have been annotated (marking the objects that need to be exported and their fields), and a `protogen` module contains a code generator that does the heavy lifting.

This approach is rather convenient because by simply annotating POJOs we can generate both the `.proto` schemas for each object _and_ the Java code to map between the immutable GRPC objects and the POJOs.

We took special care to ensure the resulting Java mapping code is very efficient and idiomatic (i.e. looks like it's written by hand). Here's what it looks like: https://github.com/vmg/conductor/blob/vmg/grpc-upstream/grpc/src/main/java/com/netflix/conductor/grpc/AbstractProtoMapper.java

Likewise, we took special care to ensure that the generated `.proto` schemas look exactly like they would if they were hand-written. Example: https://github.com/vmg/conductor/blob/vmg/grpc-upstream/grpc/src/main/proto/model/task.proto

Now, it's not clear to me whether you'd want to keep the `protogen` generation code in this repository. I've made sure to isolate all the `protogen` changes to individual, separate commits, so they can be stripped from the codebase with a simple rebase. I see several paths going forward:

- We keep the `protogen` module in the repo and use it to re-generate all the ProtoBuf schemas _and_ the mapping code when the POJOs change.
- We assume that the POJOs will never change and we remove the `protogen` module from the repository. Since the generated mapping code and schemas are idiomatic, we can manually update them in the future if we need to add/remove fields.
- We move `protogen` into a separate repository (e.g. in the github org) and add it as a dependency to the project. This is a lot more work, but it has the benefit of providing the community with a general solution to generate both ProtoBuf schemas _and_ mapping code from POJOs (there's nothing like that out there right now -- trust me I looked!).

Either way seems reasonable to me.

### On ProtoBuf3 Limitations

One of the "ugly" things in the existing implementation is caused by optional arguments in the GRPC API. The REST API has many endpoints with optional arguments (e.g. the `domain` or the `version` when looking up a task or taskdef), and there is no clean way to mimic this design using ProtoBuf3.

In Proto3, all fields in an object are optional. For Scalars like `int` or `string`, there is no real way to test whether a message contains such a field: a missing field is simply represented by the zero-value for the field (0 for ints, `""` for strings), and there is no way to tell from the Java API whether a message is missing that field, or whether the field has been explicitly set to 0 or `""`.

Fortunately, for all the cases I've mapped from the REST API, the "zero value" for a field is actually invalid, so we can check for the zero value and replace it with the default value in the GRPC server implementation. This is what we're doing right now, but I am worried that it may come back to bite us in the ass if in the future we need to distinguish between a valid zero value and a field that was explicitly _not_ set.

There are two ways to improve this situation:

- Switch the proto messages to Proto2, which explicitly supports optional and required values. I know from friends at Google that Proto2 is actively maintained because there are hundreds of internal Google APIs that _need_ optional fields, so they cannot migrate to Proto3. It is, however, a bit of a risky choice.

- Use wrapping messages for all the optional fields. The Proto3 "standard library" provides a set of wrapping messages that simply wrap an `int32` or `string` in a message. Message fields embedded in another message _can always be checked for existence, unlike scalars_. The Java API provides a `hasXXX` method for each such field. This would make the messages larger and the Java code more verbose and complex (both in client and server), but it does let us distinguish between missing fields and fields set to their zero values.

- Do nothing and ensure that the "zero value" for these optional fields is invalid, and hence can be assumed to be "unset".

Either approach has tradeoffs. I'd love to hear your thoughts on this.

### Reviewing this bunch'a'code

As you can see, this is a big PR. Fortunately, we've made sure to split it into independent, small commits that can be reviewed on their own. Of course we're willing to rebase and squash this as many times as you need to make your life easier, so just let us know what works better for you. The largest commits that you can find in the repository (the earliest ones), contain detailed commit messages that explain some of the technical choices we've made, so please check'em out.

That's basically it. Thank you for your time and willingness to merge this contributions. Looking forward to your feedback!


core/src/main/java/com/netflix/conductor/service/Lifecycle.java
```diff
@@ -0,0 +1,18 @@
+package com.netflix.conductor.service;
+
+public interface Lifecycle {
```
cyzhao: javadoc

core/src/main/java/com/netflix/conductor/service/Lifecycle.java
```diff
@@ -0,0 +1,18 @@
...
+        registerShutdownHook();
+    }
+
+    void stop() throws Exception;
+
+    default void registerShutdownHook() {
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+            try {
+                stop();
+            } catch (Exception e) {}
```
cyzhao: don't swallow exception

grpc-client/build.gradle
```diff
@@ -0,0 +1,21 @@
+plugins {
+    id 'java'
+}
+
+group 'com.netflix'
```
cyzhao: group should already be defined in root build.gradle

grpc-client/src/main/java/com/netflix/conductor/client/grpc/TaskClient.java
```diff
@@ -0,0 +1,234 @@
...
+     * @param startKey id of the task from where to return the results. NULL to start from the beginning.
+     * @param count    number of tasks to retrieve
+     * @return Returns the list of PENDING tasks by type, starting with a given task Id.
+     */
+    public List<Task> getPendingTasksByType(String taskType, @Nullable String startKey, @Nullable Integer count) {
+        Preconditions.checkArgument(StringUtils.isNotBlank(taskType), "Task type cannot be blank");
+
+        TaskServicePb.TasksInProgressRequest.Builder request = TaskServicePb.TasksInProgressRequest.newBuilder();
+        request.setTaskType(taskType);
+        if (startKey != null)
```
cyzhao: add missing curly braces for single statement blocks

grpc-server/src/main/java/com/netflix/conductor/grpc/server/service/EventServiceImpl.java
```diff
@@ -0,0 +1,98 @@
...
+import javax.inject.Inject;
+
+import io.grpc.stub.StreamObserver;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class EventServiceImpl extends EventServiceGrpc.EventServiceImplBase {
+    private static final Logger logger = LoggerFactory.getLogger(EventServiceImpl.class);
+    private static final ProtoMapper protoMapper = ProtoMapper.INSTANCE;
+    private static final GRPCHelper grpcHelper = new GRPCHelper(logger);
```
cyzhao: use all upper case names for constants

grpc-server/src/main/java/com/netflix/conductor/grpc/server/service/GRPCHelper.java
```diff
@@ -0,0 +1,116 @@
...
+     *  craft the 'grpc-status-detail-bin' trailer, in the same way that the GRPC server implementations
+     *  for Go and C++ craft and serialize the header. This will allow us to access the metadata
+     *  cleanly from Go and C++ clients by using the 'details' method which _has_ been implemented
+     *  in those two clients.
+     *
+     * @param t The exception to convert
+     * @return an instance of {@link StatusException} which will properly serialize all its
+     * headers into the response.
+     */
+    private StatusException throwableToStatusException(Throwable t) {
```
cyzhao: exception is better done via a gRPC interceptor

grpc-server/src/main/java/com/netflix/conductor/grpc/server/service/GRPCHelper.java
```diff
@@ -0,0 +1,116 @@
...
+        String[] frames = ExceptionUtils.getStackFrames(t);
+        Metadata metadata = new Metadata();
+        metadata.put(STATUS_DETAILS_KEY,
+                DebugInfo.newBuilder()
+                        .addAllStackEntries(Arrays.asList(frames))
```
cyzhao: usually not a good idea to return stack trace back to the client that exposes internal implementation details?

grpc/build.gradle
```diff
@@ -0,0 +1,43 @@
+buildscript {
+    dependencies {
+        classpath 'com.google.protobuf:protobuf-gradle-plugin:0.8.5'
```
cyzhao: define dependency versions in versionsOfDependencies.gradle.

grpc/src/main/java/com/netflix/conductor/grpc/AbstractProtoMapper.java
```diff
@@ -0,0 +1,1144 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import javax.annotation.Generated;
+
+@Generated("com.netflix.conductor.protogen.ProtoGen")
+public abstract class AbstractProtoMapper {
+    public EventExecutionPb.EventExecution toProto(EventExecution from) {
+        EventExecutionPb.EventExecution.Builder to = EventExecutionPb.EventExecution.newBuilder();
+        if (from.getId() != null) {
```
cyzhao: why is this null check needed? to simplify generator logic?

grpc/src/main/java/com/netflix/conductor/grpc/ProtoMapper.java
```diff
@@ -0,0 +1,92 @@
...
+import com.netflix.conductor.common.metadata.workflow.WorkflowTask;
+import com.netflix.conductor.proto.WorkflowTaskPb;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public final class ProtoMapper extends AbstractProtoMapper {
```
cyzhao: javadoc

grpc/src/main/java/com/netflix/conductor/grpc/ProtoMapper.java
```diff
@@ -0,0 +1,92 @@
...
+    public static final ProtoMapper INSTANCE = new ProtoMapper();
+
+    private ProtoMapper() {}
+
+    @Override
+    public Value toProto(Object val) {
```
cyzhao: javadoc

grpc/src/main/proto/grpc/event_service.proto
```diff
@@ -0,0 +1,57 @@
...
+import "google/protobuf/empty.proto";
+import "model/eventhandler.proto";
+
+option java_package = "com.netflix.conductor.grpc";
+option java_outer_classname = "EventServicePb";
+option go_package = "github.com/netflix/conductor/client/gogrpc/conductor/grpc";
+
+service EventService {
+    // POST /
+    rpc AddEventHandler(conductor.proto.EventHandler) returns (google.protobuf.Empty);
```
cyzhao: Always introduce Request and Response wrappers for all gRPC services so request/response can be extended in the future without breaking compatibility.

grpc/src/main/proto/grpc/event_service.proto
```diff
@@ -0,0 +1,57 @@
...
+
+    // PUT /
+    rpc UpdateEventHandler(conductor.proto.EventHandler) returns (google.protobuf.Empty);
+
+    // DELETE /{name}
+    rpc RemoveEventHandler(RemoveEventHandlerRequest) returns (google.protobuf.Empty);
+
+    // GET /
+    rpc GetEventHandlers(google.protobuf.Empty) returns (stream conductor.proto.EventHandler);
```
cyzhao: e.g. create a empty GetEventHandlersRequest, and GetEventHandlersResponse with EventHandler field.

grpc/src/main/proto/model/dynamicforkjointask.proto
```diff
@@ -0,0 +1,16 @@
+syntax = "proto3";
+package conductor.proto;
+
+import "google/protobuf/struct.proto";
+
+option java_package = "com.netflix.conductor.proto";
+option java_outer_classname = "DynamicForkJoinTaskPb";
+option go_package = "github.com/netflix/conductor/client/gogrpc/conductor/model";
+
+message DynamicForkJoinTask {
```
cyzhao: What is the reason for one proto definition per message?

protogen/src/main/java/com/netflix/conductor/protogen/Element.java
```diff
@@ -0,0 +1,111 @@
...
+import com.netflix.conductor.common.annotations.ProtoEnum;
+import com.netflix.conductor.common.annotations.ProtoMessage;
+import com.netflix.conductor.protogen.types.AbstractType;
+import com.netflix.conductor.protogen.types.MessageType;
+import com.squareup.javapoet.MethodSpec;
+import com.squareup.javapoet.TypeSpec;
+
+import java.util.*;
+
+public abstract class Element {
```
cyzhao: javadoc, what is this for? why is it in protogen package if it is manually created?

cyzhao: I got it now, protogen package is not for generated types, but rather for generator that generates proto files from existing  types, great idea!

protogen/src/main/java/com/netflix/conductor/protogen/ProtoGen.java
```diff
@@ -0,0 +1,104 @@
...
+    public static String PROTO_PACKAGE_NAME = "conductor.proto";
+    public static String PROTO_JAVA_PACKAGE_NAME = "com.netflix.conductor.proto";
+    public static String PROTO_GO_PACKAGE_NAME = "github.com/netflix/conductor/client/gogrpc/conductor/model";
+
+    public static String GENERATED_MAPPER_PACKAGE = "com.netflix.conductor.grpc";
+    public static String GENERATOR_NAME = "com.netflix.conductor.protogen.ProtoGen";
+
+    private List<File> files = new ArrayList<>();
+
+    public static void main(String[] args) throws Exception {
```
cyzhao: Is this integrated with build process, or executed manually?

cyzhao:
First of all, great work!

My answers to your questions:

1. I would recommend making the code generator generic and move it into its repo, create a gradle plugin that generates the code as part of build process as well. It can help others to convert existing REST APIs to gRPC, we have a need for this at Netflix as well.
2. Regarding optional field, let's not go back to proto2 and your current approach looks good to me, create a wrapping message only when necessary.

grpc/src/main/java/com/netflix/conductor/grpc/AbstractProtoMapper.java
vmg: It is not safe to call a setter on a ProtoBuf object builder with a `null` argument, so we have to explicitly check for all nullable types before we do the mapping.

grpc/src/main/proto/grpc/event_service.proto
vmg: That's a good point. I'll get to this today.

grpc/src/main/proto/model/dynamicforkjointask.proto
vmg: We found this was the easiest way to keep the generated code consistent and properly namespaced.

protogen/src/main/java/com/netflix/conductor/protogen/ProtoGen.java
vmg: It's run manually right now. I'll try to wire this up to run automatically as soon as I extract the ProtoGen code into its own repository.

grpc/build.gradle
vmg: Yes. 👍

grpc-server/src/main/java/com/netflix/conductor/grpc/server/service/GRPCHelper.java
```diff
@@ -0,0 +1,116 @@
...
+        logger.error("internal exception during GRPC request", t);
+        response.onError(throwableToStatusException(t));
+    }
+
+    void emptyResponse(StreamObserver<Empty> response) {
+        response.onNext(Empty.getDefaultInstance());
+        response.onCompleted();
+    }
+
+    String optional(@Nonnull String str) {
```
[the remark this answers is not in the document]
vmg: I don't think returning `Optional` makes sense here. This `optional` helper is meant to convert the behavior of a missing field in ProtoBuf (wherein `str` is non-null, but it is empty) into the behavior that the Conductor DAOs expect (that missing `String` arguments to function should be passed in as `null`). I guess this needs a JavaDoc to clarify it. On it!

grpc-server/src/main/java/com/netflix/conductor/grpc/server/service/GRPCHelper.java
```diff
@@ -0,0 +1,116 @@
...
+        String[] frames = ExceptionUtils.getStackFrames(t);
+        Metadata metadata = new Metadata();
+        metadata.put(STATUS_DETAILS_KEY,
+                DebugInfo.newBuilder()
+                        .addAllStackEntries(Arrays.asList(frames))
+                        .setDetail(ExceptionUtils.getMessage(t))
+                        .build()
+        );
+
+        return Status.INTERNAL
```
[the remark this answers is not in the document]
vmg: As the documentation states, we're explicitly handling all the "known error cases" such as invalid arguments and missing objects in the actual implementation, and returning those codes directly. This helper is meant to handle the remaining (internal) exceptions which we don't know how to map back to the client.

grpc-server/src/main/java/com/netflix/conductor/grpc/server/service/GRPCHelper.java
vmg: I don't have a strong opinion about this. I guess it depends on whether the client is trustworthy. Would you prefer I remove the stacktrace altogether?

grpc-server/src/main/java/com/netflix/conductor/grpc/server/service/GRPCHelper.java
vmg: I haven't found any good examples on this in the wild, so I decided to go with a helper method. Let me do some more research.

grpc-server/src/main/java/com/netflix/conductor/grpc/server/service/EventServiceImpl.java
vmg: The existing server implementation (i.e. in the Jersey module) already has a `logger` static field with a lowercase name, which is what I brought over here. Would you like me to rename these fields to uppercase and also the ones in the jersey module, for consistency?

grpc-server/src/main/java/com/netflix/conductor/grpc/server/service/GRPCHelper.java
vmg: To elaborate more on this: many (all?) of the methods in the DAOs we're using for the implementation are marked with `throws Exception`. This means that inside the `XXXServiceImplementation` for each method, we need an explicit `try {} catch` block, as it is not possible to add `throws Exception` to the actual RPC interface, as that would change the method's signature. Since we're handling the exception explicitly, I don't think it's possible to use an interceptor at all: we would need to re-throw the exception again, which would means changing the method's type signature.

Please advice.

grpc-server/src/main/java/com/netflix/conductor/grpc/server/service/EventServiceImpl.java
cyzhao: That would be great.

grpc-server/src/main/java/com/netflix/conductor/grpc/server/service/GRPCHelper.java
cyzhao: I need we need to fix the issue with "throws Exception", let's stick with what you have now and we will get back to this.

grpc/src/main/java/com/netflix/conductor/grpc/AbstractProtoMapper.java
cyzhao: Makes sense.

PR #610, opened 2018-06-21
feature/GLUE-1325 modified es5 persistence to use ES Rest instead of …

pixelthekid:
…transport


pixelthekid:
This was a mistake.  Was supposed to be a merge into a fork.  My appologies.

PR #615, opened 2018-06-21
filter out COMPLETED workflows from the returned lists of the RUNNING items

alexnetflix:
Because it's a server side pagination - this UI fix will cause some side effects: 
Even though it shows 100 total items at the bottom of the page, we'll only display  97 or so;
instead of total 1000, the user will only see 997 etc.

### DIFF
--- a/ui/src/components/workflow/executions/WorkflowList.js
+++ b/ui/src/components/workflow/executions/WorkflowList.js
@@ -234,15 +234,14 @@ const Workflow = React.createClass({
     const statusList = ['RUNNING','COMPLETED','FAILED','TIMED_OUT','TERMINATED','PAUSED'];
 
     //secondary filter to match sure we only show workflows that match the the status
-    //console.log(this.state);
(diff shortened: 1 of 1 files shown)


ui/src/components/workflow/executions/WorkflowList.js
```diff
@@ -231,6 +233,18 @@ const Workflow = React.createClass({
     const workflowNames = this.state.workflows?this.state.workflows:[];
     const statusList = ['RUNNING','COMPLETED','FAILED','TIMED_OUT','TERMINATED','PAUSED'];
 
+    //secondary filter to match sure we only show workflows that match the the status
+    //console.log(this.state);
+     var currentStatus = this.state.status;
```
cyzhao: what if user add multiple statuses to the filter?

alexnetflix: @cyzhao oh, didn't realize there is a way to do it on the UI side, will recheck

PR #617, opened 2018-06-22
Parameters utils error

huangyiminghappy:
documentContext.read may occur exception,so i think we need to catch the exception.


core/src/main/java/com/netflix/conductor/core/execution/ParametersUtils.java
```diff
@@ -214,7 +213,11 @@ private Object replaceVariables(String paramString, DocumentContext documentCont
                     }
 
                 } else {
-                    convertedValues[i] = documentContext.read(paramPath);
+                    try {
+                        convertedValues[i] = documentContext.read(paramPath);
+                    } catch (Exception e) {
```
cyzhao: Add warning level logging

apanicker-nflx:
Please resolve merge conflicts and add tests.

core/src/main/java/com/netflix/conductor/core/execution/ParametersUtils.java
huangyiminghappy: thank you review, please review the https://github.com/Netflix/conductor/pull/988 , if it ok ,current pr will close

huangyiminghappy:
> Please resolve merge conflicts and add tests.

thank you review, please review the #988 , if it ok ,current pr will close

kishorebanala:
Reopened with #988

PR #619, opened 2018-06-22
Refactor Elastic Search bootstrap and remove ES 2.x related code.

gorzell:
**This depends on #608 and will be pretty hard to view before that is merged.**

This PR brings the ES related code in line with the other boot strap changes and ensure that if the embedded server is used it will be started with the other objects that have a life cycle in the `Main.class`.  It also removes all of the code related to ES 2.x in favor of ES 5.x.  Both had been supported for compatibility sake, but it made both the code base, bootstrap and build significantly more complicated.  Since Conductor 2.x isn't meant to be backwards compatible this seemed like a good opportunity to clean that up.

- [x] Remove es2-persistence module.
- [x] Create an Elastic Search specific configuration.
- [x] Move the embedded bootstrapping from the Module loading to the Main class.
- [x] Clean up client provider/Module to be more aligned with the other modules.


elasticsearch/src/main/java/com/netflix/conductor/elasticsearch/EmbeddedElasticSearch.java
```diff
@@ -0,0 +1,47 @@
...
+    String DEFAULT_HOST = "127.0.0.1";
+    String DEFAULT_SETTING_FILE = "embedded-es.yml";
+
+    default void cleanDataDir(String path) {
+        try {
+            logger.info("Deleting contents of data dir {}", path);
+            File f = new File(path);
+            if (f.exists()) {
+                FileUtils.cleanDirectory(new File(path));
+            }
```
vmg: I think this is a bug in the original code, but we want `FileUtils.cleanDirectory(f);` here, instead of opening the file again. The current implementation, besides redundant, could be racy!

es5-persistence/src/main/java/com/netflix/conductor/elasticsearch/es5/EmbeddedElasticSearchV5.java
```diff
@@ -0,0 +1,123 @@
...
+
+    @Override
+    public void start() throws Exception {
+        start(DEFAULT_CLUSTER_NAME, DEFAULT_HOST, DEFAULT_PORT, true);
+    }
+
+    public synchronized void start(String clusterName, String host, int port, boolean enableTransportClient) throws Exception {
+
+        if (instance != null && !instance.isClosed()) {
+            logger.info("Elastic Search is already running on port {}", getPort());
```
vmg: I'm not sure this log line accomplishes much. It tries to be helpful, but seems to be masking a programming error -- calling `start` twice. Maybe we should `throw` instead, although I don't love the idea.

es5-persistence/src/main/java/com/netflix/conductor/elasticsearch/es5/EmbeddedElasticSearchV5.java
```diff
@@ -0,0 +1,123 @@
...
+                .put("http.enabled", true)
+                .put("http.type", "netty4")
+                .put("transport.type", "netty4");
+
+        return settingsBuilder.build();
+    }
+
+    public Client getClient() {
+        if (instance == null || instance.isClosed()) {
+            logger.error("Embedded ElasticSearch is not Initialized and started, please call start() method first");
```
vmg: This really ought to be a `throw`, right? Returning `null` here is only going to make debugging the issue harder, as the server will crash with a `NullPointerException` further down the line.

es5-persistence/src/main/java/com/netflix/conductor/elasticsearch/ElasticSearchRestClientProvider.java
```diff
@@ -0,0 +1,34 @@
...
+    public RestHighLevelClient get() {
+        RestClient lowLevelRestClient = RestClient.builder(convertToHttpHosts(configuration.getURIs())).build();
+        return new RestHighLevelClient(lowLevelRestClient);
+    }
+
+    private HttpHost[] convertToHttpHosts(List<URI> hosts) {
+        List<HttpHost> list = hosts.stream().map(host ->
+                new HttpHost(host.getHost(), host.getPort()))
+                .collect(Collectors.toList());
+        return list.toArray(new HttpHost[0]);
```
vmg: TIL about this gorgeous API to workaround type erasure. Oh Java, you so crusty. 😍

elasticsearch/src/main/java/com/netflix/conductor/elasticsearch/EmbeddedElasticSearch.java
gorzell: I don't there is a potential race, but it is still worth cleaning up.  File objects are immutable and while some of the operations have side affects, they don't change anything about the path/object itself.

es5-persistence/src/main/java/com/netflix/conductor/elasticsearch/es5/EmbeddedElasticSearchV5.java
gorzell: As far as I can tell this code isn't even used, so maybe it would be better to just delete it?

es5-persistence/src/main/java/com/netflix/conductor/elasticsearch/es5/EmbeddedElasticSearchV5.java
gorzell: I am on the fence about this as well, but I am leaning toward throwing.

gorzell:
@cyzhao I think this should be ready to go now that #608 has been merged.

vmg:
As soon as this branch is merged I have another PR prepared that removes the `protogen` code from this repository and adds it as a compile-time dependency. 👌

Issue #576, opened 2018-06-07
v1.10.0 BACKEND_ERROR - Deadlock found when trying to get lock; try restarting transaction


juaby:
com.netflix.conductor.core.execution.ApplicationException: BACKEND_ERROR - Deadlock found when trying to get lock; try restarting transaction
	at com.netflix.conductor.dao.mysql.MySQLBaseDAO.getWithTransaction(MySQLBaseDAO.java:103)
	at com.netflix.conductor.dao.mysql.MySQLBaseDAO.withTransaction(MySQLBaseDAO.java:125)
	at com.netflix.conductor.dao.mysql.MySQLExecutionDAO.updateLastPoll(MySQLExecutionDAO.java:451)
	at com.netflix.conductor.service.ExecutionService.poll(ExecutionService.java:133)
	at com.netflix.conductor.server.resources.TaskResource.batchPoll(TaskResource.java:115)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at com.sun.jersey.spi.container.JavaMethodInvokerFactory$1.invoke(JavaMethodInvokerFactory.java:60)
	at com.sun.jersey.server.impl.model.method.dispatch.AbstractResourceMethodDispatchProvider$TypeOutInvoker._dispatch(AbstractResourceMethodDispatchProvider.java:185)
	at com.sun.jersey.server.impl.model.method.dispatch.ResourceJavaMethodDispatcher.dispatch(ResourceJavaMethodDispatcher.java:75)
	at com.sun.jersey.server.impl.uri.rules.HttpMethodRule.accept(HttpMethodRule.java:302)
	at com.sun.jersey.server.impl.uri.rules.RightHandPathRule.accept(RightHandPathRule.java:147)
	at com.sun.jersey.server.impl.uri.rules.ResourceClassRule.accept(ResourceClassRule.java:108)
	at com.sun.jersey.server.impl.uri.rules.RightHandPathRule.accept(RightHandPathRule.java:147)
	at com.sun.jersey.server.impl.uri.rules.RootResourceClassesRule.accept(RootResourceClassesRule.java:84)
[...]

PR #621, opened 2018-06-26
Fixed Deadlock issue with MySQL database

jvemugunta:
https://github.com/Netflix/conductor/issues/576

As per the issue mentioned, I have fixed the deadlock issue.
I wrote a multi threaded test case with 10 concurrent requests. Please review and let me know if code can be merged. I have changed the below two files. All of the tests succeed after these changes

https://github.com/jvemugunta/conductor/blob/4b721c5b42177f068da88ff3d70abacb64db8d61/mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLExecutionDAO.java#L811-L818

https://github.com/jvemugunta/conductor/blob/9a378b70164f401e8c496c7bf2e1931836a7627f/mysql-persistence/src/test/java/com/netflix/conductor/dao/mysql/MySQLExecutionDAOTest.java#L144-L180

### DIFF
--- a/mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLExecutionDAO.java
+++ b/mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLExecutionDAO.java
@@ -1,5 +1,17 @@
 package com.netflix.conductor.dao.mysql;
 
+import java.sql.Connection;
+import java.text.SimpleDateFormat;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.Date;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import javax.inject.Inject;
+import javax.sql.DataSource;
+
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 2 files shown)


mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLExecutionDAO.java
```diff
@@ -805,11 +807,15 @@ private void insertOrUpdatePollData(Connection connection, PollData pollData, St
 
         if (result == 0) {
             String INSERT_POLL_DATA = "INSERT INTO poll_data (queue_name, domain, json_data) VALUES (?, ?, ?)";
+            
+            synchronized(this)
```
cyzhao: This is not very efficient, can you try out upsert approach - https://dev.mysql.com/doc/refman/5.7/en/insert-on-duplicate.html?

jvemugunta: Let me try this approach. It seems like a good idea

PR #635, opened 2018-07-01
Terminal error output fix

pctreddy:


core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -472,13 +472,21 @@ public void updateTask(TaskResult taskResult) throws Exception {
...
             //In case of a FAILED_WITH_TERMINAL_ERROR the workflow will be terminated and the output of the task is never copied
             //ensuring the task output is copied to the workflow here
             if (FAILED_WITH_TERMINAL_ERROR.equals(task.getStatus())) {
+                //Update the task in the workflow instance
+                Task taskByRefName = workflowInstance.getTaskByRefName(task.getReferenceTaskName());
+                taskByRefName.setStatus(task.getStatus());
+                taskByRefName.setOutputData(task.getOutputData());
+                taskByRefName.setReasonForIncompletion(task.getReasonForIncompletion());
+                taskByRefName.setWorkerId(task.getWorkerId());
+                taskByRefName.setCallbackAfterSeconds(task.getCallbackAfterSeconds());
```
cyzhao: Task is already updated at line 473, why is it still needed here?

pctreddy: The task is saved to the data store on line 473, the ephemeral state of the workflow needs to be updated to calculate the workflow output. This is needed to determine the workflow output.

PR #636, opened 2018-07-02
grpc: Add type-safe ProtoBuf objects to all Tasks

vmg:
This WIP PR attempts two things:

1. As requested earlier, the ProtoGen code has been moved into its own OSS repository (https://github.com/vmg/protogen) and removed from this repository.

2. During the move, the ProtoGen code has been updated and now supports serialization of external PB types in the API.

Point 2 is particularly important since it allows us to add `messageInput` and `messageOutput` fields to all tasks and task definitions. Some notes on these fields:

- The goal of these fields is providing a type-safe way to pass input and input metadata through tasks that use the GRPC API.

- To accomplish this, we're giving these fields the `Any` ProtoBuf type. In PB, an `Any` message can store any arbitrary message type in a type-safe way, without the server needing to know the exact serialization format of the message.

- The new `messageInput` and `messageOutput` fields are also exported through the REST/JSON API. Commit f2bb020 includes extensive documentation on how is this accomplished exactly.

- We have chosen **not** to replace the input and output arguments, which right now are maps from `String` to `Object`: these fields are being used extensively by Conductor to pass around metadata for the tasks, and it is not feasible to deprecate them. We think it makes sense to keep these fields as-is and add the option to pass a ProtoBuf message via a separate field.

### DIFF
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -1,7 +1,6 @@
 buildscript {
     repositories {
         jcenter()
-        maven { url "https://dl.bintray.com/vmg/protogen" }
     }
     dependencies {
         classpath "com.github.vmg.protogen:protogen-codegen:${revProtogenCodegen}"
--- a/build.gradle
+++ b/build.gradle
@@ -33,7 +33,6 @@ subprojects {
 
     repositories {
         jcenter()
-        maven { url "https://dl.bintray.com/vmg/protogen" }
     }
 
     dependencies {


vmg:
Updated the protogen code so it actually runs as part of the build. 👌

common/build.gradle
```diff
@@ -1,4 +1,9 @@
 dependencies {
     compile "com.github.rholder:guava-retrying:${revGuavaRetrying}"
     compile "org.slf4j:slf4j-api:${revSlf4j}"
+    compile "com.google.protobuf:protobuf-java:${revProtoBuf}"
+    compile "com.fasterxml.jackson.core:jackson-databind:${revJacksonDatabind}"
+    compile "com.fasterxml.jackson.core:jackson-core:${revJacksonCore}"
+    compile "javax.inject:javax.inject:1"
```
cyzhao: revJavaxInject

common/src/main/java/com/netflix/conductor/common/utils/JsonMapperProvider.java
```diff
@@ -0,0 +1,141 @@
...
+         * not able to deserialize it on the server. We simply return a dictionary with
+         * '@type' and '@value' keys, where '@type' is identical to the canonical representation,
+         * but '@value' contains a base64 encoded string with the binary data of the serialized
+         * message.
+         *
+         * Since all the provided Conductor clients are required to know this encoding, it's always
+         * possible to re-build the original {@link Any} message regardless of the client's language.
+         *
+         * {@see AnyDeserializer}
+         */
```
cyzhao: Nice javadoc!

common/src/main/java/com/netflix/conductor/common/utils/JsonMapperProvider.java
```diff
@@ -0,0 +1,141 @@
...
+         * into its original form.
+         *
+         * {@see AnySerializer} for details on this representation.
+         */
+        protected class AnyDeserializer extends JsonDeserializer<Any> {
+            @Override
+            public Any deserialize(JsonParser p, DeserializationContext ctxt)
+                    throws IOException, JsonProcessingException {
+                JsonNode root = p.getCodec().readTree(p);
+                JsonNode type = root.get("@type");
```
cyzhao: extract constant

common/src/main/java/com/netflix/conductor/common/utils/JsonMapperProvider.java
```diff
@@ -0,0 +1,141 @@
...
+                JsonNode value = root.get("@value");
+
+                if (type == null || !type.isTextual())
```
cyzhao: add curly braces

grpc/src/main/java/com/netflix/conductor/grpc/AbstractProtoMapper.java
```diff
@@ -49,7 +50,7 @@
 import java.util.stream.Collectors;
 import javax.annotation.Generated;
 
-@Generated("com.netflix.conductor.protogen.ProtoGen")
+@Generated("ProtoGen")
```
cyzhao: com.github.vmg.protogen.ProtoGen

protogen/build.gradle
```diff
@@ -1,5 +1,10 @@
 dependencies {
     compile project(':conductor-common')
-    compile 'com.squareup:javapoet:1.11.1'
-    compile 'com.github.jknack:handlebars:4.0.6'
+    compile "com.github.vmg.protogen:protogen-codegen:${revProtoGen}"
+}
+
+task generate(type: JavaExec) {
```
cyzhao: What I envision is to have a protogen Gradle plugin that scans classes annotated with ProtoMessage and generates proto definitions and mappers automatically, it should be configured to run before generateProto task (from protobuf-gradle-plugin), we don't have to do this now, but at least we should probably get rid of ConductorProtoGen and have a main class in vmg.protogen that uses Reflections to get classes annotated with @ProtoMessage within a package (passed in as an argument) to process.

client/src/main/java/com/netflix/conductor/client/http/TaskClient.java
```diff
@@ -225,6 +225,7 @@ public void updateTask(TaskResult taskResult, String taskType) {
                 taskResult.setReasonForIncompletion(String.format("The TaskResult payload: %d is greater than the permissible 3MB", taskResultSize));
                 taskResult.setStatus(TaskResult.Status.FAILED_WITH_TERMINAL_ERROR);
                 taskResult.setOutputData(null);
+                taskResult.setOutputMessage(null);
```
cyzhao: please also update the logic that limit size of workflow input, task output and task result to consider size of inputMessage and outputMessage as well.

vmg: @cyzhao: Since we're passing `taskResult` to the `countingOutputStream`, the size of the new `Any` fields is already being taken into account, as the `ObjectMapper` instance which we're injecting takes care of serializing those two fields.

protogen/build.gradle
vmg: This was exceedingly difficult to get right. See: https://github.com/Netflix/conductor/pull/636/commits/46ed3b17856a691897a6067387a5df5664d266ce

grpc/src/main/java/com/netflix/conductor/grpc/AbstractProtoMapper.java
vmg: Done.

common/src/main/java/com/netflix/conductor/common/utils/JsonMapperProvider.java
vmg: Done!

vmg:
Alright, I finally figured out how to implement ProtoGen as a gradle plugin:

https://github.com/vmg/conductor/blob/46ed3b17856a691897a6067387a5df5664d266ce/common/build.gradle#L21-L34

I think this is as good as we're gonna get, but I'd appreciate feedback from any resident Gradle experts. Besides that I tackled all your other comments in the code review, @cyzhao. Let me know if we're missing something!

cyzhao:
@vmg can you take a look at the build issue?

vmg:
@cyzhao: Fixed!

PR #654, opened 2018-07-13
Feature/add auth header forwarding

lordbender:
At Florida Blue, we have a need to ensure that JWT tokens, passed to the Express Server are passed downstream to the Spring API. This is a requirement of our Security Model.

This PR includes the filter needed to take the Authorization header that is on the inbound request, and put that onto req.token, if it is present. Later, this is passed to the superagent wrapper class HttpClient's methods as an argument so it can be added to the outbound http calls on their way to the java api.

Further, I corrected the failing test script by adding at least one passing test.

Please let me know if you have any questions. 

The changes I put up ought not affect any of the existing behavior. I tested this extensively. They should simply add the ability to have a JWT, Bearer or Basic auth header that arrives at the Express Proxy forwarded on the the Java server.


lordbender:
@hunterford I apologize for the formatting. I had my format on save turned on for the first few commits. I have disabled it. I'll make sure to make the meaningful changes and formatting changes in separate commits in the future.

Thanks for taking the time to go through my code.

PR #662, opened 2018-07-17
Fix/code consistency issues

lordbender:
Adding linting files caused watch to fail for linting errors. We need to disable linting on watch until we get all of the linting errors corrected.


lordbender:
@hunterford please take a look when you get a minute. There was an unexpected side effect to the watch script after adding the eslint file.

hunterford:
It looks like lordbender:fix/cleaning-up-imports includes most of these changes plus some more fixes. So if it's okay with you, I will decline this pull request.

lordbender:
Sounds Good.

PR #669, opened 2018-07-20
Embedded ElasticSearch runtime fix

josedab:
When running `conductor` on branch `2.0` locally (without docker), there is an issue with the embedded Elasticsearch instance.

#  Problem
The code assumes it can add an index to the search instance at any time (ElasticSearchDAOV5). At startup time, this class is loaded and part of the constructor code calls to ElasticSearch for index creation (if necessary). Later in time, the class `Main` checks for the possibility of having an embedded instance for ES, starting it if that is the case.
If ES is run externally, the issue cannot be replicated as the instance will be running. 
However, if the instance is embedded, it will end up on server error **not being able to perform reads or writes to the index**.

# Solution
I opted for moving the creation of the index out of the constructor of `ElasticSearchDAOV5` and isolating it in its own `setup` method. In my opinion, the constructor should not deal with calls to external systems (SRP).
On top of that, I chose to use ElasticSearchConfiguration instead of Config when possible as this option will default values in case they are not provided (example: indexName).
I also added an extra call to ES checking for the health status of the system and wait until the status is `green`. This happens during the new method `setup`.

# Other thoughts
Embedded ElasticSearch has been deprecated since version 5: For embedded instances, there is no way to add a listener to know when the instance has started. A solution to fix that is having a Thread sleeping for x amount of seconds, but imo is a bit hacky. My recommendation would be to deprecate the ES5 embedded module. Any thoughts on this?


core/src/main/java/com/netflix/conductor/dao/IndexDAO.java
```diff
@@ -26,159 +26,161 @@
...
-	 * @param lastModifiedHoursAgoTo - last updated date should be lastModifiedHoursAgoTo hours ago or earlier
-	 * 	 *
-	 * @return  List of workflow Ids matching the pattern
-	 */
-	List<String> searchRecentRunningWorkflows(int lastModifiedHoursAgoFrom, int lastModifiedHoursAgoTo);
+    /**
+     * This method should return an unique identifier of the indexed doc
+     * @param workflow Workflow to be indexed
+     *
+     */
```
cyzhao: fix javadoc

cyzhao:
@josedab looks good to me please resolve the conflict, and fix failed check. Eventually we will make ES an optional dependency.

es5-persistence/src/main/java/com/netflix/conductor/dao/es5/index/ElasticSearchDAOV5.java
```diff
@@ -133,20 +132,11 @@
...
-
-            initIndex();
-            updateIndexName(config);
-            Executors.newScheduledThreadPool(1).scheduleAtFixedRate(() -> updateIndexName(config), 0, 1, TimeUnit.HOURS);
-
-        } catch (Exception e) {
-            logger.error(e.getMessage(), e);
-        }
+        this.indexName = config.getIndexName();
+        this.logIndexPrefix = config.getProperty("workflow.elasticsearch.tasklog.index.name", "task_log");
```
mudit3774: Consider moving strings to local constants. It might help in readability and further changes will create single line diff.

es5-persistence/src/main/java/com/netflix/conductor/dao/es5/index/ElasticSearchDAOV5.java
```diff
@@ -158,19 +148,42 @@ public ElasticSearchDAOV5(Client elasticSearchClient, Configuration config, Obje
...
+    public void setup() throws Exception {
+
+        elasticSearchClient.admin().cluster().prepareHealth().setWaitForGreenStatus().execute().get();
 
         try {
-            elasticSearchClient.admin().indices().prepareGetIndex().addIndices(logIndexName).execute().actionGet();
+
+            initIndex();
+            updateIndexName();
+            Executors.newScheduledThreadPool(1).scheduleAtFixedRate(() -> updateIndexName(), 0, 1, TimeUnit.HOURS);
```
mudit3774: Can you please help me understand if there is any specific advantage of using newScheduledThreadPool(1) over https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Executors.html#newSingleThreadExecutor() . 

Also, please consider moving delay and interval values to constants. Again, this would result in changing the configurations easily and having smaller diffs going forward.

es5-persistence/src/main/java/com/netflix/conductor/dao/es5/index/ElasticSearchDAOV5.java
```diff
@@ -158,19 +148,42 @@ public ElasticSearchDAOV5(Client elasticSearchClient, Configuration config, Obje
...
+
+        } catch (Exception e) {
+            logger.error(e.getMessage(), e);
+        }
+
+        //1. Create the required index
+        try {
```
mudit3774: Consider moving this to a separate function. In my opinion, it should help in making the code more testable, increase readability and would make it easy to change later.

es5-persistence/src/main/java/com/netflix/conductor/dao/es5/index/ElasticSearchDAOV5.java
josedab: Sounds good.
Note this PR just moved some code around in order to allow setting up the indexDao (if necessary). This improvement can actually be done over the current code (branch 2.0). Having said that, I think it will just improve the readability as you said, so will update the PR probably today 👍

es5-persistence/src/main/java/com/netflix/conductor/dao/es5/index/ElasticSearchDAOV5.java
josedab: Will move the delay and interval values to constants.
About your question, not sure If I am the right person to answer it. This code is already on branch 2.0, this PR is just moving code around in order to support a setup for IndexDao.
I can also change it, dont see why not use `newSingleThreadExecutor`, but I need to read over documentation and might delay my code update.

core/src/main/java/com/netflix/conductor/dao/IndexDAO.java
josedab: will do 👍

es5-persistence/src/main/java/com/netflix/conductor/dao/es5/index/ElasticSearchDAOV5.java
josedab: I will see what I can do. I wanted to keep the PR simple just updating the minimum parts so the contributors were free to refactor and not having to rebase several times if they were working on some feature related to elastic search.

cyzhao:
@josedab please resolve conflict and fix the build.

josedab:
@cyzhao Apologies for the delay in answering. Will try to get them done today or tomorrow 👍

PR #679, opened 2018-07-25
API to delete workflow definition

falu2010-netflix:
Description:

API to delete workflow definition. 

Sample curl call:

```curl -X DELETE "http://localhost:8080/metadata/workflow/kitchensinkTest?version=1" -H "accept: application/json"```

### DIFF
--- a/client/go/conductorhttpclient.go
+++ b/client/go/conductorhttpclient.go
@@ -20,15 +20,15 @@ import (
 )
 
 type ConductorHttpClient struct {
-	httpClient *httpclient.HttpClient
+    httpClient *httpclient.HttpClient
 }
 
 func NewConductorHttpClient(baseUrl string) *ConductorHttpClient {
-	conductorClient := new(ConductorHttpClient)
-	headers := map[string]string{"Content-Type": "application/json", "Accept": "application/json"}
-	httpClient := httpclient.NewHttpClient(baseUrl, headers, true)
-	conductorClient.httpClient = httpClient
-	return conductorClient
+    conductorClient := new(ConductorHttpClient)
+    headers := map[string]string{"Content-Type": "application/json", "Accept": "application/json"}
+    httpClient := httpclient.NewHttpClient(baseUrl, headers, true)
+    conductorClient.httpClient = httpClient
+    return conductorClient
 }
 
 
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

File: core/src/main/java/com/netflix/conductor/service/MetadataService.java [hunk 1]
```diff
@@ -155,9 +155,26 @@ public void registerWorkflowDef(WorkflowDef def) {
...
+	public void unregisterWorkflowDef(String name, int version) {
+        if (name == null) {
+            throw new ApplicationException(Code.INVALID_INPUT, "Workflow name cannot be null");
+        }
+
+        if (version == 0) {
+            throw new ApplicationException(Code.INVALID_INPUT, "Version is not valid");
+        }
+
+	    metadata.removeWorkflowDef(name, version);
```
cyzhao: fix indent

client/go/conductorhttpclient.go
```diff
@@ -87,6 +87,23 @@ func (c *ConductorHttpClient) GetAllWorkflowDefs() (string, error) {
...
+	outputString, err := c.httpClient.Delete(url, params, nil, "")
+
+	if err != nil {
+		log.Println("Error while trying to Unregister Workflow Definition", workflowDefName, err)
+		return "", nil
+	} else {
+		return outputString, nil
+	}
+}
+
```
cyzhao: Nice!

mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLMetadataDAO.java
```diff
@@ -127,6 +127,17 @@ public WorkflowDef get(String name, int version) {
                                                                   .executeAndFetchFirst(WorkflowDef.class));
     }
 
+    @Override
+    public void removeWorkflowDef(String name, int version) {
+        final String DELETE_WORKFLOW_QUERY = "DELETE from meta_workflow_def WHERE name = ? AND version = ?";
+
+        executeWithTransaction(DELETE_WORKFLOW_QUERY, q -> {
+            if (!q.addParameter(name).addParameter(version).executeDelete()) {
+                throw new ApplicationException(ApplicationException.Code.NOT_FOUND, "No such workflow definition");
```
cyzhao: Add workflow name and version to the error message as well.

redis-persistence/src/main/java/com/netflix/conductor/dao/dynomite/RedisMetadataDAO.java
```diff
@@ -217,6 +217,17 @@ public WorkflowDef get(String name, int version) {
 		return def;
 	}
 
+	@Override
+	public void removeWorkflowDef(String name, int version) {
+		Preconditions.checkNotNull(name, "WorkflowDef name cannot be null");
+		Preconditions.checkNotNull(version, "Version cannot be null");
+		Long result = dynoClient.hdel(nsKey(WORKFLOW_DEF, name), String.valueOf(version));
+		if (!result.equals(1L)) {
+			throw new ApplicationException(Code.NOT_FOUND, "Cannot remove the workflow - no such workflow definition");
```
cyzhao: Same as above, add workflow name and version to error message.

core/src/main/java/com/netflix/conductor/service/MetadataService.java
falu2010-netflix: Looks like my indentation setting is different will fix it.

redis-persistence/src/main/java/com/netflix/conductor/dao/dynomite/RedisMetadataDAO.java
```diff
@@ -44,338 +44,349 @@
...
+            def = readValue(workflowDefJsonString, WorkflowDef.class);
+            recordRedisDaoPayloadSize("getWorkflowDef", workflowDefJsonString.length(), "n/a", name);
+        }
+        return def;
+    }
+
+    @Override
+    public void removeWorkflowDef(String name, int version) {
+        Preconditions.checkNotNull(name, "WorkflowDef name cannot be null");
+        Preconditions.checkNotNull(version, "Version cannot be null");
```
apanicker-nflx: I think, precondition checks for `isNotBlank` and `> 0` would be more useful.

test-harness/src/test/java/com/netflix/conductor/tests/integration/End2EndTests.java
```diff
@@ -203,7 +212,37 @@ public void testAll() throws Exception {
...
 		assertEquals(1, wf.getTasks().size());
-		
+
+	}
+
+	@Test
+	public void testMetadataWorkflowDefinition() {
+		WorkflowDef def = new WorkflowDef();
+		def.setName("testWorkflowDel");
+		def.setVersion(1);
```
kishorebanala: Question - Let's say we have 5 versions of workflows, can I delete in-between versions? I.e [1,2,3,4] etc.

jersey/src/main/java/com/netflix/conductor/server/resources/MetadataResource.java
```diff
@@ -81,6 +81,13 @@ public WorkflowDef get(@PathParam("name") String name, @QueryParam("version") In
 	public List<WorkflowDef> getAll() throws Exception {
 		return service.getWorkflowDefs();
 	}
+
+	@DELETE
+	@Path("/workflow/{name}")
+	@ApiOperation("Removes workflow definition")
+	public void unregisterWorkflowDef(@PathParam("name") String name, @QueryParam("version") Integer version) throws Exception {
```
kishorebanala: So version is optional, right? If not specified, version might be 0 by default and MetadataService would return an exception `Version is not valid`. If we do not want to make version optional, can we add it to Path? If it can be optional, does it deletes the latest version by default?

falu2010-netflix: It is required as per now. I can add to the path.

falu2010-netflix:
I found another issue while testing this last night. We have duplicate entry for workflow def in table with "LATEST" as key so when you do GET call **_without version_** after DELETE API call the workflow def is returned back even if after deleting it.

Issue #684, opened 2018-07-29
ElasticSearch input and output fields


jvemugunta:
In the mapping files mappings_docType_workflow.json and mappings_docType_task.json

the input and output fileds are defined as 

 "input": {
        "type": "string",
        "index": "analyzed"
      },
      "output": {
        "type": "string",
        "index": "analyzed"
      }

We want to use rely on Elastic Search Engine for querying for Tasks and Workflows instead of Conductor. Use Conductor as pure workflow engine and write all of the searches using Elastic Engine

Declaring the fields as above will be of no help from Elastic Search Perspective as JSON data is being converted as simple strings as below
https://github.com/Netflix/conductor/blob/9f221071f0177b17a5fc803f26cd7fab7a2c06b5/common/src/main/java/com/netflix/conductor/common/run/TaskSummary.java#L89-L96

How can we store as JSON payload and make all of the fields inside the JSON payload as Indexable.

I tried changing to

"input": {
        "type": "object",
        "dynamic": true
      },
      "output": {
        "type": "object",
        "dynamic": true
      },

But this seems to be not working.

Appreciate any pointers

jvemugunta:
Hi @v1r3n, 

I think you are one who wrote this code.
Can you reply on how we can store the input and output as JSON data while fully indexed searches.
This will help conductor project to get best out of Conductor and Search Engine.

v1r3n:
@jvemugunta The reason for making this a string is that different tasks could potentially have different data types for the same field.  e.g. one task might output a field as string vs. another as integer or a complex object.  The conflicting data types will cause issues with indexing as Elasticsearch indexing will fail.  Given the fields are analyzed as string, you still should be able to search for tasks and workflows based on the values of the input/output.

jvemugunta:
Hi @v1r3n , thanks for the reply.
If you look at the datatypes of Input and Output fields. they are defined as

private Map<String, Object> input = new HashMap<>();
private Map<String, Object> output = new HashMap<>();

Converting these fields as Strings, make the strings unsearchable from Elastic Search. The bigger goal for  plugging Conductor into Elastic Search Engine will be lost because the input and outout data is NOT searchable

We can still allow complex types into MAP object.

For example : 
private Map<String, Object> input = new HashMap<>();
input.put("age", new Integer(25)); will make the JSON data easily searchable

The dynamic mapping will turn as below

input": {
"dynamic": "true",
"properties": {
"age": {
"type": "int",
"fields": {
"keyword": {
[...]

PR #685, opened 2018-07-29
Support dynamic properties for input and output fields to support Elastic Search Engine mapping

jvemugunta:
https://github.com/Netflix/conductor/issues/684

The input and output fields in the Elastic Search Engine are declared as strings. I have changed them as JSONNodes to support dynamic mapping. This would make Elastic Search more powerful and allows searching for field values (JSON data) within input and output fields.

The following files were changed and I have tested them and has passed all of the test cases. This feature would bring best out of Elastic Search Engine and Netflix Conductor.

https://github.com/jvemugunta/conductor/blob/0816e51854faeeffdf6782735efa8c1c7f118fcb/common/src/main/java/com/netflix/conductor/common/run/WorkflowSummary.java#L95-L103

https://github.com/jvemugunta/conductor/blob/0816e51854faeeffdf6782735efa8c1c7f118fcb/common/src/main/java/com/netflix/conductor/common/run/TaskSummary.java#L92-L100

https://github.com/jvemugunta/conductor/blob/0816e51854faeeffdf6782735efa8c1c7f118fcb/common/build.gradle#L4-L6

https://github.com/jvemugunta/conductor/blob/0816e51854faeeffdf6782735efa8c1c7f118fcb/es5-persistence/src/main/resources/mappings_docType_task.json#L15-L21

https://github.com/jvemugunta/conductor/blob/0816e51854faeeffdf6782735efa8c1c7f118fcb/es5-persistence/src/main/resources/mappings_docType_workflow.json#L23-L29

https://github.com/jvemugunta/conductor/blob/0816e51854faeeffdf6782735efa8c1c7f118fcb/es2-persistence/src/main/resources/mappings_docType_task.json#L15-L21

https://github.com/jvemugunta/conductor/blob/0816e51854faeeffdf6782735efa8c1c7f118fcb/es2-persistence/src/main/resources/mappings_docType_workflow.json#L23-L30

### DIFF
--- a/es2-persistence/src/main/resources/mappings_docType_task.json
+++ b/es2-persistence/src/main/resources/mappings_docType_task.json
@@ -13,12 +13,12 @@
         "type": "long"
       },
       "input": {
-        "type": "string",
-        "index": "analyzed"
+        "type": "object",
+        "dynamic":  true
       },
       "output": {
-        "type": "string",
-        "index": "analyzed"
+        "type": "object",
+        "dynamic":  true
       },
       "queueWaitTime": {
         "type": "long"
--- a/es2-persistence/src/main/resources/mappings_docType_workflow.json
+++ b/es2-persistence/src/main/resources/mappings_docType_workflow.json
@@ -21,12 +21,12 @@
         "include_in_all": false
       },
       "input": {
-        "type": "string",
-        "index": "analyzed"
+        "type": "object",
+        "dynamic":  true
       },
       "output": {
-        "type": "string",
-        "index": "analyzed"
+        "type": "object",
+        "dynamic":  true
       },
       "reasonForIncompletion": {
         "type": "string",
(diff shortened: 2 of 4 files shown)


jvemugunta:
Hi @v1r3n @cyzhao 

Can you please review the pull request? 

Thank you

cyzhao:
@jvemugunta 

There are a couple of potential issues:
- fields of the same name but different data types for different workflow/task input will cause indexing errors.
- too many fields can be added dynamically leading to mapping explosion, which can cause out of memory errors, the maximum number of fields in an index is 1000 by default.

If you only have a small number of workflows that you have full control over, maybe this will work, but for us we run hundreds of workflows developed many different teams and this can lead to major problems.

We should make this configurable.

jvemugunta:
I hear your concern. Can we make at-least make  input and output fields as Objects instead of Strings.

"input": {
        "type": "object",
        "dynamic":  false
      },
      "output": {
        "type": "object",
        "dynamic":  false
      }

This will help at-least to store input and output as JSON data, instead of Strings.

cyzhao:
Changing the field type is not backward compatible and requires re-indexing. You can probably provide a custom implementation of IndexDAO that does exactly what you need.

PR #689, opened 2018-07-31
Refactor/api delete workflow definition

falu2010-netflix:
Description:

- Removed getAllLatest() api from MetadataDAO.java

- Removed latest key set for workflowdefinition keyspace in dynomite so duplicate values for worklfow definition are not maintained.

- Changed workflow delete api path from {workflowName}/version/{version} to {workflowName}/{version}


redis-persistence/src/main/java/com/netflix/conductor/dao/dynomite/RedisMetadataDAO.java
```diff
@@ -174,11 +179,21 @@ public WorkflowDef getLatest(String name) {
         Preconditions.checkNotNull(name, "WorkflowDef name cannot be null");
         WorkflowDef def = null;
 
-        String workflowDefJsonString = dynoClient.hget(nsKey(WORKFLOW_DEF, name), LATEST);
-        if (workflowDefJsonString != null) {
-            def = readValue(workflowDefJsonString, WorkflowDef.class);
-            recordRedisDaoRequests("getWorkflowDef");
-            recordRedisDaoPayloadSize("getWorkflowDef", workflowDefJsonString.length(), "n/a", def.getName());
+        // Get the list of valid versions leaving Latest
```
pctreddy: You can leverage the stream api to get the max version number and get rid of intermediate collection and sorting.

```
 Optional<Integer> optionalMaxVersion = dynoClient.hkeys(nsKey(WORKFLOW_DEF, name)).stream()
                .filter(key -> !key.equals(LATEST))
                .map(Integer::valueOf)
                .max(Comparator.naturalOrder());
``` 

or you can fully compose the function in the stream itself:
```
return dynoClient.hkeys(nsKey(WORKFLOW_DEF, name)).stream()
                .filter(key -> !key.equals(LATEST))
                .map(Integer::valueOf)
                .max(Comparator.naturalOrder())
                .map(String::valueOf)
                .map(workflowVersion -> Optional.ofNullable(dynoClient.hget(nsKey(WORKFLOW_DEF, name), workflowVersion))
                        .map(workflowDefinitionString -> readValue(workflowDefinitionString, WorkflowDef.class))
                        .orElse(null))
                .orElse(null);
```

redis-persistence/src/main/java/com/netflix/conductor/dao/dynomite/RedisMetadataDAO.java
```diff
@@ -219,14 +234,33 @@ public WorkflowDef get(String name, int version) {
...
         Preconditions.checkArgument(StringUtils.isNotBlank(name), "WorkflowDef name cannot be null");
-        Preconditions.checkArgument(version > 0, "Input version is not valid");
+        Preconditions.checkNotNull(version, "Input version cannot be null");
         Long result = dynoClient.hdel(nsKey(WORKFLOW_DEF, name), String.valueOf(version));
         if (!result.equals(1L)) {
             throw new ApplicationException(Code.NOT_FOUND, String.format("Cannot remove the workflow - no such workflow" +
                     " definition: %s version: %d", name, version));
         }
+
+        // do the clean up check if there are any more versions remaining if not delete the
```
pctreddy: Same as above

core/src/main/java/com/netflix/conductor/dao/MetadataDAO.java
```diff
@@ -103,7 +103,7 @@
 	 * @param name Name of the workflow
 	 * @param version version
 	 */
-	public void removeWorkflowDef(String name, int version);
+	public void removeWorkflowDef(String name, Integer version);
```
pctreddy: Does the removeWorkflow also delete the workflows that were associated with the definition ? If so the Java doc needs to added for this.  And also what would happen when the Workflow definition or the version does not exist.

Try to add precise details to the java doc to avoid confusion, `@param name Name of the workflow ` can be changed to `@param name Name of the WorkflowDefinition that is intended to be deleted` 
and ` @param version version` to  `@param version version of the workflow definition that will be deleted`

falu2010-netflix: It just deletes the workflow definition. I will update the docs.

falu2010-netflix: It returns 404 incase workflow definition or the version does not exist.

falu2010-netflix:
@pctreddy @cyzhao can you review this?

redis-persistence/src/main/java/com/netflix/conductor/dao/dynomite/RedisMetadataDAO.java
```diff
@@ -217,14 +233,27 @@ public WorkflowDef get(String name, int version) {
...
             throw new ApplicationException(Code.NOT_FOUND, String.format("Cannot remove the workflow - no such workflow" +
                     " definition: %s version: %d", name, version));
         }
+
+        // check if there are any more versions remaining if not delete the
+        // workflow name
+        Optional<Integer> optionMaxVersion = dynoClient.hkeys(nsKey(WORKFLOW_DEF, name)).stream()
+                .filter(key -> !key.equals(LATEST))
+                .map(Integer::valueOf)
+                .max(Comparator.naturalOrder());
```
cyzhao: extract method as it being used in two places

PR #698, opened 2018-08-02
REST and gRPC health checks.

gorzell:
Add simple health checks for the REST and gRPC APIs, based on the Netflix health check library.

The gRPC version also uses the standard proto definition/api from gRPC.


grpc-server/src/main/java/com/netflix/conductor/grpc/server/service/HealthServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+import io.grpc.stub.StreamObserver;
+
+public class HealthServiceImpl extends HealthGrpc.HealthImplBase {
+    private static final Logger LOGGER = LoggerFactory.getLogger(HealthServiceImpl.class);
+    private static final GRPCHelper GRPC_HELPER = new GRPCHelper(LOGGER);
+
+    private final HealthCheckAggregator healthCheck;
+
+    @Inject
+    public HealthServiceImpl(HealthCheckAggregator healthCheck) {
```
cyzhao: HealthModule needs to be installed.

gorzell: That is done in the `ServerModule`, which I believe at the moment is the correct place.

https://github.com/Netflix/conductor/pull/698/files#diff-d0cb0cb5585c979ec521f16dd0acce59R43

PR #699, opened 2018-08-03
Rate Limiting Abstraction for Tasks

pctreddy:
The Rate limiting configuration abstraction is currently at the TaskDef level and the rate limit is evaluated before starting execution in the WorkflowExecutor.

The Dummy implementation of the rate limit is provided by the `DummyRateLimitingService.java`

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
@@ -43,7 +43,9 @@ import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import static com.netflix.conductor.common.metadata.tasks.Task.Status.COMPLETED_WITH_ERRORS;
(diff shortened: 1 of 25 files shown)


core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
```diff
@@ -128,17 +130,16 @@ private DeciderOutcome decide(final WorkflowDef workflowDef, final Workflow work
...
 
-        Map<String, Task> tasksToBeScheduled = new LinkedHashMap<>();
-
-        preScheduledTasks.forEach(pst -> {
-            tasksToBeScheduled.put(pst.getReferenceTaskName(), pst);
-        });
+        //Traverse the pre-scheduled tasks to a linkedHasMap
+        Map<String, Task> tasksToBeScheduled = preScheduledTasks.stream()
+                .collect(Collectors.toMap(Task::getReferenceTaskName, Function.identity(),
+                        (element1, element2) -> element2, LinkedHashMap::new));
```
cyzhao: should throwingMerger be used here?

core/src/main/java/com/netflix/conductor/core/execution/mapper/TaskMapperContext.java
```diff
@@ -38,17 +40,48 @@
     private DeciderService deciderService;
 
 
-    public TaskMapperContext(WorkflowDef workflowDefinition, Workflow workflowInstance, WorkflowTask taskToSchedule,
+    /*private TaskMapperContext(WorkflowDef workflowDefinition, Workflow workflowInstance, TaskDef taskDefinition, WorkflowTask taskToSchedule,
```
cyzhao: remove

core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
pctreddy: That should be the case, but the existing code was just overwriting the value, kept the behavior of the existing code.

PR #715, opened 2018-08-11
expand inline json for event handler script evaluation

apanicker-nflx:
expand inline json for event handler script evaluation


core/src/main/java/com/netflix/conductor/core/utils/JsonUtils.java
```diff
@@ -0,0 +1,68 @@
+package com.netflix.conductor.core.utils;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.annotations.VisibleForTesting;
+
+import javax.inject.Singleton;
+import java.util.List;
+import java.util.Map;
+
+@Singleton
```
pctreddy: Why define it as a `@Singleton`  when this util class does not have any dependency and while the initialization in all the classes is done by calling the constructor `private final JsonUtils jsonUtils = new JsonUtils();` ?

core/src/main/java/com/netflix/conductor/core/utils/JsonUtils.java
```diff
@@ -0,0 +1,68 @@
...
+public class JsonUtils {
```
pctreddy: Missing java doc at the class level, having javadoc at class level will help future use cases.

core/src/main/java/com/netflix/conductor/core/utils/JsonUtils.java
```diff
@@ -0,0 +1,68 @@
...
+
+    private Object getJson(String jsonAsString) {
+        try {
+            return objectMapper.readValue(jsonAsString, Object.class);
+        } catch (Exception e) {
+            return jsonAsString;
+        }
+    }
+
+    @SuppressWarnings("unchecked")
```
pctreddy: Consider adding javadoc to this method since it can call private methods depending on the instance type.

pctreddy: public method does not need `@VisibleForTesting`

core/src/main/java/com/netflix/conductor/core/utils/JsonUtils.java
```diff
@@ -0,0 +1,68 @@
...
+    private final ObjectMapper objectMapper = new ObjectMapper();
+
+    public JsonUtils() {
+    }
+
+    private Object getJson(String jsonAsString) {
+        try {
+            return objectMapper.readValue(jsonAsString, Object.class);
+        } catch (Exception e) {
+            return jsonAsString;
```
pctreddy: Exception is swallowed here, consider propagating the exception and logger statements. If no exception is propagated add doc to this method or the calling method for the reasoning behind not propagating the exception.

core/src/main/java/com/netflix/conductor/core/utils/JsonUtils.java
```diff
@@ -0,0 +1,68 @@
...
+                expandMap((Map<String, Object>) value);
+            } else if (value instanceof List) {
+                expandList((List<Object>) value);
+            }
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    private void expandMap(Map<String, Object> input) {
+        for (Map.Entry<String, Object> e : input.entrySet()) {
```
pctreddy: use better names instead of single letters.

core/src/main/java/com/netflix/conductor/core/utils/JsonUtils.java
apanicker-nflx: Added javadoc

core/src/main/java/com/netflix/conductor/core/utils/JsonUtils.java
apanicker-nflx: Added a log message and added this to the javadoc for this method.

core/src/main/java/com/netflix/conductor/core/utils/JsonUtils.java
apanicker-nflx: Added javadoc and cleaned up.

core/src/main/java/com/netflix/conductor/core/utils/JsonUtils.java
apanicker-nflx: Done.

Issue #631, opened 2018-06-28
Add support for execution of ephemeral workflows.


gorzell:
We have a use case where we need to be able to submit a large/arbitrary number of different workflow definitions to be executed by Conductor.  I don't think that it makes sense to first add each of these definitions to the Metadata store in order to then execute at least once and quite possibly only once.  Instead I would like to add an API through which you can schedule a workflow execution by providing the definition directly, and that definition will only exist for that execution.  A side benefit is that this also has the potential to be really useful for testing.

## Design
I think that the best way to accomplish this is to modify the ExecutionService and supporting storage and execution classes to always contain the  WorkflowDefinition, rather than look it up from the Metadata service as it does today.

* Refactor WorflowId and Version into a WorkflowReference object to better express that they are always required as a pair.
* Refactor `Workflow` to contain a `WorkflowDefinition` and perhaps a `WorkflowReference`, although the reference would only be for tracking purposes and should never be used for a lookup.
* Persist the WorkflowDefinition as part of execution.
* Add an API for submitting a `WorkflowDefinition` to be scheduled for execution.

v1r3n:
@gorzell 
I think having ephemeral workflows have a merit and I can see more use cases for it.
The above design proposal makes sense.  However, few points:

* TaskDefs are also looked up at some point of time during execution.  Today some of the properties are looked up from TaskDefs, I think they should also be moved under WorkflowTask to WorkflowDefinition is self sufficient for such workflows.
* A possible enhancement could be to use this approach for all the workflows, where when submitting a workflow for execution, you could pass WorkflowDefinition or pre-defined workflow with name/version pair which is looked up to retrieve WorkflowDefinition and then used for execution.   This will make execution path streamlined.  The downside is that if the definition is changed - the running workflows will not get the updated definition and will have to be updated.

@cyzhao I think we have similar requirements too.

gorzell:
@v1r3n 
> A possible enhancement could be to use this approach for all the workflows

That is the approach I also think is best and have been working on, so I don't consider this an enhancement :-).  It is a bit complicated when considering backwards compatibility in the data objects and there is quite a bit of intermingled logic that needs to be sorted out.

I will take a look at the task defs.  I would generally think running workflows not having their definitions change is a feature rather than a downside, however I also realize that you all have some really long running flows.

cyzhao:
[...]

PR #716, opened 2018-08-12
Ephemeral workflows and tasks

josedab:
#### Current situation
At the moment, Conductor allows to execute workflows referencing to *{ workflow, task } definitions* stored as *metadata* in the system.
For example: a workflow execution that has a workflow definition associated and 10 custom tasks to run will imply:
- Creation of the 10 task definitions if they don't exist (assuming workflow task type `SIMPLE` for simplicity)
- Creation of the workflow definition
- Each time a definition needs to be retrieved, a call to db is performed.
In addition to that, the system allows current metadata that is in use to be altered, leading to possible inconsistencies/race conditions.

#### Ephemeral workflows and tasks support
Currently, we have use cases where we need to be able to submit a large/arbitrary number of different workflow definitions to be executed by Conductor.
As stated on issue https://github.com/Netflix/conductor/issues/631, adding first each of these definitions to the *Metadata store* in order to then execute them only once might be overkilling.

This PR adds to the current code the support of having workflow definitions embedded within the workflow execution *(ephemeral workflows)*.
We also thought it might be useful to have also *ephemeral tasks* (embedded task definition within the workflow task).

### Architectural changes and considerations
- At the moment, metadata definitions are fetched across the code. We thought, in order to 
keep the execution of a workflow consistent, to only fetch the metadata definitions 
at the beginning of a client request.
  - This concept of immutable definitions within a specific workflow execution led us 
  to refactor the code in order to not use `MetadataDAO` on `DeciderService` 
  and `ExecutionDAO` implementations.
- `Workflow` contains a `WorkflowDefinition`. 
  - When requesting a workflow execution, if the definition is not embedded, the new `MetadataMapperService` will get the WorkflowDef data and populate it within `Workflow`.
- `WorkflowTask` contains a `TaskDef`.
  - TaskDefinition will be populated on the workflowTask during the request for starting a workflow.
- Prefetching all the metadata for workflows and tasks during the execution and then saving them will add the following benefits:
  - Immutable workflow and task definitions: no inconsistencies during execution.
  - It is one step forward towards testing different scenarios on a workflow execution.
  - Reduce stress on db.

### Data migration
With this PR, the current scenarios are *{Stored, Ephemeral} workflow definition with {Stored, Ephemeral} tasks*.
However, if workflows are currently running on the system, these would need to finish before running Conductor with the new code.
If this is not suitable, please let us know.

### Examples
In addition to the `kitchensink.json` example, a couple of files were added on behalf of showcasing the current possibilities:
- kitchenSink-ephemeralWorkflowWithEphemeralTasks.json
- kitchenSink-ephemeralWorkflowWithStoredTasks.json
These examples are also loaded when `loadSample=true`. UI has been changed in order to accommodate to ephemeral workflows.


josedab:
This Sunday merge from dev caused several merge conflicts. Looking into that

josedab:
Resolved all conflicts from last Sunday dev merge. Will go through the diff early on Monday to check nothing got added by mistake on the merge conflict resolutions. 
Fixed unit tests and tested ui+backend with the ephemeral workflows

cyzhao:
@josedab regarding Data Migration, it is not suitable to require all existing workflows to be finished before we push code changes in this PR, most of Conductor workflows are long running and this would make the migration very hard (if not impossible). Please consider updating the logic so existing workflows still work with these changes.

common/src/main/java/com/netflix/conductor/common/metadata/tasks/Task.java
```diff
@@ -606,6 +608,15 @@ public void setOutputMessage(Any outputMessage) {
         this.outputMessage = outputMessage;
     }
 
+    /**
+     * @return {@link Optional} containing the task definition if available
+     */
+    public Optional<TaskDef> getTaskDefinition() {
+        return Optional.ofNullable(this.getWorkflowTask())
+                .map(workflowTask -> Optional.ofNullable(workflowTask.getTaskDefinition()))
+                .orElse(Optional.empty());
```
cyzhao: orElse is redundant, can be removed.

common/src/main/java/com/netflix/conductor/common/metadata/workflow/TaskType.java
```diff
@@ -0,0 +1,46 @@
...
+    public static final String TASK_TYPE_JOIN = "JOIN";
+    public static final String TASK_TYPE_FORK_JOIN_DYNAMIC = "FORK_JOIN_DYNAMIC";
+    public static final String TASK_TYPE_EVENT = "EVENT";
+    public static final String TASK_TYPE_WAIT = "WAIT";
+    public static final String TASK_TYPE_SUB_WORKFLOW = "SUB_WORKFLOW";
+    public static final String TASK_TYPE_FORK_JOIN = "FORK_JOIN";
+    public static final String TASK_TYPE_USER_DEFINED = "USER_DEFINED";
+    public static final String TASK_TYPE_SIMPLE = "SIMPLE";
+
+    private static Set<String> systemTasks = new HashSet<>();
```
cyzhao: I know this is existing code, but we should introduce a isSystemTask flag to the enum to remove this set.

core/src/main/java/com/netflix/conductor/core/metadata/MetadataMapperService.java
```diff
@@ -0,0 +1,63 @@
...
+import com.netflix.conductor.common.metadata.workflow.WorkflowTask;
+import com.netflix.conductor.core.execution.TerminateWorkflowException;
+import com.netflix.conductor.dao.MetadataDAO;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.inject.Inject;
+
+@Singleton
+public class MetadataMapperService {
```
cyzhao: javadoc

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -78,121 +80,250 @@
...
-            logger.info("A new instance of workflow {} created with workflow id {}", workflowName, workflowId);
-            //then decide to see if anything needs to be done as part of the workflow
-            decide(workflowId);
+        metadataMapperService.populateTaskDefinitions(workflowDefinition);
 
-            return workflowId;
+        // Obtain the names of the tasks with missing definitions
+        Set<String> missingTaskDefinitionNames = workflowDefinition.collectTasks().stream()
+                .filter(workflowTask ->
+                        (workflowTask.getType().equals(TaskType.SIMPLE.name()) && workflowTask.getTaskDefinition() == null))
```
cyzhao: duplicated knowledge, use `metadataMapperService::shouldPopulateDefinition`

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -78,121 +80,250 @@
...
-            workflow.setTaskToDomain(taskToDomain);
-            executionDAO.createWorkflow(workflow);
-            logger.info("A new instance of workflow {} created with workflow id {}", workflowName, workflowId);
-            //then decide to see if anything needs to be done as part of the workflow
-            decide(workflowId);
+        metadataMapperService.populateTaskDefinitions(workflowDefinition);
 
-            return workflowId;
+        // Obtain the names of the tasks with missing definitions
+        Set<String> missingTaskDefinitionNames = workflowDefinition.collectTasks().stream()
```
cyzhao: Consider moving this logic to `metadataMapperService.populateTaskDefinitions`

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -358,17 +503,43 @@ void completeWorkflow(Workflow wf) {
...
-        queueDAO.remove(deciderQueue, workflow.getWorkflowId());    //remove from the sweep queue
+        Monitors.recordWorkflowCompletion(workflow.getWorkflowName(), workflow.getEndTime() - workflow.getStartTime(), wf.getOwnerApp());
+        queueDAO.remove(DECIDER_QUEUE, workflow.getWorkflowId());    //remove from the sweep queue
         logger.debug("Removed workflow {} from decider queue", wf.getWorkflowId());
     }
 
+    @VisibleForTesting
+    Optional<WorkflowDef> lookupWorkflowDefinition(String workflowName, int workflowVersion) {
+        // FIXME: Add messages.
+        checkNotNull(workflowName);
```
cyzhao: Add a note - update to use ServiceUtils once this is merged with dev.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -566,15 +763,16 @@ public void updateTask(TaskResult taskResult) {
...
      * @param workflowId ID of the workflow to evaluate the state for
      * @return true if the workflow has completed (success or failed), false otherwise.
+     * @throws ApplicationException If there was an error - caller should retry in this case.
      */
     public boolean decide(String workflowId) {
 
         //If it is a new workflow the tasks will be still empty even though include tasks is true
         Workflow workflow = executionDAO.getWorkflow(workflowId, true);
-        WorkflowDef def = metadataDAO.get(workflow.getWorkflowType(), workflow.getVersion());
+        // FIXME: The workflow could be null.
```
cyzhao: In what scenario could it be null?

common/src/main/java/com/netflix/conductor/common/metadata/workflow/TaskType.java
josedab: Sure, sounds something feasible for this PR 👍

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
josedab: I have just checked on the `executionDAO` implementations and it seems to be fine. 
@gorzell Any extra context you can give on this one?

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
josedab: Moved and added test on top of that

josedab:
@cyzhao Thanks for taking the time for reviewing :)
Will test now and fix the concern about migrating workflows that have not been finished 👍

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
gorzell: It was probably just being defensive and may have been more related to calling metadata, as I think that might have returned null for an ID that didn't exist.

josedab:
@cyzhao Updated code in order to be backwards compatible with already running workflows. Let me know if any issues/concerns/suggestions

### Notes for the new changes related to backwards compatibility
- Reused integration tests for testing each of the sub cases for legacy workflows.
  - Moved test logic on `WorkflowServiceTest` to an abstract class `AbstractWorkflowServiceTest`. This abstract class contains all the tests, but relies on its implementations to load or start a workflow.
    - Two implementations:
      - WorkflowServiceTest: current code test
      - WorkflowLegacyMigrationTest: new test suite that loads workflow snapshots and tests there is backwards compatibility for running workflows in all the scenarios specified on `AbstractWorkflowServiceTest`. (40 new test have been added)
- Most of the time dedicated to this PR has been allocated to the understanding and creation of the scenarios for legacy running workflows. You can see all the scenarios on `resources/scenarios/legacy/` folder. I think I covered all cases, however if you think there is something missing we can add more scenarios in the future.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -78,121 +79,225 @@
...
+    private final Configuration config;
+
+    private final MetadataMapperService metadataMapperService;
 
     private RateLimitingService rateLimitingService;
 
-    private ParametersUtils parametersUtils = new ParametersUtils();
+    private final ParametersUtils parametersUtils = new ParametersUtils();
+
+    private int activeWorkerLastPollInSecs;
```
cyzhao: add javadoc to explain what this is

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -434,6 +606,14 @@ public void updateTask(TaskResult taskResult) {
 
         String workflowId = taskResult.getWorkflowInstanceId();
         Workflow workflowInstance = executionDAO.getWorkflow(workflowId);
+
+
+        // FIXME Backwards compatibility for legacy workflows already running.
+        // This code will be removed in a future version.
+        if(workflowInstance.getWorkflowDefinition() == null) {
```
cyzhao: This logic should be encapsulated in ExecutionDAO which automatically populates workflowDefinition and taskDefinitions for legacy workflows, now this is repeated in several places.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
josedab: That code was already there. No additions. I just grouped it with the rest of the final attributes

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -566,15 +746,19 @@ public void updateTask(TaskResult taskResult) {
...
      */
     public boolean decide(String workflowId) {
 
         //If it is a new workflow the tasks will be still empty even though include tasks is true
         Workflow workflow = executionDAO.getWorkflow(workflowId, true);
-        WorkflowDef def = metadataDAO.get(workflow.getWorkflowType(), workflow.getVersion());
+
+        // FIXME Backwards compatibility for legacy workflows already running.
+        // This code will be removed in a future version.
+        workflow = metadataMapperService.populateWorkflowWithDefinitions(workflow);
```
cyzhao: As mentioned earlier, this should not be needed.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -786,7 +997,7 @@ public void setTaskDomains(List<Task> tasks, Workflow wf) {
                 String[] domains = domainstr.split(",");
                 tasks.forEach(task -> {
                     // Filter out SystemTask
-                    if (!WorkflowTask.Type.isSystemTask(task.getTaskType())) {
+                    if (!TaskType.isSystemTask(task.getTaskType())) {
```
cyzhao: `task.getTaskType().isSystemTask()`

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -795,7 +1006,7 @@ public void setTaskDomains(List<Task> tasks, Workflow wf) {
 
             } else {
                 tasks.forEach(task -> {
-                    if (!WorkflowTask.Type.isSystemTask(task.getTaskType())) {
+                    if (!TaskType.isSystemTask(task.getTaskType())) {
```
cyzhao: `task.getTaskType().isSystemTask()`, add `isSystemTask` method that returns the boolean attribute back if it doesn't exist yet.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
josedab: The code is doing the same as before. `getTaskType()` is a String. Does not contain a method called `isSystemTask`. There is a possibility of refactoring the code in order to be a taskType but the change is not trivial: I believe SystemTaskType and TaskType overlap on a couple of scenarios where this String -> TaskType conversion fails. 
Happy to refactor the code to use TaskType in the future but I feel it is out of the scope of this PR. Do you think it is reasonable?

cyzhao: Got it, let's not worry about in this PR, we will refactor it later.

core/src/main/java/com/netflix/conductor/core/execution/mapper/SubWorkflowTaskMapper.java
```diff
@@ -64,19 +57,17 @@ public SubWorkflowTaskMapper(ParametersUtils parametersUtils, MetadataDAO metada
...
         subWorkflowTask.setReferenceTaskName(taskToSchedule.getTaskReferenceName());
         subWorkflowTask.setWorkflowInstanceId(workflowInstance.getWorkflowId());
-        subWorkflowTask.setWorkflowType(workflowInstance.getWorkflowType());
+        subWorkflowTask.setWorkflowType(workflowInstance.getWorkflowName());
         subWorkflowTask.setCorrelationId(workflowInstance.getCorrelationId());
         subWorkflowTask.setScheduledTime(System.currentTimeMillis());
         subWorkflowTask.setEndTime(System.currentTimeMillis());
         subWorkflowTask.getInputData().put("subWorkflowName", subWorkflowName);
-        subWorkflowTask.getInputData().put("subWorkflowVersion", subWorkflowVersion);
+        subWorkflowTask.getInputData().put("subWorkflowVersion", subWorkflowParams.getVersion());
```
cyzhao: is it safe to use this and remove existing `getSubWorkflowVersion` method?

core/src/main/java/com/netflix/conductor/core/metadata/MetadataMapperService.java
```diff
@@ -0,0 +1,166 @@
...
+                .orElseGet(() -> {
+                    WorkflowDef wd = lookupForWorkflowDefinition(workflow.getWorkflowName(), workflow.getWorkflowVersion());
+                    workflow.setWorkflowDefinition(wd);
+                    return wd;
+                });
+
+        workflowDefinition.collectTasks().stream().forEach(
+                workflowTask -> {
+                    if (shouldPopulateDefinition(workflowTask)) {
+                        workflowTask.setTaskDefinition(metadataDAO.getTaskDef(workflowTask.getName()));
```
cyzhao: Only call metadataDAO if taskDefinition is not set?

core/src/main/java/com/netflix/conductor/core/metadata/MetadataMapperService.java
```diff
@@ -0,0 +1,166 @@
...
+        workflowDefinition.collectTasks().stream().forEach(
+                workflowTask -> populateWorkflowTaskWithDefinition(workflowTask)
+        );
+        checkNotEmptyDefinitions(workflowDefinition);
+        return workflowDefinition;
+    }
+
+    private WorkflowTask populateWorkflowTaskWithDefinition(WorkflowTask workflowTask) {
+        if (shouldPopulateDefinition(workflowTask)) {
+            workflowTask.setTaskDefinition(metadataDAO.getTaskDef(workflowTask.getName()));
```
cyzhao: as above, check existence of taskDefinition first?

core/src/main/java/com/netflix/conductor/core/metadata/MetadataMapperService.java
cyzhao: Never mind, it is already done in shouldPopulateDefinition method.

core/src/test/java/com/netflix/conductor/metadata/MetadataMapperServiceTest.java
```diff
@@ -0,0 +1,221 @@
...
+
+import static junit.framework.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoMoreInteractions;
+import static org.mockito.Mockito.verifyZeroInteractions;
+import static org.mockito.Mockito.when;
+
+@RunWith(MockitoJUnitRunner.class)
+public class MetadataMapperServiceTest {
```
cyzhao: Nice test

redis-persistence/src/main/java/com/netflix/conductor/dao/dynomite/RedisMetadataDAO.java
```diff
@@ -107,277 +104,277 @@ private void refreshTaskDefs() {
...
-    private EventHandler getEventHandler(String name) {
-        EventHandler eventHandler = null;
-        String json = dynoClient.hget(nsKey(EVENT_HANDLERS), name);
-        if (json != null) {
-            eventHandler = readValue(json, EventHandler.class);
-        }
-        return eventHandler;
-
-    }
+	}
```
cyzhao: Please avoid format change to the entire file in a feature PR in the future.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
josedab: ExecutionDAO does not populate legacy workflows anymore, this was done intentionally. This class is only in charge of running and creating objects related to workflow and task executions.
 
If the problem is that you would like to see the population of the workflowDefinition on the ExecutionDAO layer, the code on that layer will just be a proxy to the MetadataMapperService.

However I would not recommend that because:
- Lets say we have a proxy method on the executionDao layer. We are going to be adding a deprecated method that consists on being a proxy to another method. On top of that we are likely to have duplication of code on each of the implementations of executionDAO.
- The way we start a workflow is loading the metadata on WorkflowExecutor and running decisions on it. That was the main idea behind immutable workflow definitions during workflow execution.
- When we need to update a legacy task, we load the metadata and run the decisions on it. At the moment of running decisions, the task definition for that task is immutable. Achieving similar benefits as a new workflow execution.

Having said that, if it does not convince the idea of use MetadataMapperService, we can create a new method for legacy reasons on the ExecutionDAO layer and proxy to metadataMapperService.

cyzhao:
@josedab @gorzell I've completed the review, once the last few comments are addressed and build is fixed, then we can merge it. We are also planning merge 2.0 back to dev/master branch and release 2.0 sometime in Sep, let me know if you still have other PRs that need to be merged.

Great work!

core/src/main/java/com/netflix/conductor/core/execution/mapper/SubWorkflowTaskMapper.java
josedab: The subworkflowParamVersion is now populated at the beginning of the process (MetadataMapperService). Wrote tests for that in that class. I believe there are tests also on the legacy ones that handle subworkflow task creations.
Are you thinking on a specific scenario?

core/src/main/java/com/netflix/conductor/core/metadata/MetadataMapperService.java
josedab: Only if not set and taskType is SIMPLE

redis-persistence/src/main/java/com/netflix/conductor/dao/dynomite/RedisMetadataDAO.java
josedab: Sure, a thought that comes to my mind is the possibility of using checkstyle plugin on gradle for the future and hook it up to the build? Hopefully changes like this will not be added anymore. Any way I will put extra attention to the code style changes.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
josedab: Resolving as this is same explanation and resolution as in https://github.com/Netflix/conductor/pull/716#discussion_r213918055

PR #742, opened 2018-08-29
Support configuring MySQL connections.

gorzell:
Add some more configuration options for MySQL connections.  The defaults are inline with what is currently happen and therefore shouldn't cause a behavior change for existing users.


jvemugunta:
Hi @gorzell 

Did you look at the class MySQLWorkflowModule I have made changes few weeks back?
It addresses all of the properties you mentioned. I like your idea of separating MYSQL Configuration from code. However, Consider having  setMaximumPoolSize and setMinimumIdle settings. These are very critical for High Volume applications. 

How can I test your code? what branch/repo I should pull from?


```
HikariDataSource dataSource = new HikariDataSource();
        dataSource.setJdbcUrl(config.getProperty("jdbc.url", "jdbc:mysql://localhost:3306/conductor"));
        dataSource.setUsername(config.getProperty("jdbc.username", "conductor"));
        dataSource.setPassword(config.getProperty("jdbc.password", "password"));
        dataSource.setAutoCommit(false);
        
        dataSource.setMaximumPoolSize(config.getIntProperty("jdbc.maxPoolSize", 20));
        dataSource.setMinimumIdle(config.getIntProperty("jdbc.minIdleSize", 5));
        dataSource.setIdleTimeout(config.getIntProperty("jdbc.idleTimeout", 1000*300));
        dataSource.setTransactionIsolation(config.getProperty("jdbc.isolationLevel", "TRANSACTION_REPEATABLE_READ"));
```

gorzell:
@jvemugunta I have been working off of and only really looking at the `2.0` branch.  There are significant changes to how modules/configuration/bootstrapping works in 2.0, so you might want to take a look at that.  If you need to test something you are welcome to pull the branch from this pr gorzell:gorzell/upstream/mysql-config.  Keep in mind that it is based on the 2.0 branch though, so it will probably look quite a bit different than `dev`/`master`.

If you have need for those additional configuration values I am happy to expose them.  But note that the names will likely change to enable better name spacing and ensure that they are consistent with the other values.

PR #748, opened 2018-09-05
Grpc integration tests for ephemeral workflows and tasks

josedab:
- Added missing tests for creating workflow executions with {ephemeral,stored} tasks

### DIFF
--- a/common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowDef.java
+++ b/common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowDef.java
@@ -273,4 +273,19 @@ public class WorkflowDef extends Auditable {
 				getSchemaVersion()
 		);
 	}
+
+	@Override
+	public String toString() {
+		return "WorkflowDef{" +
+				"name='" + name + '\'' +
+				", description='" + description + '\'' +
+				", version=" + version +
+				", tasks=" + tasks +
+				", inputParameters=" + inputParameters +
+				", outputParameters=" + outputParameters +
+				", failureWorkflow='" + failureWorkflow + '\'' +
+				", schemaVersion=" + schemaVersion +
+				", restartable=" + restartable +
+				'}';
+	}
 }
--- a/common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowDef.java
+++ b/common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowDef.java
@@ -28,6 +28,7 @@ import java.util.Optional;
 
 import com.github.vmg.protogen.annotations.ProtoField;
 import com.github.vmg.protogen.annotations.ProtoMessage;
+import com.google.common.base.MoreObjects;
 import com.netflix.conductor.common.metadata.Auditable;
 
 /**
@@ -276,16 +277,16 @@ public class WorkflowDef extends Auditable {
 
 	@Override
 	public String toString() {
-		return "WorkflowDef{" +
-				"name='" + name + '\'' +
-				", description='" + description + '\'' +
-				", version=" + version +
-				", tasks=" + tasks +
-				", inputParameters=" + inputParameters +
-				", outputParameters=" + outputParameters +
-				", failureWorkflow='" + failureWorkflow + '\'' +
-				", schemaVersion=" + schemaVersion +
-				", restartable=" + restartable +
-				'}';
+		return MoreObjects.toStringHelper(getClass())
+				.add("name", name)
+				.add("description", description)
+				.add("version", version)
+				.add("tasks", tasks)
+				.add("inputParameters", inputParameters)
+				.add("outputParameters", outputParameters)
+				.add("failureWorkflow", failureWorkflow)
+				.add("schemaVersion", schemaVersion)
+				.add("restartable", restartable)
+				.toString();
 	}
 }
--- a/test-harness/src/test/java/com/netflix/conductor/tests/integration/AbstractEndToEndTest.java
+++ b/test-harness/src/test/java/com/netflix/conductor/tests/integration/AbstractEndToEndTest.java
@@ -2,16 +2,29 @@ package com.netflix.conductor.tests.integration;
 
 import com.netflix.conductor.common.metadata.tasks.TaskDef;
 import com.netflix.conductor.common.metadata.workflow.TaskType;
+import com.netflix.conductor.common.metadata.workflow.WorkflowDef;
 import com.netflix.conductor.common.metadata.workflow.WorkflowTask;
 
 public class AbstractEndToEndTest {
+
     protected static final String TASK_DEFINITION_PREFIX = "task_";
+    private static final String DEFAULT_DESCRIPTION = "description";
+    // Represents null value deserialized from the redis in memory db
+    private static final String DEFAULT_NULL_VALUE = "null";
+
 
     protected WorkflowTask createWorkflowTask(String name) {
         WorkflowTask workflowTask = new WorkflowTask();
         workflowTask.setName(name);
         workflowTask.setWorkflowTaskType(TaskType.SIMPLE);
         workflowTask.setTaskReferenceName(name);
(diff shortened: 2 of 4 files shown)


common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowDef.java
```diff
@@ -273,4 +273,19 @@ public int hashCode() {
 				getSchemaVersion()
 		);
 	}
+
+	@Override
+	public String toString() {
+		return "WorkflowDef{" +
+				"name='" + name + '\'' +
```
gorzell: **Nit:** I am guessing this is auto-generated code?  It would be nice to not do so much string creation via `+`.  But maybe the JIT can optimize it away?

test-harness/src/test/java/com/netflix/conductor/tests/integration/AbstractEndToEndTest.java
```diff
@@ -0,0 +1,23 @@
+package com.netflix.conductor.tests.integration;
+
+import com.netflix.conductor.common.metadata.tasks.TaskDef;
+import com.netflix.conductor.common.metadata.workflow.TaskType;
+import com.netflix.conductor.common.metadata.workflow.WorkflowTask;
+
+public class AbstractEndToEndTest {
+    protected static final String TASK_DEFINITION_PREFIX = "task_";
```
gorzell: **Question:** I wonder if in the future it would make sense to setup a shared field/getter that could be set during construction to help with the name spacing issues we have discussed?

test-harness/src/test/java/com/netflix/conductor/tests/integration/End2EndGrpcTests.java
```diff
@@ -190,40 +190,117 @@ public void testAll() throws Exception {
...
+                .withName(workflowExecutionName)
+                .withWorkflowDef(workflowDefinition);
+
+        String workflowId = workflowClient.startWorkflow(workflowRequest);
+        assertNotNull(workflowId);
+
+        Workflow workflow = workflowClient.getWorkflow(workflowId, true);
+        WorkflowDef ephemeralWorkflow = workflow.getWorkflowDefinition();
+        assertNotNull(ephemeralWorkflow);
+        assertEquals(workflowDefinition.getName(), ephemeralWorkflow.getName());
```
gorzell: **Nit:** Can we compare the whole object here, rather than just the names?

test-harness/src/test/java/com/netflix/conductor/tests/integration/End2EndGrpcTests.java
```diff
@@ -190,40 +190,117 @@ public void testAll() throws Exception {
...
+        List<WorkflowTask> ephemeralTasks = ephemeralWorkflow.getTasks();
+        assertEquals(2, ephemeralTasks.size());
+        for (WorkflowTask ephemeralTask : ephemeralTasks) {
+            assertNotNull(ephemeralTask.getTaskDefinition());
+        }
+
+    }
+
+    // Helper method for creating task definitions on the server
+    private List<TaskDef> createAndRegisterTaskDefinitions(String prefixTaskDefinition, int numberOfTaskDefinitions) {
```
gorzell: **Suggestion:** Perhaps this can be `protected` and moved to the parent/abstract class?  If there is logic specific to the client type then you could implement that as an abstract method?

test-harness/src/test/java/com/netflix/conductor/tests/integration/End2EndGrpcTests.java
```diff
@@ -190,40 +190,117 @@ public void testAll() throws Exception {
...
+        assertEquals(workflowDefinition.getName(), ephemeralWorkflow.getName());
+
+        List<WorkflowTask> ephemeralTasks = ephemeralWorkflow.getTasks();
+        assertEquals(2, ephemeralTasks.size());
+        for (WorkflowTask ephemeralTask : ephemeralTasks) {
+            assertNotNull(ephemeralTask.getTaskDefinition());
+        }
+
+    }
+
```
gorzell: **Concern:** Can we also add a test that is using a mix of ephemeral and stored task definitions?

test-harness/src/test/java/com/netflix/conductor/tests/integration/End2EndTests.java
```diff
@@ -339,7 +324,7 @@ private TaskDef createTaskDefinition(String name) {
             def.setTimeoutPolicy(TimeoutPolicy.RETRY);
             definitions.add(def);
         }
-        taskClient.registerTaskDefs(definitions);
+        metadataClient.registerTaskDefs(definitions);
```
gorzell: **Suggestion:** See comment above.

common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowDef.java
josedab: Yeah, autogenerated code. I think JIT will take care of that, however I don't see why not fix it at code level.
Will use `MoreObjects.toStringHelper` from guava, i think it is the way to go on this one.

test-harness/src/test/java/com/netflix/conductor/tests/integration/AbstractEndToEndTest.java
josedab: Good question. I am sure we can do something. I noticed there is a property called `workflow.namespace.prefix` that you can pass to the TestEnvironment. I played with it in the past when working on the integration tests. Will make a note for this.
In this case, this prefix is just used in order to avoid hardcoding `task_` over the code.

josedab: In addition to my last comment and answering your question,  I think the namespace is something we can do in a near future modifying slightly the `TestEnvironment` class.

test-harness/src/test/java/com/netflix/conductor/tests/integration/End2EndGrpcTests.java
josedab: This is a tricky one:
Client types are different and they are the implementations themselves. It would be great if they both implemented an interface but it is not the case. Also, I believe these are generated meaning I cannot change them.
One thing i thought yesterday to abstract this was to use Consumer<...> but will play around with it today.
Will update this thread with any ideas

josedab: I see what you are saying now. Sure, moving it up 👍

test-harness/src/test/java/com/netflix/conductor/tests/integration/AbstractEndToEndTest.java
```diff
@@ -0,0 +1,65 @@
...
+    // Represents null value deserialized from the redis in memory db
+    private static final String DEFAULT_NULL_VALUE = "null";
+
+    protected WorkflowTask createWorkflowTask(String name) {
+        WorkflowTask workflowTask = new WorkflowTask();
+        workflowTask.setName(name);
+        workflowTask.setWorkflowTaskType(TaskType.SIMPLE);
+        workflowTask.setTaskReferenceName(name);
+        workflowTask.setDescription(getDefaultDescription(name));
+        workflowTask.setDynamicTaskNameParam(DEFAULT_NULL_VALUE);
```
josedab: This default null values population are necessary as the readJson mechanism populates this attributes. In order to use `WorkflowDef.equals` on the tests these need to be populated.

test-harness/src/test/java/com/netflix/conductor/tests/integration/End2EndTests.java
```diff
@@ -247,100 +265,4 @@ public void testMetadataWorkflowDefinition() {
         }
     }
 
-    @Test
```
josedab: These tests have been moved all the way up to the abstract class on behalf of not duplicating code. They are used on `End2EneTests` and `End2EndGrpcTests`

PR #749, opened 2018-09-05
GRPC server reflection support 

josedab:
- Having the ability of enabling reflection on the grpc server will allow grpc clients to know which services are exposed.
- The feature can be controlled through the property `conductor.grpc.server.reflection.enabled`. By default the feature is enabled.

### DIFF
--- a/grpc-server/src/main/java/com/netflix/conductor/grpc/server/GRPCServer.java
+++ b/grpc-server/src/main/java/com/netflix/conductor/grpc/server/GRPCServer.java
@@ -1,18 +1,15 @@
 package com.netflix.conductor.grpc.server;
 
 import com.netflix.conductor.service.Lifecycle;
-
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-import java.io.IOException;
-import java.util.Arrays;
-
-import javax.inject.Singleton;
-
 import io.grpc.BindableService;
 import io.grpc.Server;
 import io.grpc.ServerBuilder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.inject.Singleton;
+import java.io.IOException;
+import java.util.List;
 
 @Singleton
 public class GRPCServer implements Lifecycle {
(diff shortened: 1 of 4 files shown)


grpc-server/src/main/java/com/netflix/conductor/grpc/server/GRPCServer.java
```diff
@@ -21,9 +18,9 @@
 
     private final Server server;
 
-    public GRPCServer(int port, BindableService... services) {
+    public GRPCServer(int port, List<BindableService> services) {
```
gorzell: **Question:** Why the change to a `List` from varargs?

grpc-server/src/main/java/com/netflix/conductor/grpc/server/GRPCServerProvider.java
```diff
@@ -43,15 +46,25 @@ public GRPCServerProvider(
...
-                                eventServiceImpl,
-                                metadataServiceImpl,
-                                taskServiceImpl,
-                                workflowServiceImpl
-                        ))
+                Optional.of(getGRPCServer())
                 : Optional.empty();
     }
+
+    private GRPCServer getGRPCServer() {
```
gorzell: **Nit:** Rename to `buildGRPCServer`.

**Suggestion:** Pass reflectionEnabled and port as arguments, or pass the configuration as an argument.  I think it improves clarity and testability.

grpc-server/src/main/java/com/netflix/conductor/grpc/server/GRPCServerProvider.java
```diff
@@ -43,15 +46,25 @@ public GRPCServerProvider(
...
+                taskServiceImpl,
+                workflowServiceImpl);
+
+        if (configuration.isReflectionEnabled()) {
+            services.add(ProtoReflectionService.newInstance());
+        }
+
+        return new GRPCServer(
+                configuration.getPort(),
+                services.build()
```
gorzell: Related to the above question about varargs, I am pretty sure you could have just passed an Array here rather than change the signature. /shrug

grpc-server/src/main/java/com/netflix/conductor/grpc/server/GRPCServer.java
josedab: Only consumer of this constructor was `GRPCServerProvider`. As I needed to add some optional bindable services, the question was more related to choose one of these options:
- Updating GRPCServer to take a list of BindableServices(GRPCServer change)
- Converting to array the BindableService list (GRPCServerProvider change).
I chose to make the change on GRPCServer because:
- I thought that converting to array the BindableService list on GRPCServerProvider was unnecessary as there was only one use of the constructor.
- I thought we were not losing in readability that much

Issue #754, opened 2018-09-06
mysql timeoutSeconds and responseTimeoutSeconds are not working


jvemugunta:
The same works with REDIS and doesn't work with mysql. Did any one encounter this issue?

You can recreate the issue as below

Task and Workflow definations
```
curl -X POST --header 'Content-Type: application/json' --header 'Accept: application/json' 'http://localhost:8080/api/metadata/taskdefs' -d '
[{
  "name": "T1",
  "retryCount": 1,
  "timeoutSeconds": 60,
  "inputKeys": [
    "sourceRequestId",
    "qcElementType"
  ],
  "outputKeys": [
    "state",
    "skipped",
    "result"
  ],
  "timeoutPolicy": "TIME_OUT_WF",
  "retryLogic": "FIXED",
  "retryDelaySeconds": 30,
  "responseTimeoutSeconds": 60
}]
'

curl -X PUT --header 'Content-Type: application/json' --header 'Accept: application/json' 'http://localhost:8080/api/metadata/workflow' -d '
[{
  "name": "sample",
  "description": "sample workflow",
  "version": 1,
  "tasks": [
    {
      "name": "T1",
      "taskReferenceName": "T1",
      "type": "SIMPLE",
      "inputParameters": {
        "fileName": "${workflow.input.fileName}"
      }
    }
  ],
  "outputParameters": {
   
  },
  "schemaVersion": 2
}]
'
```
Workflow Instance

```
curl -X POST --header 'Content-Type: application/json' --header 'Accept: text/plain' 'http://localhost:8080/api/workflow' -d '
{
  "name": "sample",
  "version": 1,
  "correlatonId": "corr1",
  "input": {
    "fileName": "f1"
  }
}
'
```

Call this URL
curl -X GET --header 'Accept: application/json' 'http://localhost:8080/api/tasks/poll/T1'

wait for 2 mins, you will see the task is in_progress. Repeat the same with REDIS and backend and the task will go into time_out status after 2 mins.

It apparently looks like a bug to me in mysql.

jvemugunta:
Hi @mashurex and @picaron 
Any pointers on this issue. I just realized it. Are you have the same problem?

https://github.com/Netflix/conductor/blob/b8fa64ba8a0735192e56c06a790dbf193d600171/core/src/main/java/com/netflix/conductor/core/execution/WorkflowSweeper.java#L74-L96


[...]

PR #757, opened 2018-09-08
Fix for #754 Add scheduled sweeper to MysqlQueueDAO

mashurex:
The MySqlQueueDAO did not implement its own sweeper for processing un-acknowled messages that had been popped and there is no external service calling `QueueDAO::processUnacks` at any regular interval.

Following the pattern that `RedisDynoQueue` uses, I added a scheduled thread to call a new method `processAllUnacks` on a 60s interval. This should resolve issue #754.


mashurex:
@jvemugunta This should be the fix you were looking for.

mashurex:
The last two builds failed because of `test-harness` failures which do not test the mysql persistence module changes.

mashurex:
Closed for PR #771

PR #761, opened 2018-09-10
Fixes the docker conductor:ui build.

bitbier:
There appears to be a weird error with the node dependencies and the current node version
used in the docker UI container image. You can find more information: https://github.com/gulpjs/gulp/issues/2162.

If you run the conductor UI container, it will not startup properly because the `npm run build --server` will fail with the following error:

```
../src/node_contextify.cc:631:static void node::contextify::ContextifyScript::New(const FunctionCallbackInfo<v8::Value> &): Assertion `args[1]->IsString()' failed.
```

Which is caused by the natives library. I simply just `npm update` in order to upgrade the packages.

This isn't caught in the Dockerfile for some reason. Even after using `set -e`, docker will run the command and continue on. Not sure why that happens, but you can run the command from the specific layer to see the real error message. Because the `dist` folder is never created, the `startup.sh` script will not run.

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -8346,9 +8346,9 @@
       }
     },
     "natives": {
-      "version": "1.1.2",
-      "resolved": "https://registry.npmjs.org/natives/-/natives-1.1.2.tgz",
-      "integrity": "sha512-5bRASydE1gu6zPOenLN043++J8xj1Ob7ArkfdYO3JN4DF5rDmG7bMoiybkTyD+GnXQEMixVeDHMzuqm6kpBmiA==",
+      "version": "1.1.4",
+      "resolved": "https://registry.npmjs.org/natives/-/natives-1.1.4.tgz",
+      "integrity": "sha512-Q29yeg9aFKwhLVdkTAejM/HvYG0Y1Am1+HUkFQGn5k2j8GS+v60TVmZh6nujpEAj/qql+wGUrlryO8bF+b1jEg==",
       "dev": true
     },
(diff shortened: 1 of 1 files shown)


bitbier:
I fixed the merge conflict and simply just updated the `natives` package directly in the package-lock.json file to fix this issue. I'm unsure how to use npm to update with what was in master. For some reason, `npm upgrade` didn't do it's thing.

Anyways, the new lock file and `natives` 1.1.4 fixes the docker compose build.

kakakikikeke:
I also met this problem. I changed the version of natives according to this pull request and I solved it. I think that this pull request should be merged.

davidwadden:
Can we get this reviewed for a merge?  I too have to maintain this change in a fork

Also looks like upstream could use a rebase

davidwadden:
Actually, I'm not able to reproduce the problem he reports in Docker when using the `node:9-alpine` image.  According to gulpjs/gulp#2162, this is a problem with `gulp` on Node 10 when the dependency `natives < 1.1.3`, which by default resolves to `1.1.2`.

Node 9 is past end-of-life.  Node 10 is current LTS release.

As far as I can find, NPM doesn't provide for a way to force resolution without using special tools that would change the build toolchain, hence this being changed using the `package-lock.json`.   It would make sense to update the Docker base image to Node 10 along with this change.

apanicker-nflx:
I believe this should already be resolved in the newest conductor versions. Please reopen/submit a new PR if you still see the issue.

PR #771, opened 2018-09-17
Mysql bugs - Regarding issue 754

jvemugunta:
This is reference to mysql bugs I have encountered as the part of issue

https://github.com/Netflix/conductor/issues/754

I have created integration tests to match with Redis DAOs. All of them has passed along with all of the race conditions. I have modified some 

 @mashurex - Can you review them and let me know. I have fixed some race conditions from your pull request. If you are good with my code, we will ask to merge the code into master


jvemugunta:
@mashurex - Can you review this code? The current code has race conditions and I have addressed all of them.

mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLExecutionDAO.java
```diff
@@ -575,17 +584,12 @@ private void removeWorkflow(Connection connection, String workflowId) {
     }
 
     private void addPendingWorkflow(Connection connection, String workflowType, String workflowId) {
-        String EXISTS_PENDING_WORKFLOW = "SELECT EXISTS(SELECT 1 FROM workflow_pending WHERE workflow_type = ? AND workflow_id = ?)";
 
-        boolean exist = query(connection, EXISTS_PENDING_WORKFLOW,
-                q -> q.addParameter(workflowType).addParameter(workflowId).exists());
+        String INSERT_PENDING_WORKFLOW = "INSERT IGNORE INTO workflow_pending (workflow_type, workflow_id) VALUES (?, ?)";
```
mashurex: Good use of insert ignore, should really help w/ concurrency issues.

mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLWorkflowModule.java
```diff
@@ -42,9 +42,10 @@ public DataSource getDataSource(Configuration config) {
 
     @Override
     protected void configure() {
-        bind(MetadataDAO.class).to(MySQLMetadataDAO.class);
-        bind(ExecutionDAO.class).to(MySQLExecutionDAO.class);
-        bind(QueueDAO.class).to(MySQLQueueDAO.class);
+        
+        bind(MetadataDAO.class).to(MySQLMetadataDAO.class).asEagerSingleton();
+        bind(ExecutionDAO.class).to(MySQLExecutionDAO.class).asEagerSingleton();
+        bind(QueueDAO.class).to(MySQLQueueDAO.class).asEagerSingleton();
```
mashurex: Was this solving a booting problem you were running into?

I'm not sure of the implications of `asEagerSingleton()` with respect to the rest of the modules. This will likely require feedback from @v1r3n or someone with more overall knowledge of the module bootstrapping process to weigh in.

mashurex:
@jvemugunta I incorporated your changes into my previous PR, the only difference is how I've setup the test harness. If you merged in my PR branch to yours I could just close it and we can use this PR to post both of our changes.

jvemugunta:
If you look at the instance variables of MyDAOs,

 protected final Logger logger = LoggerFactory.getLogger(getClass());

    protected final ObjectMapper objectMapper;

    protected final DataSource dataSource;


All these three variables are thread safe. Yes, Creating many of
MYSQLQueueDAO instances results in calling the below piece of code in all
of instances. This results in not only race conditions, but also heavy load
on queue_message table


unackScheduler = Executors.newScheduledThreadPool(1);

        unackScheduler.scheduleAtFixedRate(this::processAllUnacks,
UNACK_SCHEDULE_MS, UNACK_SCHEDULE_MS,

                TimeUnit.MILLISECONDS);



MYSQL itself is a transactional database - where all of the queues, tasks,
workflows are in their own tables. In case of Redis, this is not the case.
A typical queue is associated with a workflowID. This means you will have
1000s of queue instances at runtime.

I would like to know

jvemugunta:
> @jvemugunta I incorporated your changes into my previous PR, the only difference is how I've setup the test harness. If you merged in my PR branch to yours I could just close it and we can use this PR to post both of our changes.

ok @mashurex , I will merge your changes into mine so that we work with one code base.
Is all your code in this repo
https://github.com/mashurex/conductor/tree/fix/754_mysql_timeout?

jvemugunta:
@mashurex , Your branch is merged into this PR. Please review and let me know what you think.

@mashurex and @picaron - Can we revisit the design of queue_message table?

https://github.com/Netflix/dyno-queues/blob/dev/dyno-queues-redis/src/main/java/com/netflix/dyno/queues/redis/RedisDynoQueue.java

If you look at the push, pop and processUnacks methods - It separates unackTime time value from myQueueShard time value where as we are merging these two into deliver_on column.


Can you design our implementation similar to REDIS?

We are doing select and updates on queue_message table - This would require locks when selecting and updating data with high volume

How about inserts and deletes - This way the code will be clean and runtime with MYSQL may be better.

Your thoughts?

mashurex:
`MySQLQueueDAO` is designed to be a singleton, so I'm not too worried about race conditions on thread safe instance variables. If they were **not** thread safe then we'd be staring down some real issues, but you are right about the unack scheduler. It should be made static to absolutely ensure there's only one sweeper running. Or even move the sweeper off to it's own class.

> If you look at the instance variables of MySQLDAOs, protected final Logger logger = LoggerFactory.getLogger(getClass()); protected final ObjectMapper objectMapper; protected final DataSource dataSource; All these three variables are thread safe. Yes, Creating many of MYSQLQueueDAO instances results in calling the below piece of code in all of instances. This results in not only race conditions, but also heavy load on queue_message table unackScheduler = Executors.newScheduledThreadPool(1); unackScheduler.scheduleAtFixedRate(this::processAllUnacks, UNACK_SCHEDULE_MS, UNACK_SCHEDULE_MS, TimeUnit.MILLISECONDS); MYSQL itself is a transactional database - where all of the queues, tasks, workflows are in their own tables. In case of Redis, this is not the case. A typical queue is associated with a workflowID. This means you will have 1000s of queue instances at runtime.

jvemugunta:
@mashurex I prefer Sweeper moving to its own classes.

I prefer SQLDAOs as singleton objects because all of the queues, tasks and workflows are handled by an unique connection using datasource.getConnection(). I am not 100% affirmative but a new DAO object is created for every new workflow instance which can result into thousands of objects under high volume.

jvemugunta:
@mashurex , Please review all the final changes.

test-harness/build.gradle
```diff
@@ -18,9 +19,13 @@ dependencies {
 
 	testCompile "com.google.inject.extensions:guice-servlet:${revGuiceServlet}"
 	testCompile "io.swagger:swagger-jersey-jaxrs:${revSwaggerJersey}"
+
+	testCompile "ch.vorburger.mariaDB4j:mariaDB4j:${revMariaDB4j}"
+	testCompile "ch.qos.logback:logback-core:${revLogbackCore}"
+//	testCompile "ch.qos.logback:logback-classic:${revLogbackClassic}"
```
mashurex: Should remove the commented out testCompile include (this was leftover from my PR branch)

server/src/main/java/com/netflix/conductor/server/ServerModule.java
```diff
@@ -114,6 +116,8 @@ protected void configure() {
 		} else {
 			bind(ExternalPayloadStorage.class).to(DummyPayloadStorage.class);
 		}
+		
+		bind(WorkflowSweeper.class).asEagerSingleton();
```
mashurex: Why are we adding this binding? I don't believe our work on this should affect the core module loading in any way at all.

mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLQueueDAO.java
```diff
@@ -23,11 +23,17 @@
 import com.netflix.conductor.core.execution.ApplicationException;
 import com.netflix.conductor.dao.QueueDAO;
 
-public class MySQLQueueDAO extends MySQLBaseDAO implements QueueDAO {
+public class MySQLQueueDAO extends MySQLBaseDAO implements QueueDAO, Closeable {
+    private static final Long UNACK_SCHEDULE_MS = 60_000L;
+    private final ScheduledExecutorService unackScheduler;
```
mashurex: This is the reference I was suggesting should be made static or moved off into a separate queue sweeper, I will push something to my branch to show what I meant.

mashurex:
If you review the latest in my #757 PR branch, you'll see the changes I made to ensure our DAOs are singleton and reduce the thread pool for our scheduled sweeps. Comparing our DAOs to the Redis DAOs, I realized we weren't using a `@Singleton` annotation. The Redis ones do, which means we should expect in all cases that only a single instance of the MySQL DAOs will be instantiated and we are safe to schedule the sweeper thread directly from the MySQL QueueDAO

server/src/main/java/com/netflix/conductor/server/ServerModule.java
jvemugunta: For timeouts to work, we need this binding. WorkflowSweeper is the background thread that sweeps _deciderQueue regularly to check from Timeouts. I am not sure why this was removed in the original implementation. May be @v1r3n has the reason for omitting this change

mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLQueueDAO.java
jvemugunta: Good point and the @singleton annotation should solve our issue. If you want to move the processAllUnacks, please do. That would be a good move

test-harness/build.gradle
jvemugunta: This is complete

server/src/main/java/com/netflix/conductor/server/ServerModule.java
jvemugunta: I have reverted the ui/package-lock.json from origin/dev. I am not sure why this file shows as change. I could have entered a space by mistake

jvemugunta:
@mashurex - Two things are left
1) In MYSQLQueue DAO move the below code to a separate class
Executors.newSingleThreadScheduledExecutor()
                .scheduleAtFixedRate(this::processAllUnacks,
                        UNACK_SCHEDULE_MS, UNACK_SCHEDULE_MS, TimeUnit.MILLISECONDS);

2) bind(WorkflowSweeper.class).asEagerSingleton();  This change is mandatory for the TimeOuts to work

cyzhao:
@mashurex @jvemugunta can you confirm that https://github.com/Netflix/conductor/pull/757 can be closed given it is merged into this one? also let us know when this PR is good to go then we will review and merge accordingly.

jvemugunta:
@cyzhao - Please close #757 . @mashurex and I are working on this PR

jvemugunta:
@mashurex - Once you review the final changes, we can ask the code to be merged. I didn't create a separate sweeper class for processAllUnacks() method in MySQLQueueDAO. The change is small and i am thinking there is not need of a separate class

jvemugunta:
Hi @joshlreese , Which branch your  code is running with?

ON DUPLICATE KEY UPDATE would not create a deadlock. 

You probably are seeing the deadlock because of 

https://github.com/Netflix/conductor/blob/2c419cff934959f72bdb48284b433c6d26f52a24/mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLQueueDAO.java#L210

This is because of Pessimistic we implemented, but later realized this would cause deadlocks under heavy load. With this fix we are going to change as Optimistic lock. This would avoid the dead lock situation.

If you can recreate the deadlock, I want to see the behavior. can you run this command
SHOW ENGINE INNODB STATUS \G

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -615,6 +615,16 @@ public boolean decide(String workflowId) {
 
             stateChanged = scheduleTask(workflow, tasksToBeScheduled) || stateChanged;
 
+            if (!outcome.tasksToBeUpdated.isEmpty()) {
```
mashurex: Is this a change coming from upstream? If so it shouldn't be part of this commit PR. If it isn't, I'm pretty sure we shouldn't be modifying WorkflowExecutor in this PR.

server/dependencies.lock
```diff
@@ -10,7 +10,7 @@
             "firstLevelTransitive": [
                 "com.netflix.conductor:conductor-contribs"
             ],
-            "locked": "1.11.415"
+            "locked": "1.11.418"
```
mashurex: Changes to server/dependencies.lock should not be included in this PR

server/src/main/java/com/netflix/conductor/server/ServerModule.java
```diff
@@ -1,23 +1,29 @@
 /**
  * Copyright 2016 Netflix, Inc.
- *
+ * <p>
```
mashurex: No changes to the server module should be included in this PR

jvemugunta: Removed it as it would be merged from another PR #781

server/dependencies.lock
jvemugunta: Removed this file

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
jvemugunta: That is part of bug I that I have opened on another PR #781 . I have removed this file

jvemugunta:
@mashurex . Removed files from a different PR #781 . Let me know what you think

cyzhao:
@jvemugunta we merged 2.0 to dev and it introduced conflicts to this PR, please resolve the conflict when you get a chance, although you can stick with 1.12 release but I highly recommend you to upgrade your deployment to 2.0 which should be backward compatible.

jvemugunta:
@cyzhao - I will take care of it this weekend.

jvemugunta:
@cyzhao  - I have resolved all of the conflicts and merged changes from dev branch. All the tests passed on my local, but are being filed on WorkflowServiceTest. I have not control on this class. Can you look into it when you get a  chance.

jvemugunta:
@cyzhao - These are the errors I am seeing with Travis-CI, but same works on my local

```
com.netflix.conductor.tests.integration.WorkflowServiceTest > testCaseStatements FAILED
    com.netflix.conductor.core.execution.ApplicationException: Cannot find the task definitions for the following tasks used in workflow: [junit_task_4, junit_task_3, junit_task_2, junit_task_1]
        at com.netflix.conductor.core.metadata.MetadataMapperService.checkNotEmptyDefinitions(MetadataMapperService.java:146)
        at com.netflix.conductor.core.metadata.MetadataMapperService.populateTaskDefinitions(MetadataMapperService.java:104)
        at com.netflix.conductor.core.execution.WorkflowExecutor.startWorkflow(WorkflowExecutor.java:262)
        at com.netflix.conductor.core.execution.WorkflowExecutor.startWorkflow(WorkflowExecutor.java:231)
        at com.netflix.conductor.core.execution.WorkflowExecutor.startWorkflow(WorkflowExecutor.java:153)
        at com.netflix.conductor.tests.integration.WorkflowServiceTest.startOrLoadWorkflowExecution(WorkflowServiceTest.java:30)
```

cyzhao:
@apanicker-nflx please take a look at WorkflowServiceTest test failures with this PR.

jvemugunta:
@cyzhao and @apanicker-nflx - This puzzles me and I am waiting for your input on this failed tests.
I have installed travis-ci on my local docker to match with github travis-ci . My local travis-ci passes all the tests.

apanicker-nflx:
@jvemugunta I tried out the changes on a test branch (forked from this branch). This is the branch with my changes - https://github.com/Netflix/conductor/commits/jvemugunta-mysql-bugs. All the tests and the build is successful when run on my local machine.

The travis builds are still failing, however, now it seems to be failing due to the MySQLWorkflowServiceTest tests.

The commit that showed this is - https://github.com/Netflix/conductor/commit/9ef195a6719c988b2219781205c21d7a59a03d0a.

From the build failures on your branch and then on my branch, we can see that the `*ServiceTest` that runs first (Redis or MySQL) is always successful and the latter one always fails. The common pattern in the failures is that all of the test cases fail because they are unable to find the workflow and task metadata.

jvemugunta:
@apanicker-nflx - why would you think this happens on Travis-CI with GITHUB?
I have installed Travis-CI docker image on my local machine and it works with no issues? Do you think if these tests *ServiceTest are running in parallel resulting into some kind of race conditions

jvemugunta:
@apanicker-nflx - I have dumbed-down my test case (MySQLWorkflowServiceTest) just with one simple method. The build still fails. I suspect there is a problem with the Travis-CI environment

```
@Test
    public void testingCrunchifyAddition() {
        assertEquals("Here is test for Addition Result: ", 30, addition(27, 3));
    }



    public int addition(int x, int y) {
        return x + y;
    }
```

jvemugunta:
@apanicker-nflx - I think the problem is with WorkflowLegacyMigrationTest and WorkflowServiceTest sharing the same TestRunner class resulting into this race condition

I have made WorkflowLegacyMigrationTest as a simple test independent of TestRunner. the build succeeds

Can you look and tell me if you have insight on the rootcause?

cyzhao:
@jvemugunta please resolve the conflict then we will merge this.

jvemugunta:
@cyzhao - The build fails because of WorkflowLegacyMigrationTest and I don't own this class.
I am not sure who the owner of the class. I can delete this class to pass all the tests. Let me know

jvemugunta:
@cyzhao - WorkflowLegacyMigrationTest.java is the culprit and @apanicker-nflx is the contributor to this test case. Once I removed this test case, all builds are clean along with all the test cases.

apanicker-nflx:
@jvemugunta The class `WorkflowLegacyMigrationTest.java` is actually used to test the case  where existing workflows within conductor are not impacted by the major version upgrade. As you can see within this class, @josedab added tests such that existing workflow executions are read from resources and used within the tests.

For now, since it looks like the addition of the MySQLTestModule and accompanying test cases are what caused this regression, we could `@Ignore` the `MySQLWorkflowServiceTest` and get your changes merged in and try to fix the tests in a later PR.

josedab:
Thanks @apanicker-nflx for jumping on this :)
Just my 2 cents on that file:
- I would not delete this file as I believe it still helps to ensure stored vs ephemeral workflows are run in the same way.
- I think it is ok to add `@Ignore` to it because:
  - This is a use case that only affects those with running workflows that cannot be stopped with the code update.
  - It should not block improvements on the current code
Having said that, i will add this to my todo list in case we can make it run in a near future.

jvemugunta:
@apanicker-nflx and @josedab - I have ignore MySQL TEST case for the time being to pass the builds. It surprises the race condition Gradle causes build failures with Travis-CI and the same works on local machine

There should be some environment difference that could be causing this issue. However, you should be able to merge the code now

jdoupe:
Is the removal of "SELECT ... FOR UPDATE" in PEEK_MESSAGES on purpose?  Without it, this causes the occasional duplicate processing of tasks when running multiple instances of conductor.

I must say that it is not as easy as just add "FOR UPDATE" to the select statement, because that ends up essentially locking the whole table if you don't provide an appropriate LIMIT.  The default is to "LIMIT 10", and if there are less than 10 records, this will prohibit not only updates to the locked rows, but also other rows as well as inserts to the table as a whole (MySQL 5.7).  I've circumvented this by doing a SELECT COUNT within peekMessages() and using the resulting count instead of the passed in value if it is less.  Still a little bit of a race condition, but better than the alternatives.

If there's not a reason, I can submit a patch, or feel free to convert my description into code.

PR #772, opened 2018-09-17
add responseTimeoutSeconds field validation

alexnetflix:
enforcing responseTimeoutSeconds field to be not null whenever task definition gets created or updated (default is 1h)

### DIFF
--- a/core/src/main/java/com/netflix/conductor/service/MetadataService.java
+++ b/core/src/main/java/com/netflix/conductor/service/MetadataService.java
@@ -18,6 +18,8 @@
  */
 package com.netflix.conductor.service;
 
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
 import com.netflix.conductor.annotations.Trace;
 import com.netflix.conductor.common.metadata.events.EventHandler;
 import com.netflix.conductor.common.metadata.tasks.TaskDef;
@@ -61,6 +63,11 @@ public class MetadataService {
             taskDefinition.setCreateTime(System.currentTimeMillis());
             taskDefinition.setUpdatedBy(null);
             taskDefinition.setUpdateTime(null);
+
+            ServiceUtils.checkNotNull(taskDefinition,"TaskDef object cannot be null");
+            ServiceUtils.checkNotNull(taskDefinition.getName(),"TaskDef name cannot be null");
+            ServiceUtils.checkArgument(taskDefinition.getResponseTimeoutSeconds()>0, "ResponseTimeoutSeconds must be positive");
+
             metadataDAO.createTaskDef(taskDefinition);
         }
     }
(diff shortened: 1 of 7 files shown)


mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLMetadataDAO.java
```diff
@@ -263,6 +263,7 @@ public void removeEventHandlerStatus(String name) {
     private void validate(TaskDef taskDef) {
         Preconditions.checkNotNull(taskDef, "TaskDef object cannot be null");
         Preconditions.checkNotNull(taskDef.getName(), "TaskDef name cannot be null");
+        Preconditions.checkNotNull(taskDef.getResponseTimeoutSeconds(), "ResponseTimeoutSeconds cannot be null");
```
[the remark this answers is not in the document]
cyzhao: Agreed, move all precondition checks to service level and use ServiceUtil instead.

cyzhao: Also check responseTimeoutSeconds must be greater than 0 (i.e. it can't be 0 or negative)

alexnetflix:
@cyzhao @falu2010-netflix  I moved all validation from dao to services for this class(not just for responsetimeout) and created a separate MetadataServiceTest class

core/src/test/java/com/netflix/conductor/service/MetadataServiceTest.java
```diff
@@ -0,0 +1,142 @@
+package com.netflix.conductor.service;
```
falu2010-netflix: Awesome 👍

redis-persistence/src/main/java/com/netflix/conductor/dao/dynomite/RedisMetadataDAO.java
```diff
@@ -80,10 +80,6 @@ public String updateTaskDef(TaskDef taskDef) {
     }
 
     private String insertOrUpdateTaskDef(TaskDef taskDef) {
```
falu2010-netflix: Is this API being used from other resource then Metadataservice? If, yes then you might want to keep the code.

core/src/main/java/com/netflix/conductor/service/MetadataService.java
```diff
@@ -233,7 +242,8 @@ public void removeEventHandlerStatus(String name) {
         return metadataDAO.getEventHandlersForEvent(event, activeOnly);
     }
```
falu2010-netflix: You might want to add validation for event.

redis-persistence/src/main/java/com/netflix/conductor/dao/dynomite/RedisMetadataDAO.java
alexnetflix: it's not

mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLMetadataDAO.java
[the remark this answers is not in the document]
alexnetflix: ok, but in the future, we should prob separate refactoring from the functional changes

core/src/main/java/com/netflix/conductor/service/MetadataService.java
alexnetflix: good catch, there are more methods in dao that could use extra validation and tests, will clean up

Issue #764, opened 2018-09-11
Response of api updating a task has been changed.


shouhei:
After [this commit](https://github.com/Netflix/conductor/commit/ae403c9a0a43b7a525dcf0012fc9b080cf44ba8e#diff-7143b9a69e2e1c1c9127eada224a0920L147), Response of api updating a task  has changed from json string  to plain text string. In its impact python example script `kitchensink_workers.py` has been raising some exceptions.  This change has intend ?

before:

> "7efe673f-0c3c-4f60-bdd5-b4d17590d093"

after:

> 7efe673f-0c3c-4f60-bdd5-b4d17590d093

kishorebanala:
Apologies for the inconvenience. We use Java client, and Python client is completely managed by the community. Hence, it's really hard for us to track breaking changes in Python client. We'll be more careful in reviewing breaking changes hereon. But, please feel free to send a PR with patch if something has to be fixed immediately.

PR #774, opened 2018-09-19
Fix Python client task update procedure 

kbristow:
As raised in #764 the task update API response has changed and no longer returns a JSON formatted string. This breaks the Python client when it tries to parse the response as a JSON response. 

This change updates the task POST endpoint to allow `'Accept': 'text/plain'` headers and implements the corresponding change in the Python client so that it does not expect a JSON response. I have left the `MediaType.APPLICATION_JSON` as valid response type so that any clients that are currently expecting such a response do not break with the change assuming they are able to parse the non JSON formatted string.

### DIFF
--- a/client/python/conductor/conductor.py
+++ b/client/python/conductor/conductor.py
@@ -171,7 +171,8 @@ class TaskClient(BaseClient):
 
     def updateTask(self, taskObj):
         url = self.makeUrl('')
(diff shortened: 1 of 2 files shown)


MLauri:
Thank you!
This problem seems also to affect the Conductor UI.

PR #784, opened 2018-09-26
fix(test): updated workflow and task name being used in test

falu2010-netflix:
Description:
There were some tests which kept failing due task being shared among different workflow. Updated these tests to use different task name to avoid concurrency issue.


cyzhao:
@falu2010-netflix please resolve conflict then merge.

PR #803, opened 2018-10-09
Adds ES5 REST index DAO for conductor 1.12.x

bitbier:
With ES5 and later, there is a big push to use the REST client rather than the TCP transport client.
Not only that, but some services such as AWS ElasticSearch only support the HTTP REST client. This
is an attempt to add the ES 5 REST client for conductor. You would configure this by setting
`workflow.elasticsearch.version=5-rest` and also setting the `workflow.elasticsearch.url=http://<HOST>:<PORT>`.

You can set multiple hosts by adding separating them with a comma.

I was going to push this into v2.0, but I'm unclear how stable that is and we are currently using 1.12 right now at our work. I'm going to be doing more testing with these changes at our company to make sure they work, but this is at least a start that incorporates and configures the REST client for ES5.


bitbier:
I have not done Java work for a while, so not sure if this is the appropriate approach. I basically went through [ElasticSearchDAOV5](https://github.com/Netflix/conductor/blob/1.12/es5-persistence/src/main/java/com/netflix/conductor/dao/es5/index/ElasticSearchDAOV5.java) and ported everything over to use the HTTP rest client.

You can find more information on the REST client here:
https://www.elastic.co/guide/en/elasticsearch/client/java-rest/5.6/index.html

I also tried to reuse when I could so this new module depends on `conductor-es5-persistence` module for the query parser and the EmbeddedElasticSearchV5.

bitbier:
@malpani  Ah good catch, just been working on the actual module. There are still some additional improvements that need to happen. For instance, if you are using AWS ElasticSearch with IAM rules or non-VPC version, you need to add auth headers and also encode the request messages so that they work, but those can be added on later.

I'm still in the middle of improving my company's CI build so that it builds the conductor jars so we can actually deploy it in our testing environment. Right now my CI process relies on the published jars; so I can't do a proper integration test, but hope to get it done today and will get your patch integrated soon. All in all, hope to be done by Friday on this.

es5-rest-persistence/README.md
```diff
@@ -0,0 +1,15 @@
+# ES5 REST persistence
+
+This module provides ES5 REST persistence instead of relying on the Transport protocol.
+
+## Usage
+
+To use this module, set the following values in your conductor configuration:
```
kishorebanala: Apart from below configuration, This module also has to be injected into Server module (Preferably replacing existing ES5 module, so to avoid dependency conflicts. Before Conductor V2.0, ES2 is shadowed to avoid this).

core/src/test/java/com/netflix/conductor/core/execution/TestConfiguration.java
```diff
@@ -17,6 +17,7 @@
 
 import com.netflix.conductor.core.config.Configuration;
 
+import java.util.HashMap;
```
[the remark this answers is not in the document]
bitbier: ah you're right. I got lost in my IDE and edited this file instead of the TestConfiguration I was using for my tests. Will remove.

es5-rest-persistence/README.md
bitbier: So right now, I use a different package name for the ElasticSearchModuleV5Rest. It's under the `com.netflix.conductor.dao.es5rest`. This is because the es5rest depends on the `conductor-es5-persistence` for the `query.parser` logic as well as the `EmbeddedElasticSearchV5`. Because of the way gradle ties the dependencies of the various modules together (via the versionsOfDependencies.gradle file); there shouldn't be a dependency issue as they are relying on the same ES version.

That being said... maybe I should instead push the REST version of the IndexDAO inside the `conductor-es5-persistence` project? I don't have much experience with guice, but it seems like I can simply pass a REST flag to the `ElasticSearchModuleV5()` and then bind the IndexDAO to the rest version if the flag exists?

kishorebanala: Actually, that makes much better sense. There is no reason to maintain a separate module, if we can have ElasticSearchModuleV5Rest within module. The server module can still install one of them based on config.

bitbier: ok sounds good. This should be an easy change to make. Will get it up later today.

bitbier: Ok this is done the REST DAO is now inside the es5-persistence module now.

bitbier:
I've moved the REST DAO over to the ES5 persistence module instead of it being in it's own separate module. To activate the REST DAO, you need to set the `workflow.elasticsearch.transport=rest`. This should configure the rest DAO instead of using the default tcp transport.

I also updated the README to point out the other options that are available to configure this module:

https://github.com/Netflix/conductor/blob/e5803dafad9ff30a25243e01ef754a7a87543ed6/es5-persistence/README.md

bitbier:
So, we are currently running this in our testing environment and will push up to production by the end of this week if things look good. Not sure if there is anything additional that I should do in this PR?

Any other comments? Otherwise, it would be nice to get this code in.

kishorebanala:
@bitbier Awesome, it looks good. However, I just noticed you were trying to merge this into 1.12; We're actively working on 2.0 right now, and this should go to `dev` branch. Would you mind changing it to `dev` branch please. Let me know if you need any help in doing that.

bitbier:
@kishorebanala The reason why I'm going to 1.12 is because that's currently the version that we are using and as far as I see 2.0 is not released yet.

I'm happy to create another branch for `dev` so that this can get pushed to 2.0, but right now 2.0's directory structure is not compatible with 1.12, so that would require two different PRs. I still think this may be valuable to others using 1.12.x still.

bitbier:
@kishorebanala any word on being able to patch this into 1.12? I also created another PR to patch this into Conductor 2.0. The pattern is a little different, but I was able to port most of what I wrote here to there.

https://github.com/Netflix/conductor/pull/843

kishorebanala:
@bitbier Sorry for delays, I've been sidetracked. I'll get to this today.

kishorebanala:
Released this as v1.12.4
v2.0 changes are still pending CodeCov cleanup.

JasperB:
@bitbier It's really helpful for us, thank you so much to apply this. In this way we already migrated it to AWS ES without native port 9300. but during the migration we met one problem is we're use the ES 5.3 which will caused the conduct UI cannot work well, because the REST high level API only supported from 5.6 and later, it's really took some time to figure it out, seems I missed you 5.6 doc in the beginning haha.. but this just a reminder for other guys.

PR #804, opened 2018-10-10
fix TaskSummary Serializing, missed no-arguments constructor

wefine:
In `TaskSummary` class,  missed no-arguments constructor, but it is necessary for jersy `getEntity` method to parse to object. If missed, a exception broken out, for example:
```
Caused by: com.sun.jersey.api.client.ClientHandlerException: 
com.fasterxml.jackson.databind.JsonMappingException: 
Can not construct instance of com.netflix.conductor.common.run.TaskSummary: 
no suitable constructor found, can not deserialize from Object value (missing default constructor or creator, or perhaps need to add/enable type information?)
```

So, add no-arguments constructor to resolve it .

### DIFF
--- a/common/src/main/java/com/netflix/conductor/common/run/TaskSummary.java
+++ b/common/src/main/java/com/netflix/conductor/common/run/TaskSummary.java
@@ -86,6 +86,9 @@ public class TaskSummary {
 	@ProtoField(id = 16)
 	private String taskId;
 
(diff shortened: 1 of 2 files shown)


wefine:
@apanicker-nflx Hi, I reabased  my fix-TaskSummary-serialize-for-client. :)

Issue #833, opened 2018-10-24
Failure to start conductor server after upgrading to release version v2.0.0


wesleysong9:
Hi Conductor team,

Previously we use the release version v1.12.3, when starting the conductor server by gradle task conductor:server server with our customized config, it works fine. We have our customized configuration to connect to our redis cluster and ElasticSearch. 

Below attached our customized configuration file for the conductor server(Using redis cluster and elasticsearch)

server.properties

db=redis_cluster
workflow.dynomite.cluster.hosts=reconciliation-ecc.nquffl.clustercfg.apn2.cache.amazonaws.com:6379:ap-northeast-2c
workflow.dynomite.cluster.name=reconciliation-ecc
workflow.dynomite.connection.maxConnsPerHost=31
workflow.namespace.prefix=conductor
workflow.namespace.queue.prefix=conductor_queues
queues.dynomite.threads=10
queues.dynomite.nonQuorum.port=6379
workflow.elasticsearch.instanceType=external
workflow.elasticsearch.url=10.213.0.252:9300
workflow.elasticsearch.index.name=conductor
workflow.elasticsearch.version=5
EC2_AVAILABILITY_ZONE=ap-northeast-2c

But we find today that there is a new release version of v2.0.0, so we upgraded the conductor version to the latest one. After upgraded, we meet the below issue when start the conductor server.

Error message:
Exception in thread "main" com.google.inject.CreationException: Unable to create injector, see the following errors:

1) No implementation for com.netflix.dyno.queues.ShardSupplier was bound.
  while locating com.netflix.dyno.queues.ShardSupplier
    for the 3rd parameter of com.netflix.conductor.dyno.RedisQueuesProvider.<init>(RedisQueuesProvider.java:32)
  at com.netflix.conductor.dao.RedisWorkflowModule.configure(RedisWorkflowModule.java:33)

2) No implementation for redis.clients.jedis.JedisCommands annotated with @com.google.inject.name.Named(value=DynoReadClient) was bound.
  while locating redis.clients.jedis.JedisCommands annotated with @com.google.inject.name.Named(value=DynoReadClient)
    for the 2nd parameter of com.netflix.conductor.dyno.RedisQueuesProvider.<init>(RedisQueuesProvider.java:32)
  at com.netflix.conductor.dao.RedisWorkflowModule.configure(RedisWorkflowModule.java:33)

2 errors
	at com.google.inject.internal.Errors.throwCreationExceptionIfErrorsExist(Errors.java:470)
[...]

PR #835, opened 2018-10-25
fix compatibility with 1.x

apanicker-nflx:


common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowTask.java
```diff
@@ -13,33 +13,60 @@
...
  */
 @ProtoMessage
 public class WorkflowTask {
 
+	/**
+	 * This field is deprecated and will be removed in the next version.
+	 * Please use {@link TaskType} instead.
+	 */
+	@Deprecated
+    public enum Type {
```
cyzhao: How will this fix the issue of backward compatibility if this isn't used anywhere?

cyzhao: Overload setType method to take WorkflowTask.Type as input and convert it to TaskType, so workers that create dynamic tasks will work with latest client.

ChengZheBai:
@cyzhao could you help to look at issue of #833 , for V2 release, it does have backwards compatibility issue, really appreciate your help on this.

common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowTask.java
apanicker-nflx: overloaded getWorkflowTaskType() method and added @deprecated annotation to it.

apanicker-nflx: > overloaded setWorkflowTaskType() method and added @deprecated annotation to it.

i take that back, cannot add an overloaded setter otherwise the object mapper fails

PR #838, opened 2018-10-25
Converts index templates to ES5 syntax and upgrade to ES5 in Docker

davidwadden:
With the [v2 release](https://github.com/Netflix/conductor/releases/tag/v2.0.0), some of the configurations are not in sync with the upgrade to Elasticsearch 5.

> This PR is based against `dev` branch which is not up-to-date with `master`.  I didn't want to submit the fast-forward as part of this merge.  Will rebase this PR once `dev` is up-to-date with `master`.

This PR does the following:

- Syncs up the default `docker-compose.yml` to start ES 5.6.8 (identical to [`docker-compose.dependencies.yaml`](https://github.com/Netflix/conductor/blob/master/docker/grpc/docker-compose.dependencies.yaml#L21-L33)
- Removes references the deprecated ES 2 components and templates (as well as the templates themselves)
- Converts the templates to replace the deprecated `string` type per https://www.elastic.co/blog/strings-are-dead-long-live-strings
  - Removes `include_in_app` directive in favor of default (based on boolean value of `index`)
    - https://www.elastic.co/guide/en/elasticsearch/reference/6.2/breaking_60_mappings_changes.html#_the_literal_include_in_all_literal_mapping_parameter_is_now_disallowed
    - https://www.elastic.co/guide/en/elasticsearch/reference/5.4/include-in-all.html#include-in-all
- Provide default values for Hikari JDBC pool configs for the `mysql` provider when run inside of Docker
  - `conductor.mysql.connection.pool.size.max=10`
  - `conductor.mysql.connection.pool.idle.min=2`

Didn't see much in terms of integration tests to cover these changes.  I did exercise the system against ES and reviewed `DEBUG` level logging on ES to verify the deprecated warnings were no longer there and the system generally still worked.

Happy to incorporate any recommendations or feedback for this PR


artpdr:
Hi @davidwadden can you run conductor + dynomite + elasticsearch using docker-compose with the configurations that you provide in this PR? I lost some time trying to run conductor + dynomite + elasticsearch in my machine using the docker-compose after making some changes to it but conductor is much more slower than when I run it without dynomite and elasticsearch in different containers, as I mentioned here: #839.

You just need to apply these changes to run everything correctly or have you done other changes?

Thanks!

davidwadden:
@artpdr This PR doesn't directly address the startup problem.  Check out [`/docker/grpc`](https://github.com/Netflix/conductor/tree/master/docker/grpc) folder for an example of how to delay the startup of `conductor-server` until after the data services have started (MySQL in that case).  I'll take a look at updating this PR to follow that same model to address startup delay.

The other solution is just to restart the container after `elasticsearch` has finished initializing using `docker start docker_conductor-server_1` 

I can't speak to performance difference.  But you're comparing in-memory Redis with `dynomite` so it's not really apples to apples.

davidwadden:
I utilized the [`healthcheck`](https://docs.docker.com/compose/compose-file/compose-file-v2/#healthcheck) directive with the [`depends_on#service_healthy`](https://docs.docker.com/compose/compose-file/compose-file-v2/#depends_on) to ensure the port was open on `elasticsearch`, `mysql` and `dynomite` prior to starting `conductor-server`.

- Consolidated `docker-compose.dependencies.yaml` and `docker-compose.apps.yaml` under `grpc` Docker definition by leveraging the same feature.
- Generally made the two `docker-compose` definitions more similar.

artpdr:
Thanks @davidwadden.

I will try to run the system in docker using your suggestions.

I know that I can't expect to have the same performance with in-memory Redis and Dynomite. However, I thin that the loss of performance with Dynomite is excessive. 

For example in a workflow with 10 parallel tasks that I set just for tests and in which the workers just set the task as COMPLETED and do not do any other work. I can have a worker pulling the task every second in the in-memory Redis version but not in the dynomite version where it was taking approximately 6 seconds or more to pull and work on the task... This is a very strange behavior and I need to inspect better what is happening. I can be using a bad configuration... I don't know.

artpdr:
@davidwadden just to let you know that the docker-compose you provided is working perfectly.
In my docker-compose.yaml I changed the timeout command to have a larger timeout in the healtcheck of dynomite and elasticsearch: "test: timeout 5 bash -c ...", as the elasticsearch wasn't starting prior to conductor server.
Thanks :)

davidwadden:
@artpdr Thanks for testing it out!  I revised the timeouts per your recommendation to not constrain the timeout to 1sec and to reduce the interval from `2s` to `5s`.

Also rebased to `dev` as of `v2.1.0` pre-release

docker/server/config/config-mysql-grpc.properties
```diff
@@ -13,6 +13,9 @@ db=mysql
 
 jdbc.url=jdbc:mysql://mysql:3306/conductor
 
+conductor.mysql.connection.pool.size.max=10
```
kishorebanala: Nice Work @davidwadden, thanks. Is there a reason to set mysql parameters explicitly? If so, and if it is configurable, shall we add comments mentioning the same.

davidwadden: @kishorebanala On `master`, building the gRPC docker images under [`/docker/grpc`](https://github.com/Netflix/conductor/tree/master/docker/grpc) results in `conductor-server` aborting startup with the following logs (also reproduced outside of Docker):

https://gist.github.com/davidwadden/6590bf56ff01a0aca2ec7639fd63dab7

Looking at [`MySQLConfiguration.java`](https://github.com/Netflix/conductor/blob/v2.1.0/mysql-persistence/src/main/java/com/netflix/conductor/mysql/MySQLConfiguration.java#L25-L30), which to the commenter's point [are the `HikariConfig` defaults](https://github.com/brettwooldridge/HikariCP/blob/HikariCP-3.2.0/src/main/java/com/zaxxer/hikari/HikariConfig.java#L111-L112), the default values of `-1` prevent startup on Conductor.

The values of `2` for `minIdle` and `10` for `maxPoolSize` were arbitrary values.

kishorebanala: Great, would you mind adding the same in comments please. It was not clear as to why these values are chosen in first look.

PR #846, opened 2018-10-31
Feature/add workflow completion listener al

alexnetflix:
Adding a listener for when workflow gets to a COMPLETED or TERMINATED state.
This listener could be redefined in projects that use or inherit conductor library with some custom implementation as needed.

WorkflowDefinition has a flag to trigger these listeners so that we do not trigger these for every workflow.

### DIFF
--- a/common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowDef.java
+++ b/common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowDef.java
@@ -67,7 +67,7 @@ public class WorkflowDef extends Auditable {
 	private boolean restartable = true;
 
 	@ProtoField(id = 10)
-	private boolean terminalWorkflowListenerEnabled = false;
+	private boolean workflowStatusListenerEnabled = false;
 
 	/**
 	 * @return the name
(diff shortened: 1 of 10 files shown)


core/src/main/java/com/netflix/conductor/core/execution/WorkflowStatusListenerStub.java
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Stub listener default implementation
+ */
+public class WorkflowStatusListenerStub implements WorkflowStatusListener {
+
+    private static final Logger LOG = LoggerFactory.getLogger(WorkflowStatusListenerStub.class);
+
+    @Override
+    public void onWorkflowCompleted(Workflow workflow) {
+        LOG.debug("Workflow {} is completed", workflow.toString());
```
cyzhao: only log workflowId

common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowDef.java
```diff
@@ -66,6 +66,9 @@
 	@ProtoField(id = 9)
 	private boolean restartable = true;
 
+	@ProtoField(id = 10)
+	private boolean terminalWorkflowListenerEnabled = false;
```
cyzhao: consider naming it workflowStatusListenerEnabled to be consistent with interface name, also be generic as we should also notify on other status changes, e.g. when a workflow is restarted.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowStatusListenerStub.java
alexnetflix: done

common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowDef.java
alexnetflix: renamed

PR #847, opened 2018-11-02
Donot return task ack failed

apanicker-nflx:
Do not return the task from the client to the server if the ack fails or throws an exception. The task will be available to be polled again in this case after the unack timeout. 
The catch all in the task service ensures that even if the ack fails on the server side due to an exception with the queueDAO, the client sees an ack failure and does not process the task.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/service/TaskService.java
+++ b/core/src/main/java/com/netflix/conductor/service/TaskService.java
@@ -152,7 +152,16 @@ public class TaskService {
     public String ackTaskReceived(String taskId, String workerId) {
         ServiceUtils.checkNotNullOrEmpty(taskId, "TaskId cannot be null or empty.");
         LOGGER.debug("Ack received for task: {} from worker: {}", taskId, workerId);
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/netflix/conductor/service/TaskService.java
```diff
@@ -152,7 +152,14 @@ public String updateTask(TaskResult taskResult) {
     public String ackTaskReceived(String taskId, String workerId) {
         ServiceUtils.checkNotNullOrEmpty(taskId, "TaskId cannot be null or empty.");
         LOGGER.debug("Ack received for task: {} from worker: {}", taskId, workerId);
-        return String.valueOf(executionService.ackTaskReceived(taskId));
+        boolean ackResult;
+        try {
+            ackResult = executionService.ackTaskReceived(taskId);
+        } catch (Exception e) {
```
cyzhao: Write a comment explaining rational behind swallowing exception here.

apanicker-nflx: done

PR #849, opened 2018-11-03
use two-stage building for conductor:server

houz42:
With two-stage building, we do not need a separate 'builder' image anymore. It will make the dockerfile   easy to read and maintain, and keep the resulting image 'optimized'.


cyzhao:
@houz42 please explain why these changes are needed. Please always provide a description in the PR moving forward.

houz42:
Sorry for the late response

docker/server/Dockerfile
```diff
@@ -1,3 +1,17 @@
...
+FROM openjdk:8-jdk AS builder
+
+MAINTAINER Netflix OSS <conductor@netflix.com>
+
+# Copy the project directly onto the image
+COPY . /conductor
+WORKDIR /conductor
+
+# Build the server on run
+ENTRYPOINT ./gradlew build -x test
```
cyzhao: Should be `RUN ./gradlew build -x test`

docker/server/Dockerfile
```diff
@@ -1,3 +1,17 @@
+#
+# conductor:server - Netflix conductor server
+#
```
cyzhao: Please update the comment to explain this stage.

houz42: How about comment like this:
> 0. builder stage
> 1. bin stage

houz42:
@cyzhao how can I fix the coverage error?

docker/server/Dockerfile
cyzhao: Sounds good

cyzhao:
> @cyzhao how can I fix the coverage error?

Don't worry about it given your PR doesn't touch any Java code.

PR #855, opened 2018-11-07
Adds support for redis sentinel deployments

RuriRyan:
New RedisSentinelJedisProvider class which uses the
JedisSentinelPool to do all the magic.
Also a new config option to choose the redis_sentinel as DB type.

### DIFF
--- a/redis-persistence/src/main/java/com/netflix/conductor/jedis/RedisSentinelJedisProvider.java
+++ b/redis-persistence/src/main/java/com/netflix/conductor/jedis/RedisSentinelJedisProvider.java
@@ -1,0 +1,53 @@
+package com.netflix.conductor.jedis;
+
+import com.google.common.collect.Lists;
+
+import com.netflix.conductor.dyno.DynomiteConfiguration;
+import com.netflix.dyno.connectionpool.Host;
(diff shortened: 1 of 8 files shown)


stevenctl:
Do you have an example of the configuration and redis deployment you would use with this?

redis-persistence/src/main/java/com/netflix/conductor/jedis/RedisSentinelJedisProvider.java
```diff
@@ -0,0 +1,53 @@
...
+        jedisPool = new JedisSentinelPool(configuration.getClusterName(), sentinels, poolConfig);
+    }
+
+    @Override
+    public JedisCommands get() {
+        return jedisPool.getResource();
+    }
+
+    @Override
+    protected void finalize() throws Throwable {
```
cyzhao: Don't rely on finalize to close the pool, as a matter of fact, finalize is deprecated in Java 9 and newer versions.

cyzhao:
Please add unit tests.

RuriRyan:
I will add unit tests later. But I still would like to hear some opinions about the implementation.

apanicker-nflx:
Please add some comment or ReadMe about the configuration that is to be used with this deployment.

PR #861, opened 2018-11-09
fix task data copied for retries

apanicker-nflx:


core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -467,6 +465,9 @@ private Task taskToBeRescheduled(Task task) {
         taskToBeRetried.setRetriedTaskId(task.getTaskId());
         taskToBeRetried.setStatus(SCHEDULED);
         taskToBeRetried.setRetryCount(task.getRetryCount() + 1);
+        taskToBeRetried.setRetried(false);
```
cyzhao: Does it make sense to fix Task.copy(...) method instead?

apanicker-nflx: The Task.copy() method is used to create a deep copy of the task object. Its primary usage is when creating a clone of a workflow object with the tasks copied over, at this time, all the information needs to be copied over, so I wouldn't want to change that and explicitly set these values here.

PR #864, opened 2018-11-13
remove fallback to default domain

apanicker-nflx:
Removed the fallback to the default domain if an active domain is not found.
Changed this such that it falls back to the lowest priority domain specified in the list.

### DIFF
--- a/common/src/main/java/com/netflix/conductor/common/metadata/tasks/PollData.java
+++ b/common/src/main/java/com/netflix/conductor/common/metadata/tasks/PollData.java
@@ -1,8 +1,4 @@
-package com.netflix.conductor.common.metadata.tasks;
-
-import com.github.vmg.protogen.annotations.*;
-
-/**
+/*
  * Copyright 2016 Netflix, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -17,19 +13,26 @@ import com.github.vmg.protogen.annotations.*;
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+package com.netflix.conductor.common.metadata.tasks;
+
+import com.github.vmg.protogen.annotations.ProtoField;
+import com.github.vmg.protogen.annotations.ProtoMessage;
+
+import java.util.Objects;
+
 @ProtoMessage
 public class PollData {
 	@ProtoField(id = 1)
-	String queueName;
+	private String queueName;
 
 	@ProtoField(id = 2)
-	String domain;
+	private String domain;
 
 	@ProtoField(id = 3)
-	String workerId;
+	private String workerId;
 
 	@ProtoField(id = 4)
-	long lastPollTime;
+	private long lastPollTime;
 	
 	public PollData() {
 		super();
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -1062,16 +1062,18 @@ private void setTaskDomains(List<Task> tasks, Workflow wf) {
...
+        // In sequence check if any worker has polled for last 30 seconds, if so that is the Active domain
+
         return Arrays.stream(domains)
                 .map(domain -> executionDAO.getPollData(taskType, domain.trim()))
                 .filter(Objects::nonNull)
                 .filter(validateLastPolledTime)
                 .findFirst()
                 .map(PollData::getDomain)
-                .orElse(null);
+                .orElse(domains[domains.length - 1].trim());
```
kishorebanala: Would it be possible to throw ArrayIndexOutOfBoundsException when domains length is 0?

common/src/main/java/com/netflix/conductor/common/metadata/tasks/PollData.java
```diff
@@ -102,8 +107,7 @@ public synchronized boolean equals(Object obj) {
 		} else if (!queueName.equals(other.queueName))
 			return false;
 		if (workerId == null) {
-			if (other.workerId != null)
-				return false;
+			return other.workerId == null;
```
alexnetflix: consideration: I like Jave 7's  "if(!Objects.equals(x,y) return false" statements, a bit more concise, handles nulls, and leaves less room for error if somebody adds more checks at the end of the method for example and other.workerId == null returns true

cyzhao: I agree, let's regenerate equals() and hashCode(), also remove synchronized keyword from these methods.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -1062,16 +1062,18 @@ private void setTaskDomains(List<Task> tasks, Workflow wf) {
         }
     }
 
-    private String getActiveDomain(String taskType, String[] domains) {
+    @VisibleForTesting
+    String getActiveDomain(String taskType, String[] domains) {
```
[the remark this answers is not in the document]
apanicker-nflx: https://github.com/Netflix/conductor/pull/864/files#diff-a28de3002a75934bb7c693d03b674144L1036
This is the only method accessing this function, null values cannot be passed as domains here. I will update the javadoc with the same.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
apanicker-nflx: As mentioned above, it is guaranteed that the domains array is atleast of size 1 when passed as param into the function. I will update the javadoc.

core/src/test/java/com/netflix/conductor/core/execution/TestWorkflowExecutor.java
```diff
@@ -522,7 +526,27 @@ public void testRetryWorkflow() {
         assertEquals(2, updateWorkflowCalledCounter.get());
         assertEquals(7, updateTasksCalledCounter.get());
         assertEquals(1, removeQueueEntryCalledCounter.get());
+    }
+
+    @Test
+    public void testGetActiveDomain() {
```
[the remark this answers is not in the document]
apanicker-nflx: Added single empty domain test case.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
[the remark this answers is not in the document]
apanicker-nflx: Added checks and test cases

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
apanicker-nflx: Added checks and tests

PR #870, opened 2018-11-14
added option to restart with latest definitions

apanicker-nflx:


test-harness/src/test/java/com/netflix/conductor/tests/integration/AbstractWorkflowServiceTest.java
```diff
@@ -3035,7 +3035,7 @@ public void testRestart() {
         assertNotNull(es);
         assertEquals(WorkflowStatus.FAILED, es.getStatus());
 
-        workflowExecutor.rewind(es.getWorkflowId());
+        workflowExecutor.rewind(es.getWorkflowId(), false);
```
alexnetflix: Is it worth throwing in another Int test with "true" value here??

apanicker-nflx: added

PR #898, opened 2018-12-14
Add workflow creation form

haoch:
screen shot 2018-12-14 at 8 14 35 am


kishorebanala:
@haoch This is a great start, but seems out of place. I'm not sure if this PR helps create new Workflow from UI. If so, that's great. But ideally, the flow should be a Create task and workflow button in Metadata section of UI, and a form with all the options, which on submit would register a Workflow definition. If this is not what you've intended to do, can you please describe you plan?

Also, I'm closing this for now as it has been a while. Please feel let me know if you're still working on this, and I'd be glad to reopen.

haoch:
@kishorebanala The button is to trigger a workflow instance instead of registering workflow definition, as it would be very useful to create workflow through UI instead of API.

haoch:
@kishorebanala as we are using conductor and needs the function for daily operations, could you add a similar feature?

PR #902, opened 2018-12-14
Removed date range from the query to search for archivable workflows.

s3nthil:


es2-persistence/src/main/java/com/netflix/conductor/dao/es/index/ElasticSearchDAO.java
```diff
@@ -499,19 +498,21 @@ public String get(String workflowInstanceId, String fieldToGet) {
...
 				.setTypes("workflow")
 				.setQuery(q)
-				.setSize(1000);
+				.addSort("endTime", SortOrder.ASC)
+				.setSize(archiveSearchBatchSize);
 
 		SearchResponse response = s.execute().actionGet();
 
 		SearchHits hits = response.getHits();
+		logger.info("Archive search totalHits - {}", hits.getTotalHits());
```
alexnetflix: would it be helpful to have a metric for this one? To be able to catch it if the process returns no data for long time?

PR #916, opened 2018-12-20
fix for remove task in C* dao

apanicker-nflx:


cassandra-persistence/src/main/java/com/netflix/conductor/dao/cassandra/CassandraExecutionDAO.java
```diff
@@ -536,6 +536,11 @@ private void removeTask(Task task) {
         }
     }
 
+    private void removeTaskLookup(Task task) {
+        recordCassandraDaoRequests("removeTaskLookup", task.getTaskType(), task.getWorkflowType());
```
alexnetflix: do you need both "removeTaskLookup" and "removeTask" with your recordCassandraDaoRequests? Feels like  line 525 is not needed

apanicker-nflx: `removeTask` can be called just to remove a task from the datastore.
`removeTaskLookup` can be called when removing a workflow from the datastore.

Either ways, these are 2 separate calls to the datastore and would be helpful to collect metrics for each of them.

PR #924, opened 2019-01-02
set consistency levels based on config

apanicker-nflx:


cassandra-persistence/src/main/java/com/netflix/conductor/dao/cassandra/CassandraExecutionDAO.java
```diff
@@ -78,23 +79,23 @@
     public CassandraExecutionDAO(Session session, ObjectMapper objectMapper, CassandraConfiguration config, Statements statements) {
         super(session, objectMapper, config);
 
-        this.insertWorkflowStatement = session.prepare(statements.getInsertWorkflowStatement());
-        this.insertTaskStatement = session.prepare(statements.getInsertTaskStatement());
+        this.insertWorkflowStatement = session.prepare(statements.getInsertWorkflowStatement()).setConsistencyLevel(config.getWriteConsistencyLevel());
+        this.insertTaskStatement = session.prepare(statements.getInsertTaskStatement()).setConsistencyLevel(config.getWriteConsistencyLevel());
```
alexnetflix: can it be done at a cluster level rather than per individual read\write?

apanicker-nflx: I don't believe there is a way to set consistency for read and writes separately at cluster level.

PR #929, opened 2019-01-05
Bugfix/support env variable task input param

falu2010-netflix:
Description:

- Added validation for environment variable used in WorkflowDef Task InputParam
- Added validation for HTTP task such that http_request can be a part of inputTemplate or inputParameters.

### DIFF
--- a/common/src/main/java/com/netflix/conductor/common/utils/EnvUtils.java
+++ b/common/src/main/java/com/netflix/conductor/common/utils/EnvUtils.java
@@ -6,8 +6,7 @@ public class EnvUtils {
     public enum SystemParameters {
         CPEWF_TASK_ID,
         NETFLIX_ENV,
-        NETFLIX_STACK,
-        NETFLIX_ENVIRONMENT
+        NETFLIX_STACK
     }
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/netflix/conductor/validations/WorkflowTaskTypeConstraint.java
```diff
@@ -142,12 +144,26 @@ private boolean isDynamicForkJoinValid(WorkflowTask workflowTask, ConstraintVali
...
+                isInputParameterSet = true;
+            }
+
+            TaskDef taskDef = Optional.ofNullable(workflowTask.getTaskDefinition()).orElse(ValidationContext.getMetadataDAO().getTaskDef(workflowTask.getName()));
+
+            if (taskDef != null && taskDef.getInputTemplate() != null  && taskDef.getInputTemplate().containsKey("http_request")) {
+                isInputTemplateSet = true;
+            }
+
+            if (!(isInputParameterSet || isInputTemplateSet)) {
```
apanicker-nflx: nit: this could just be a single boolean.

PR #936, opened 2019-01-11
Reset the reason when task retry

X-Ultra:


core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
```diff
@@ -380,7 +380,7 @@ Task retry(TaskDef taskDefinition, WorkflowTask workflowTask, Task task, Workflo
         rescheduled.setPollCount(0);
         rescheduled.setInputData(new HashMap<>());
         rescheduled.getInputData().putAll(task.getInputData());
-
+        rescheduled.setReasonForIncompletion("");
```
cyzhao: Good catch, but set it to null instead.

PR #942, opened 2019-01-15
fix retries logic for failed http tasks

alexnetflix:
workflow does not have task definition attached to http tasks so we can get it on demand when necessary rather than adding it to every workflow.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
@@ -30,6 +30,7 @@ import com.netflix.conductor.core.execution.mapper.TaskMapperContext;
 import com.netflix.conductor.core.utils.ExternalPayloadStorageUtils;
 import com.netflix.conductor.core.utils.IDGenerator;
 import com.netflix.conductor.core.utils.QueueUtils;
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
```diff
@@ -350,6 +353,11 @@ private String getNextTasksToBeScheduled(Workflow workflow, Task task) {
     Task retry(TaskDef taskDefinition, WorkflowTask workflowTask, Task task, Workflow workflow) throws TerminateWorkflowException {
 
         int retryCount = task.getRetryCount();
+
+        if(taskDefinition == null) {
+            taskDefinition = metadataDAO.getTaskDef(task.getTaskType());
```
apanicker-nflx: `metadataDAO.getTaskDef` can return a null value and this can still fail.
We can possibly throw an exception here,
```java
taskDefinition = Optional.ofNullable(taskDefinition)
                .orElseGet(() -> Optional.ofNullable(metadataDAO.getTaskDef(task.getTaskType()))
                        .orElseThrow(() -> new TerminateWorkflowException(String.format("Invalid task specified. Cannot find task by name %s in the task definitions", task.getTaskDefName())));
```

PR #950, opened 2019-01-18
(fix): disable predefine environment variables validation

falu2010-netflix:
Description:
Disable predefine environment variable validation.


common/src/main/java/com/netflix/conductor/common/utils/ConstraintParamUtil.java
```diff
@@ -62,35 +64,45 @@
...
+                    boolean isPredefinedEnum = false;
+
+                    for (SystemParameters systemParameters : SystemParameters.values()) {
+                        if (systemParameters.name().equals(paramPath)) {
+                            isPredefinedEnum = true;
+                            break;
+                        }
+                    }
+
+                    if (!isPredefinedEnum) {
```
alexnetflix: could you combine two 'if's here?

falu2010-netflix: No.

PR #954, opened 2019-01-18
Fix unencoded params when searching

mstier-nflx:


ui/src/api/wfe.js
```diff
@@ -40,11 +40,11 @@ router.get('/', async (req, res, next) => {
     const url =
       baseURL2 +
       'search?size=100&sort=startTime:DESC&freeText=' +
-      freeText.join(' AND ') +
+      encodeURIComponent(freeText.join(' AND ')) +
       '&start=' +
       start +
       '&query=' +
-      query;
+      encodeURIComponent(query);
```
apanicker-nflx: This should probably also be applied for the `search-by-task` api below.

mstier-nflx: I commited a change to search-by-task.  Place review

PR #959, opened 2019-01-23
conductor ConductorHttpClient methods now return errors

anuccio1:
Previously, the ConductorHttpClient methods were all returning `nil` instead of the error.


anuccio1:
https://github.com/Netflix/conductor/issues/977

PR #964, opened 2019-01-29
fix(py-client): proper naming for task def update function

srabenhorst:
Naming of the _registerTaskDef_ function is misleading since `PUT /metadata/taskdefs` [updates a task definition](https://netflix.github.io/conductor/runtime/#task-workflow-metadata). Proper name should be _updateTaskDef_

### DIFF
--- a/client/python/conductor/conductor.py
+++ b/client/python/conductor/conductor.py
@@ -18,6 +18,7 @@ import requests
 import json
 import sys
 import socket
+import warnings
 
 
 hostname = socket.gethostname()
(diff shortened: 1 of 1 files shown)


srabenhorst:
> Hey @srabenhorst , this is great. But, it would break backwards compatibility for users already using `registerTaskDef`. Instead, this method can be marked deprecated, and another `updateTaskDef` method can be added, with comments asking the users to use `updateTaskDef` hereon.
Is it better like this? I hope I deprecated it properly.

kishorebanala:
Looks good. I just noticed you were trying to merge this branch to master instead of dev. Can you please update this to merge to dev. More about this here: https://github.com/Netflix/conductor#contributions

srabenhorst:
> Looks good. I just noticed you were trying to merge this branch to master instead of dev. Can you please update this to merge to dev. More about this here: https://github.com/Netflix/conductor#contributions

done

Issue #871, opened 2018-11-18
[BUG]Multiple workers concurrent race Issue 


JasperB:
We have one workflow with 3 tasks, one of them is dynamically fork task, when we start 3 clients with 300 workers on each client, then ran some tests we found some issue regarding the workflow status inconsistently. 
We took some time to investigate this problem, seems it caused by the multi-thread issue, one thread index the workflow as "COMPLETED" to ES, but later(several mis-seconds) the other task thread index the workflow as "RUNNING" to ES because in this thread the workflow instance still own the status.termina as false, that's cause the workflow status incorrect.

@cyzhao @v1r3n Did you guys met this issue? Any suggestion to fix this？

![1542523462346](https://user-images.githubusercontent.com/7287153/48669488-8cf37580-eb40-11e8-92d4-19e5fbe035f3.jpg)

JasperB:
From above screenshot, you can see the first thread index the workflow as "TIMEOUT" to ES, but later the other thread index the workflow as "RUNNING" to ES, that means sometimes the workflow status will be override by the later threads.

JasperB:
@cyzhao @v1r3n did you guys met this issue, it also happened in poll tasks, means multiple workers may poll and ack same task, it's a bug which will cause the same task execute multiple times, do you guys have any idea to fix this issue? BTW we're using the aws redis-cluster and 4 clients server with multiple workers in each client.

Multiple worker race the tasks as below:
![image](https://user-images.githubusercontent.com/7287153/48890620-48cfe000-ee74-11e8-8b24-536a58434281.png)

cyzhao:
@JasperB sorry for late response, we are aware of this issue and will look into this soon. However, if you already have plan to fix this please let us know.

@apanicker-nflx @kishorebanala

kishorebanala:
@JasperB We have this on our roadmap, though not a higher priority. Please feel free to send us a PR, if this needs to be fixed sooner. Thanks for reporting the issue.

HenryLauu:
any updates? I got a task polled by workers twice. 
`"retryCount":0,
         "seq":2,
         "pollCount":2,`

kishorebanala:
@HenryLauu Can you provide more details about task being poll twice by workers please? It'd help to see the workflow execution with timestamps, posted somewhere as a snippet.

HenryLauu:
Hi @kishorebanala . Sorry for the delay . Here are some task details:
"ownerApp": "",
   "createTime": 1599471668218,
[...]

PR #965, opened 2019-01-30
fix concurrent workflow state modify problem.

ynfeng:
First modify the current workflow state and then schedule the task. Otherwise, the status of the workflow may be incorrect, which may cause the COMPLETE state to return to the RUNNING state.


ynfeng:
Looks like this is related to this problem #871 . Almost like my scene. @cyzhao

PR #966, opened 2019-01-30
Change WorkflowDef default schema version to 2

falu2010-netflix:
Description:
Change WorkflowDef default schema version to 2.

Implications:
This require changes in way you define input parameters for WorkflowTask. For schema version 1 you don't need to pass $ for input parameter expression resolution but for schema version 2 you need to pass in following format ${workflow.input...} for mapping.
```
WorkflowTask workflowTask = new WorkflowTask();
        Map<String, Object> ip1 = new HashMap<>();
        if (schemaVersion <= 1) {
            ip1.put("p1", "workflow.input.param1");
            ip1.put("p2", "workflow.input.param2");
        } else {
            ip1.put("p1", "${workflow.input.param1}");
            ip1.put("p2", "${workflow.input.param2}");
        }
workflowTask.setInputParameters(ip1);
```

### DIFF
--- a/common/src/test/java/com/netflix/conductor/common/workflow/WorkflowDefTest.java
+++ b/common/src/test/java/com/netflix/conductor/common/workflow/WorkflowDefTest.java
@@ -318,7 +318,7 @@ public class WorkflowDefTest {
 		workflowDef.setSchemaVersion(3);
 		workflowDef.setName("test_env");
 
-		WorkflowTask workflowTask = new WorkflowTask();//name is null
+		WorkflowTask workflowTask = new WorkflowTask();
 
 		workflowTask.setName("t1");
 		workflowTask.setWorkflowTaskType(TaskType.SIMPLE);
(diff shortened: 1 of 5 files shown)


common/src/test/java/com/netflix/conductor/common/workflow/WorkflowDefTest.java
```diff
@@ -128,7 +128,7 @@ public void testWorkflowDefConstraints() {
 	@Test
 	public void testWorkflowDefConstraintsWithMultipleEnvVariable() {
 		WorkflowDef workflowDef = new WorkflowDef();//name is null
```
kishorebanala: Not related to this PR, but this test could reuse WorkflowDef creation logic using @before or other constructs.

apanicker-nflx:
The build is terminated due to excessive logs from the test, please rebase against dev to get the fix and rerun the builds.

PR #974, opened 2019-02-05
ConductorWorker execute function error is now handled

anuccio1:
Previously, when an non-nil error value was returned from the `executeFunction`, the conductorworker was returning early, before sending an update to conductor that the task failed.

In this PR, I modified the `Execute` method to properly handle errors


anuccio1:
https://github.com/Netflix/conductor/issues/978

PR #975, opened 2019-02-05
Fix verification on the conductor client written in python

artpdr:
The current code on the conductor client written in Python expects a string from the server after sending an ACK request.

However, the server response is application/json and it will sen a true or false and Python will convert it to True or False. 

The function ackTask should return the value that is received from the server instead of returning the verification, which is currently implemented, and returns always False as it is comparing a boolean and a string.

Cheers.
Artur


client/python/conductor/conductor.py
```diff
@@ -219,7 +219,7 @@ def ackTask(self, taskId, workerid):
         params['workerid'] = workerid
         headers = {'Accept': 'application/json'}
         value = self.post(url, params, None, headers)
-        return value == 'true'
+        return value
```
kishorebanala: So, the Java client returns ack value as Boolean, which is explicit because of types. Whereas, the rest endpoint converts Boolean to String and returns that. In which case, the value would be amongst "true" or "false". Couple of concerns, changing value from boolean to string might break existing users. Do you have different expectations from this function? I.e do you expect it to return plain string, or json, or some other form?

artpdr:
Sorry, I think my first comment has wrong stuff. I will explain it again...

As you can see in the code above, value is assigned with the return value from self.post, which can be either resp.text() or resp.json(), being resp the response from the request. However, resp.text happens only if the headers used in the request are {"Accept": "text/plain"}. But, as you can see from the code above, headers = {"Accept": "application/json"}, thus value will get the result of resp.json(), which will be either True or False (Python booleans) not true or false (js) or "true" / "false".

As the ackTask method was written before it would evaluate always to False because both True == "true" and False == "true" are False...

You might want to check the methods post and __return from the same file to better understand what I'm talking about...

PR #979, opened 2019-02-05
Task output of last executed task

jkaipa:
Exclusive join takes output from one of the previous successful task based on the configuration. 
If the scenario is T1 -> [T2/T3] -> EJ -> T4 then, based on the decision T1 is followed by T2 or T3. If the decision cases are setup to execute T2 on true and T3 on false then, here is the outcome.
true  => T1 -> T2 -> EJ -> T4. EJ will produce the output of T2
false => T1 -> T3 -> EJ -> T4. EJ will produce the output of T3
null . => T1 -> EJ -> T4. EJ will produce the output of T1. Here the output of EJ is the result of the defaultExclusiveJoinTask configuration.
Ex:
```
{
        "name": "exclusive_join",
        "taskReferenceName": "exclusiveJoin",
        "type": "EXCLUSIVE_JOIN",
        "joinOn": [
          "task2",
          "task3"
        ],
        "defaultExclusiveJoinTask": [
          "task1"
        ]
}
```

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/tasks/ExclusiveJoin.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/tasks/ExclusiveJoin.java
@@ -1,80 +1,90 @@
 package com.netflix.conductor.core.execution.tasks;
 
-import com.netflix.conductor.common.metadata.tasks.Task;
-import com.netflix.conductor.common.run.Workflow;
-import com.netflix.conductor.core.execution.WorkflowExecutor;
+import java.util.List;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.List;
+import com.netflix.conductor.common.metadata.tasks.Task;
+import com.netflix.conductor.common.run.Workflow;
+import com.netflix.conductor.core.execution.WorkflowExecutor;
 
 public class ExclusiveJoin extends WorkflowSystemTask {
-    private static Logger logger = LoggerFactory.getLogger(ExclusiveJoin.class);
+	private static Logger logger = LoggerFactory.getLogger(ExclusiveJoin.class);
+	private static final String NAME = "EXCLUSIVE_JOIN";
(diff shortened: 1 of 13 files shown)


core/src/main/java/com/netflix/conductor/core/execution/tasks/ExclusiveJoin.java
```diff
@@ -0,0 +1,109 @@
...
+		}
+
+		logger.debug("Status of flags: foundExlusiveJoinOnTask: {}, hasFailures {}, useWorkflowInput {}",
+				foundExlusiveJoinOnTask, hasFailures, useWorkflowInput);
+		if (foundExlusiveJoinOnTask || hasFailures || useWorkflowInput) {
+			if (hasFailures) {
+				task.setReasonForIncompletion(failureReason.toString());
+				task.setStatus(Task.Status.FAILED);
+			} else {
+				task.setOutputData((useWorkflowInput) ? workflow.getInput() : exclusiveTask.getOutputData());
```
apanicker-nflx: Can you explain why workflow input is wired as the output of this task if an exclusive task is not defined?

jkaipa: By contract exclusive join has to provide the previous tasks output. Proper value is expected as exclusive task or the default task. If neither is matched, the workflows input is passed as is as output to drive the next task in the sequence.  Do you think it is a better idea to have a configuration flag that determines if this task has to consider workflow input as the output?

apanicker-nflx: In this case, the output of the task should be `null` or `empty` to explicitly indicate that no cases matched and the default fallback was used. The workflow input is accessible from within any task as-is and does not need this additional indirection.

jkaipa: Incorporated review comments.

PR #983, opened 2019-02-08
Fixed NullpointerException in couple of places

falu2010-netflix:
* Added precondition check to determine exact place for NPE
* Fixed NPE in WebApplicationExceptionMapper


core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -978,6 +978,10 @@ public void addTaskToQueue(Task task) {
     public void executeSystemTask(WorkflowSystemTask systemTask, String taskId, int unackTimeout) {
         try {
             Task task = executionDAOFacade.getTaskById(taskId);
+            if (task == null){
```
kishorebanala: Wondering why executionDAO would return null instead of throwing error in the first place.

PR #988, opened 2019-02-12
ParametersUtils some code need catch exception

huangyiminghappy:
documentContext.read may occur exception,so i think we need to catch the exception.
i will close the pr :https://github.com/Netflix/conductor/pull/617,use this replace it


kishorebanala:
@huangyiminghappy Thanks for resubmitting the PR. Can you fix the code coverage errors please.

huangyiminghappy:
> @huangyiminghappy Thanks for resubmitting the PR. Can you fix the code coverage errors please.

thank you review. @kishorebanala i can not found the  parametersUtils test case, do not know add the case in which test class,could i add a new test class to cover the uncoverd code?can you help me where i can add the test method?

kishorebanala:
@huangyiminghappy https://github.com/Netflix/conductor/blob/master/core/src/test/java/com/netflix/conductor/core/execution/TestParametersUtils.java is the Test file for ParameterUtils.

Also, I see there are several commits for this PR. Could you also rebase it and cleanup the commit history, so that the history would only have one commit.

huangyiminghappy:
> @huangyiminghappy https://github.com/Netflix/conductor/blob/master/core/src/test/java/com/netflix/conductor/core/execution/TestParametersUtils.java is the Test file for ParameterUtils.
> 
> Also, I see there are several commits for this PR. Could you also rebase it and cleanup the commit history, so that the history would only have one commit.

ok,thank you

apanicker-nflx:
@huangyiminghappy Please rebase against dev and squash changes into 1 commit.

huangyiminghappy:
> @huangyiminghappy Please rebase against dev and squash changes into 1 commit.

thank you reply,i have rebase the commit

core/src/main/java/com/netflix/conductor/core/execution/ParametersUtils.java
```diff
@@ -205,7 +207,12 @@ private Object replaceVariables(String paramString, DocumentContext documentCont
                     }
 
                 } else {
-                    convertedValues[i] = documentContext.read(paramPath);
+                    try {
+                        convertedValues[i] = documentContext.read(paramPath);
+                    }catch (Exception e) {
+                        logger.warn("documentContext read paramPath: {} error", paramPath);
```
apanicker-nflx: Please also print out the exception stack trace in the log statement.
logger.warn("Error reading documentContext for paramPath: {}. Exception: {}", paramPath, e);

huangyiminghappy: ok

PR #1025, opened 2019-03-05
Adds HTTP TaskTimeouts

naveenchlsn:
1) Adds ReadTimeout and connection timeout to HTTP Tasks. 
     We take the parameters from HTTP input. If they are not present,  then timeouts from config are used. 
2) I have removed the input parameter in the RestManager.getClient method as it is an unused parameter. 
3)  RestManager.getClient uses threadlocal to create clients as we are modying the client object from various threads currently.

### DIFF
--- a/contribs/src/main/java/com/netflix/conductor/contribs/http/RestClientManager.java
+++ b/contribs/src/main/java/com/netflix/conductor/contribs/http/RestClientManager.java
@@ -47,7 +47,7 @@ public class RestClientManager {
 		this.defaultConnectTimeout = config.getIntProperty(HTTP_TASK_CONNECT_TIMEOUT, DEFAULT_CONNECT_TIMEOUT);
 	}
 
-	public Client getClient() {
+	public Client getClient(Input input) {
 		Client client = threadLocalClient.get();
 		client.setReadTimeout(defaultReadTimeout);
 		client.setConnectTimeout(defaultConnectTimeout);
--- a/contribs/src/main/java/com/netflix/conductor/contribs/http/RestClientManager.java
+++ b/contribs/src/main/java/com/netflix/conductor/contribs/http/RestClientManager.java
@@ -18,6 +18,7 @@
  */
 package com.netflix.conductor.contribs.http;
 
+import com.netflix.conductor.core.config.Configuration;
 import javax.inject.Singleton;
 
 import com.netflix.conductor.contribs.http.HttpTask.Input;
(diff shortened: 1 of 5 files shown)


contribs/src/main/java/com/netflix/conductor/contribs/http/RestClientManager.java
```diff
@@ -30,9 +31,27 @@
...
+	private final int defaultReadTimeout;
+	private final int defaultConnectTimeout;
+
+	public RestClientManager(Configuration config) {
+		this.threadLocalClient = ThreadLocal.withInitial(()->Client.create());
+		this.defaultReadTimeout =config.getIntProperty(HTTP_TASK_READ_TIMEOUT, DEFAULT_READ_TIMEOUT);
+		this.defaultConnectTimeout = config.getIntProperty(HTTP_TASK_CONNECT_TIMEOUT, DEFAULT_CONNECT_TIMEOUT);
 	}
+
+	public Client getClient() {
```
kishorebanala: The Input parameter is actually useful in switching between the clients provided based on input. For eg: Returning client with vipAddress vs client with URL. We'll change this to interface to make it clear. But, please add it back for now to maintain backwards compatibility with existing implementations.

contribs/src/main/java/com/netflix/conductor/contribs/http/RestClientManager.java
```diff
@@ -30,9 +31,27 @@
...
+	static final String HTTP_TASK_READ_TIMEOUT = "http.task.read.timeout";
+	static final String HTTP_TASK_CONNECT_TIMEOUT = "http.task.connect.timeout";
+
+
+	private final ThreadLocal<Client> threadLocalClient;
+	private final int defaultReadTimeout;
+	private final int defaultConnectTimeout;
+
+	public RestClientManager(Configuration config) {
+		this.threadLocalClient = ThreadLocal.withInitial(()->Client.create());
```
kishorebanala: Given this class is Singleton, not sure how ThreadLocal is helpful here.

naveenchlsn: We are mutating the Client object in the HttpTask.java, httpCall method. As this method can be called from multiple threads, I have tried to make the Client object threadlocal. Please let me know if I am missing something.

contribs/src/main/java/com/netflix/conductor/contribs/http/RestClientManager.java
naveenchlsn: I have added the input parameter back.

contribs/src/main/java/com/netflix/conductor/contribs/http/RestClientManager.java
naveenchlsn: @kishorebanala  Can you please let me know your thoughts.

PR #1050, opened 2019-03-15
Fix edges for joinOn references

demichej:
This is a fix for a UI bug where the edges from a vertex would point to the incorrect "join". This changes it so that we build up a mapping of joinOns for a given task, and then we create edges that point to those joinOns.

It looks like maybe someone started to do something like this at some point, but it wasn't totally completed.

We have a task with a referencename of  "review_liabilities" over in a Fork branch, and we have it defined to joinOn "join5". But the original code would point to the wrong JOIN:

badPointing

Instead it should be pointing to join5. This PR will change that so it correctly illustrates what is actually happening:

goodPointing


demichej:
@mstier-nflx and @hunterford I have resolved the conflicts. Could you take a look? Thanks.

ui/src/api/wfegraph.js
```diff
@@ -17,28 +18,45 @@ class Workflow2Graph {
...
         let forkedTasks = tt.inputData && tt.inputData.forkedTasks || [];
         forkedTasks.forEach(ft =>{
-          wfts.push({name: ft, referenceTaskName: ft, type: 'SIMPLE'});
+          wfts.push({name: ft, taskReferenceName: ft, type: 'SIMPLE'});
         });
-        forks[tt.referenceTaskName] = wfts;
+        forks[tt.taskReferenceName] = wfts;
       }
-    });
+    }), this;
```
mstier-nflx: I think the parentheses is misplaced.  Should it be moved to after the 'this'?

demichej: Good catch @mstier-nflx 

What's interesting is that I just triple-checked, and this actually still builds and compiles and works. I am changing the parenthesis as you suggested......just not sure why. Maybe there's something we depend on or webpack itself that compiles this correctly? Really weird that it works this way :shrug:

PR #1058, opened 2019-03-20
Feature/event wait

jkaipa:


kishorebanala:
I'm closing this PR. But please feel free to reopen the conversation in gitter if you think we're missing anything.

jkaipa:
I tried other ways, but this seem to be the ideal way the solution works. How about calling it as a COMMAND task if you don't want it to be called it as EVENT_WAIT?

kishorebanala:
We've added asyncComplete flag to Event & Http tasks, which sends an Event and waits in progress: https://netflix.github.io/conductor/configuration/systask/#event. Try and see if that solves your use case.

Issue #989, opened 2019-02-12
User defined workflow ID


mohelsaka:
Hello,

Is there a way to ensure that workflow is running at most once for some particular payload?
So that if `POST /workflow/{name}` was called twice at the same time with the same workflow input, only one call will be accepted and the others will be rejected (or do nothing).

I thought of three ways to solve this problem:
1. From the client side. Store the workflow ID in some storage and check this storage before calling conductor server to start the workflow.
2. In the workflow itself, add a task that queries conductor search API with the payload and checks if there are any running WFs before or not, and if there are any, fail the WF (It can be done using HTTP System Task, maybe).
3. Having user-defined unique workflow id, something similar to what SWF is doing with workflowid, details are [here](https://docs.aws.amazon.com/amazonswf/latest/apireference/API_StartWorkflowExecution.html#SWF-StartWorkflowExecution-request-workflowId).

I believe the 3rd solution is the cleanest, as it's less overhead on the client (unlike the 1st approach) and less irrelevant tasks in the workflow (unlike the 2nd approach).

If you have similar use problem, can you please share how you solve it?

Also if you think that the 3rd approach makes sense then I'm up to pushing a PR for it.

Thanks!

apanicker-nflx:
@mohelsaka Currently, conductor does not provide a construct to ensure that only one workflow is currently running for a particular payload. However, as you pointed out, this can be achieved by one of the ways that you have suggested.

1. This involves the client maintaining a storage for workflow ids and ensuring sync with the conductor server.
2. This is not a reliable model, since the search API uses ES which can have a latency depending on the indexing of the data in the cluster, resulting in stale results.
3. I agree that this is the cleanest solution.

Please feel free to submit a PR for the same and we will be happy to review and merge it.
Please ensure that the user-defined workflow id is a UUID and that there is no contention with the workflow ids that the conductor server mints.

PR #1064, opened 2019-03-21
Idempotent workflows

mohelsaka:
This merge request introduces idempotent workflows in Conductor, as per what have been discussed with @apanicker-nflx in this [issue](https://github.com/Netflix/conductor/issues/989).

Idempotency Keys of workflows can be defined in `startWorfklow` API call. The new parameter is optional, so this change should not impact users how are not using this new feature.

Idempotency feature works now with only Redis backend, there is no implementation yet for Cassandra or MySql, but can be easily added by implementing three functions `ExecutionDAO#getWorkflowIdByIdempotencyKey`, `ExecutionDAO#obtainLockForWorkflow` and `ExecutionDAO#releaseLockForWorkflow`.

The merge request is missing tests for the new feature, I'm willing to do them just if I get an 'Ack' on the merge request.

Workflow Bulk APIs have not been tested yet. Also, "Start a new workflow with StartWorkflowRequest" was not supported yet.

Here's a sample of API respose when there is another WF running with the same idempotency key:
```
curl -X POST  -H 'Content-Type: application/json' -d '{}' http://localhost:8080/api/workflow/test_small_wf\?correlationId\=some_id\&idempotencyKey\=my_unique_key
{"code":"CONFLICT","message":"A workflow is already running with the same idempotencyKey","retryable":false,"instance":"C02SK0JQFVH6"}
```

Thanks!

### DIFF
--- a/common/src/main/java/com/netflix/conductor/common/run/Workflow.java
+++ b/common/src/main/java/com/netflix/conductor/common/run/Workflow.java
@@ -115,6 +115,9 @@ public class Workflow extends Auditable{
     @ProtoField(id = 21)
 	private String externalOutputPayloadStoragePath;
 
+    @ProtoField(id = 22)
+    private String userDefinedId;
+
 	public Workflow(){
 
 	}
@@ -412,6 +415,14 @@ public class Workflow extends Auditable{
 	 */
 	public void setExternalOutputPayloadStoragePath(String externalOutputPayloadStoragePath) {
 		this.externalOutputPayloadStoragePath = externalOutputPayloadStoragePath;
+	}
+
+	public String getUserDefinedId() {
+		return this.userDefinedId;
+	}
+
+	public void setUserDefinedId(String userDefinedId) {
+		this.userDefinedId = userDefinedId;
 	}
 
 	public Task getTaskByRefName(String refName) {
--- a/common/src/main/java/com/netflix/conductor/common/run/Workflow.java
+++ b/common/src/main/java/com/netflix/conductor/common/run/Workflow.java
@@ -116,7 +116,7 @@ public class Workflow extends Auditable{
 	private String externalOutputPayloadStoragePath;
 
     @ProtoField(id = 22)
-    private String userDefinedId;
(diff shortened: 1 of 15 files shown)


apanicker-nflx:
@mohelsaka Firstly, thanks for the contribution, this is a great start. Couple of comments about the pull request:
- Please remove all the changes for the `StartWorkflowParameters`, this is a refactoring change and does not belong here, can be submitted in a separate PR.
- Please add `idempotencyKey` to the StartWorkflowRequest object so that it is available for use within the Java client.
- The error message during conflict scenarios `"A workflow is already running with the same idempotencyKey"` could be enhanced to contain the workflowId that is running for the specified idempotencyKey so that the user has better visibility.

Please make above changes alongwith adding tests and some description which can be added to the documentation.

mohelsaka:
@apanicker-nflx Thanks for the feedback! ... I'll make the changes soon.

For the first point of removing the refactoring for`StartWorkflowParameters`. I do agree it doesn't belong here, but honestly, I wasn't able to proceed with the change without doing this refactoring. There were about 10 functions named `startWorkflow` that rely on parameters overloading, each has >5 parameters. It won't be really easy to undo this refactoring now. Does this sound like a good reason for keeping this refactoring in the same merger request?

apanicker-nflx:
Agreed with the reasoning, but for sake of simplicity and conciseness, can you break the refactoring into a separate PR and we can merge that first before this one. I apologize for the extra work that this involves but it would make it easier to debug and isolate any issues in the long run.

apanicker-nflx:
@mohelsaka Please let us know your thoughts on how to take this PR forward and if you're actively looking into it.

mohelsaka:
@apanicker-nflx 
Sorry for the late update, I'll split it into two merge request as requested before. I expect that I'll finish it before the end of May.

apanicker-nflx:
@mohelsaka Can you please provide an update on this PR? Thanks

apanicker-nflx:
Closing PR due to inactivity, please reopen whenever you are able to resume the work.

Issue #1061, opened 2019-03-20
Wrong times for decision and fork tasks


emiteze:
During some tests we noticed that scheduled, start and end times for decision and fork tasks were a little bit confusing. We ran the above workflow:

![Screenshot from 2019-03-20 17-07-57](https://user-images.githubusercontent.com/13343278/54715831-1fc61000-4b33-11e9-9940-1de1ea5c965a.png)

Here are the resumed JSON from the execution that I took from conductor-ui:

```json
{  
   "createTime":1553108639406,
   "updateTime":1553108644261,
   "status":"COMPLETED",
   "endTime":1553108644261,
   "workflowId":"8684d51a-73b5-4b5f-bdc7-b9eba85441d3",
   "tasks":[  
      {  
         "taskType":"DECISION",
         "status":"COMPLETED",
         "referenceTaskName":"Conditional01",
         "scheduledTime":1553108639892,
         "startTime":1553108640457,
         "endTime":1553108639655,
         "updateTime":1553108644319
      },
      {  
         "taskType":"FORK",
         "status":"COMPLETED",
         "referenceTaskName":"Fork01",
         "scheduledTime":1553108640052,
         "startTime":1553108640535,
         "endTime":1553108639575,
         "updateTime":1553108644399
      },
      {  
         "taskType":"HTTP",
         "status":"COMPLETED",
         "referenceTaskName":"Request01",
         "scheduledTime":1553108640209,
         "startTime":1553108641042,
         "endTime":1553108643079,
         "updateTime":1553108644479
      },
      {  
         "taskType":"HTTP",
         "status":"COMPLETED",
         "referenceTaskName":"Request02",
         "scheduledTime":1553108640338,
         "startTime":1553108641201,
         "endTime":1553108643075,
         "updateTime":1553108644557
      },
      {  
         "taskType":"JOIN",
         "status":"COMPLETED",
         "referenceTaskName":"Join01",
         "scheduledTime":1553108641579,
         "startTime":1553108641683,
         "endTime":1553108643847,
         "updateTime":1553108644635
      }
   ]
}
```

In this case, DECISION task has scheduledTime greater than startTime and endTime less than both. For the FORK task, just the endTime is less than scheduled and start times.

After some debugging we noticed that in some cases scheduledTime was being set more than one time, and endTime was not being updated after being set. Have you guys noticed this as well or this is the first report for this problem? Maybe this is a good contribution that we can do.

apanicker-nflx:
[...]

PR #1073, opened 2019-03-27
Fixing conductor execution times

emiteze:
This PR solves the issue related in #1061 

I will be glad to hear what you guys think about this solution!


core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -1155,7 +1155,9 @@ boolean scheduleTask(Workflow workflow, List<Task> tasks) {
                 if (workflowSystemTask == null) {
                     throw new ApplicationException(NOT_FOUND, "No system task found by name " + task.getTaskType());
                 }
-                task.setStartTime(System.currentTimeMillis());
+                if (task.getStatus() != null && task.getStatus().isTerminal() && task.getStartTime() == 0) {
```
apanicker-nflx: This should be `!task.getStatus().isTerminal()`

test-harness/src/test/java/com/netflix/conductor/tests/integration/WorkflowLegacyMigrationTest.java
```diff
@@ -118,4 +118,10 @@ private String getWorkflowResourcePath(String workflowName) {
      */
     public void testForkJoinNestedWithSubWorkflow() {
     }
+
+    @Ignore
+    @Test
+    @Override
+    public void testExecutionTimes(){}
```
apanicker-nflx: Why is this test ignored?

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
emiteze: Fixed

test-harness/src/test/java/com/netflix/conductor/tests/integration/WorkflowLegacyMigrationTest.java
emiteze: I tought that these tests were not supposed to run based on new code. These legacy tests are a bit confusing and to be honest I still don't get how it works properly.

Can you explain why these legacy tests exists and how to test it properly? Even making a test that passes, I don't know if I mounted the .json correctly.

apanicker-nflx: The legacy tests were added when moving to Conductor 2.0 to ensure backwards compatibility such that workflows in progress are not affected by the rolling update. As long as the test is added to the `AbstractWorkflowServiceTest` class, it will be run as both set of tests (legacy and current).

PR #1096, opened 2019-04-05
fix for retry when using external payload storage

apanicker-nflx:


client/src/main/java/com/netflix/conductor/client/http/TaskClient.java
```diff
@@ -247,6 +248,9 @@ public void updateTask(TaskResult taskResult, String taskType) {
         } catch (IOException e) {
             String errorMsg = String.format("Unable to update task: %s with task result", taskResult.getTaskId());
             logger.error(errorMsg, e);
+            // reset output data and external output storage payload path
+            taskResult.setOutputData(outputData);
+            taskResult.setExternalOutputPayloadStoragePath(null);
             throw new ConductorClientException(errorMsg, e);
         }
         postForEntityWithRequestOnly("tasks", taskResult);
```
kishorebanala: There is still an edge case when `uploadToExternalPayloadStorage` succeeds, but `postForEntityWithRequestOnly` fails. The retry kicks off again and fails validation. I'll try to separate upload to s3 and post task result in chained retires.

PR #1098, opened 2019-04-07
Updated WorkflowTaskCoordinator to separate retry logic for large pay…

kishorebanala:
…loads and task update, there by avoiding precondition failures on retry.

### DIFF
--- a/client/src/main/java/com/netflix/conductor/client/http/TaskClient.java
+++ b/client/src/main/java/com/netflix/conductor/client/http/TaskClient.java
@@ -218,9 +218,13 @@ public class TaskClient extends ClientBase {
      * it is uploaded to {@link ExternalPayloadStorage}, if enabled, else the task is marked as FAILED_WITH_TERMINAL_ERROR.
      *
      * @param taskResult the {@link TaskResult} of the executed task to be updated.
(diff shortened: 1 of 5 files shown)


client/src/main/java/com/netflix/conductor/client/http/TaskClient.java
```diff
@@ -218,9 +218,13 @@ public Task getPendingTaskForWorkflow(String workflowId, String taskReferenceNam
      * it is uploaded to {@link ExternalPayloadStorage}, if enabled, else the task is marked as FAILED_WITH_TERMINAL_ERROR.
      *
      * @param taskResult the {@link TaskResult} of the executed task to be updated.
-     * @param taskType   the type of the task
      */
-    public void updateTask(TaskResult taskResult, String taskType) {
+    public void updateTask(TaskResult taskResult) {
```
pctreddy: The contract of this public method has changed from: 
`public void updateTask(TaskResult taskResult, String taskType)` to `public void updateTask(TaskResult taskResult)`, does this break backward compatibility for the existing usage ? 

If that is the case, I would suggest leave the existing method and mark it as deprecated and create a new overloaded method

PR #1133, opened 2019-05-08
Limit configuration for worker queue size of Async Index Dao

naveenchlsn:
Issue: 

The executor service used in async functions of Indexdao's has workerQueue Size as unbounded. 
new LinkedBlockingQueue<>() -> This creates a blocking queue size of Integer.MAX_VALUE. 

During high throughput, tasks are queued in the Linkedblocking queue due to which memory is exhausted. This is causing GC issues. 

Following are the changes. 

1. I am limiting the size to 100 by default
2. Read from configuration if this needs to be changed. 

@cyzhao @apanicker-nflx @mashurex @kishorebanala

### DIFF
--- a/es5-persistence/README.md
+++ b/es5-persistence/README.md
@@ -17,6 +17,8 @@ If using the `http` or `https`, then conductor will use the REST transport proto
 Defaults to `conductor`
 * `workflow.elasticsearch.tasklog.index.name` - The name of the task log index.
 Defaults to `task_log`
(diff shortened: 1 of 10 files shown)


kishorebanala:
@naveenchlsn This looks good. Do you mind improving module documentation on how to tune this parameter. Also, to maintain consistency with ES6 module, can you also add this to `es6-persistence` please.

Issue #1054, opened 2019-03-16
conductor:ui npm install failed


gregoirer:
Step 1/11 : FROM node:9-alpine
 ---> a56170f59699
Step 2/11 : MAINTAINER Netflix OSS <conductor@netflix.com>
 ---> Using cache
 ---> 475d5ce19b9b
Step 3/11 : RUN apk update && apk add   autoconf   automake   libtool   build-base   libstdc++   gcc   abuild   binutils   nasm   libpng   libpng-dev   libjpeg-turbo   libjpeg-turbo-dev   python
 ---> Using cache
 ---> b776d67dd815
Step 4/11 : RUN mkdir -p /app/ui
 ---> Using cache
 ---> 659fcd1789c9
Step 5/11 : COPY ./docker/ui/bin /app
 ---> e010e1893c7d
Step 6/11 : COPY ./ui /app/ui
 ---> a55889874880
Step 7/11 : RUN chmod +x /app/startup.sh
 ---> Running in 7f182567f2c9
Removing intermediate container 7f182567f2c9
 ---> 4720ed77b58b
Step 8/11 : RUN cd /app/ui   && npm install   && npm run build --server
 ---> Running in 37a633261d35
npm WARN notice [SECURITY] minimatch has the following vulnerability: 1 high. Go here for more details: https://nodesecurity.io/advisories?search=minimatch&version=0.3.0 - Run `npm i npm@latest -g` to upgrade your npm version, and then `npm audit` to get more info.
npm WARN notice [SECURITY] minimatch has the following vulnerability: 1 high. Go here for more details: https://nodesecurity.io/advisories?search=minimatch&version=0.2.14 - Run `npm i npm@latest -g` to upgrade your npm version, and then `npm audit` to get more info.
npm WARN notice [SECURITY] debug has the following vulnerability: 1 low. Go here for more details: https://nodesecurity.io/advisories?search=debug&version=2.2.0 - Run `npm i npm@latest -g` to upgrade your npm version, and then `npm audit` to get more info.
npm WARN notice [SECURITY] node.extend has the following vulnerability: 1 moderate. Go here for more details: https://nodesecurity.io/advisories?search=node.extend&version=1.1.6 - Run `npm i npm@latest -g` to upgrade your npm version, and then `npm audit` to get more info.
[...]

PR #1134, opened 2019-05-09
Fix the UI docker-compose build step

### LINKED ISSUE #1123: conductor-ui is failing to build with docker-compose build on fresh checkout
I am not sure why, but this ticket was closed out, but never actually resolved:

#1054 

One of the dependencies had its resolve url updated to a new version......but nothing else was updated to use that new version. So it fails with an EINTEGRITY error.

@apanicker-nflx , is there a bug tracking this? It's currently broken.

demichej:
This fixes #1123  and #1054

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -3266,7 +3266,7 @@
       "dev": true,
       "requires": {
         "debug": "2.6.9",
-        "lazy-debug-legacy": "0.0.1",
+        "lazy-debug-legacy": "0.0.3",
         "object-assign": "4.1.0"
       },
       "dependencies": {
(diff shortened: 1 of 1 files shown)


demichej:
> I do not see a version 0.0.3 in NPM. Where are you getting this version?

@mstier-nflx so I think what's going on here is that if you go here directly to the url, you get the download you want:
https://registry.npmjs.org/lazy-debug/-/lazy-debug-0.0.3.tgz

But really, I think that's the code from `lazy-debug`, not `lazy-debug-legacy`. 

So this fixes the build for right now, so I think we should get this merged in so that people are not blocked from building/using Conductor. 

It seems like some larger investigation should go into why we are even depending on this package really, or at least investigate upgrading `debug-fabulous`, which is what depends on this.

Issue #1115, opened 2019-04-22
Bug in workflow metadata validation


artpdr:
I updated conductor to version 2.10.0 and after that I am seeing a bug that was not present on a previous version (2.1.1).
The bug happens during workflow definition, when a worker task has a vector as an input parameter and this vector has an entry with an empty string (e.g. inputParameters: {distinct_paths_used_per_task: ["", "foo", "bar"]}).

Here is the response from conductor server:
ERROR: {"status":400,"message":"Validation failed, check below errors for detail.","retryable":false,"validationErrors":[{"path":"updateWorkflowDef.arg0[0]","message":"key: distinct_paths_used_per_task input parameter value: is null or empty"}]}

apanicker-nflx:
@artpdr Thanks for reporting the bug. Please feel free to contribute a fix for this issue.

artpdr:
@apanicker-nflx I may contribute to this if there isn't anyone working on this issue.

apanicker-nflx:
There isn't anyone actively working on this atm, so feel free to send a PR with the fix.

artpdr:
PR sent. I will be waiting for news from your side...

PR #1140, opened 2019-05-15
Remove validation of empty strings.

artpdr:
Hi everyone. This PR is motivated by the conversation that we had in issue #1115. 

It seems that there is a validation step in the code that doesn't allow empty strings to be input parameters. However, I don't see why we should not have empty strings as input parameters. It was possible before, and it resulted in errors for me in an updated version of conductor.

I ran the unit tests but didn't define and execute a workflow with empty strings as input parameters after my changes (yet...)

### DIFF
--- a/common/src/main/java/com/netflix/conductor/common/utils/ConstraintParamUtil.java
+++ b/common/src/main/java/com/netflix/conductor/common/utils/ConstraintParamUtil.java
@@ -56,12 +56,6 @@ public class ConstraintParamUtil {
     private static List<String> extractParamPathComponentsFromString(String key, String value, String taskName, WorkflowDef workflow) {
         ArrayList<String> errorList = new ArrayList<>();
 
-        if (StringUtils.isEmpty(value)) {
-            String message = String.format( "key: %s input parameter value: is null or empty", key);
-            errorList.add(message);
-            return errorList;
-        }
-
         String[] values = value.split( "(?=\\$\\{)|(?<=\\})" );
 
         for (int i = 0; i < values.length; i++)
--- a/common/src/main/java/com/netflix/conductor/common/utils/ConstraintParamUtil.java
+++ b/common/src/main/java/com/netflix/conductor/common/utils/ConstraintParamUtil.java
@@ -56,6 +56,12 @@ public class ConstraintParamUtil {
     private static List<String> extractParamPathComponentsFromString(String key, String value, String taskName, WorkflowDef workflow) {
         ArrayList<String> errorList = new ArrayList<>();
 
+        if (value == null) {
+            String message = String.format("key: %s input parameter value: is null", key);
+            errorList.add(message);
+            return errorList;
+        }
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/netflix/conductor/common/utils/ConstraintParamUtil.java
```diff
@@ -56,12 +56,6 @@
     private static List<String> extractParamPathComponentsFromString(String key, String value, String taskName, WorkflowDef workflow) {
         ArrayList<String> errorList = new ArrayList<>();
 
-        if (StringUtils.isEmpty(value)) {
```
apanicker-nflx: `StringUtils.isEmpty` checks for both null and empty values. Removing this check altogether removes null check validation as well. Please update so that only empty strings are permitted but null values are validated against.

common/src/test/java/com/netflix/conductor/common/utils/ConstraintParamUtilTest.java
```diff
@@ -175,7 +175,29 @@ public void testExtractParamPathComponentsWithInputParamEmpty() {
         workflowDef.setTasks(tasks);
 
         List<String> results = ConstraintParamUtil.validateInputParam(inputParam,"task_1", workflowDef);
-        assertEquals(results.size(), 1);
+        assertEquals(results.size(), 0);
+    }
+
+    @Test
+    public void testExtractParamPathComponentsWithListInputParamWithEmptyString() {
```
apanicker-nflx: 👍

common/src/test/java/com/netflix/conductor/common/workflow/WorkflowDefTest.java
```diff
@@ -304,12 +304,7 @@ public void testWorkflowTaskInputParamValueInvalid() {
...
 		Validator validator = factory.getValidator();
 		Set<ConstraintViolation<Object>> result = validator.validate(workflowDef);
-		assertEquals(1, result.size());
-
-		List<String> validationErrors = new ArrayList<>();
-		result.forEach(e -> validationErrors.add(e.getMessage()));
-
-		assertTrue(validationErrors.contains("key: blabla input parameter value: is null or empty"));
+		assertEquals(0, result.size());
 	}
```
apanicker-nflx: Consider adding a test case for `listInputParamWithEmptyString` to this class as well.

common/src/main/java/com/netflix/conductor/common/utils/ConstraintParamUtil.java
artpdr: done :)

common/src/test/java/com/netflix/conductor/common/workflow/WorkflowDefTest.java
artpdr: done

PR #1156, opened 2019-05-22
Issue 1148 : Reset start time when rerunning a workflow from that task

renuka-raju:
Reset task start time when rerunning a workflow from a specific task to prevent workflow from timing out using the old value of start time.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
@@ -1300,11 +1300,16 @@ public class WorkflowExecutor {
                     executionDAOFacade.removeTask(task.getTaskId());
                 }
             }
-            //reset start time of rerunTask
-            rerunFromTask.setStartTime(System.currentTimeMillis());
+            //reset fields before restarting the task
+            rerunFromTask.setScheduledTime(System.currentTimeMillis());
+            rerunFromTask.setStartTime(0);
+            rerunFromTask.setUpdateTime(0);
+            rerunFromTask.setEndTime(0);
+            rerunFromTask.setOutputData(null);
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -1300,6 +1300,8 @@ private boolean rerunWF(String workflowId, String taskId, Map<String, Object> ta
                     executionDAOFacade.removeTask(task.getTaskId());
                 }
             }
+            //reset start time of rerunTask
+            rerunFromTask.setStartTime(System.currentTimeMillis());
```
apanicker-nflx: Here, the task is only set to `SCHEDULED` state and not started hence startTime must be 0. It is only set to `IN_PROGRESS` and started if it is a SUB_WORKFLOW.
This should be added at line 1307.

Since rerunning a task basically means that the task is being restarted, other fields within the task also need to be reset.
```java
rerunFromTask.setScheduledTime(System.currentTimeMillis());
rerunFromTask.setStartTime(0);
rerunFromTask.setUpdateTime(0);
rerunFromTask.setEndTime(0);
rerunFromTask.setOutputData(null);
```

renuka-raju: Just reiterating, we need to reset all the fields to default but reset starttime to currenttime only if this task is a SUB_WORKFLOW.

apanicker-nflx: Also, need to reset `rerunFromTask.setExternalOutputPayloadStoragePath(null);`

apanicker-nflx:
Thanks for the fix and the quick turnaround on the comments.

PR #1158, opened 2019-05-23
defines locking API with a no-op implementation.

milongapau:
Locking API with a no-op implementation.  Data persistence changes will be added to each database individually, the persistence interface ExecutionDAO.updateworkflow(workflow) remains the same.

@kishorebanala @apanicker-nflx @naveenchlsn please take a look.


core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -1319,4 +1350,34 @@ private boolean rerunWF(String workflowId, String taskId, Map<String, Object> ta
...
+     * @param workflowId
+     */
+    private void waitForLock(String workflowId) {
+        if (config.enableWorkflowExecutionLock()) {
+            lockService.acquireLock(workflowId);
+            LOGGER.info("Thread {} acquired lock to workflowId {}.", Thread.currentThread().getId(), workflowId);
+        }
+    }
+
+    private void releaseLock(String workflowid) {
```
mudit3774: Consider sticking to camelCase for workflowId in the param

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -885,31 +901,41 @@ public boolean decide(String workflowId) {
...
      * @param workflowId
      * @throws IllegalStateException
      */
     public void resumeWorkflow(String workflowId) {
-        Workflow workflow = executionDAOFacade.getWorkflowById(workflowId, false);
-        if (!workflow.getStatus().equals(WorkflowStatus.PAUSED)) {
-            throw new IllegalStateException("The workflow " + workflowId + " is not PAUSED so cannot resume. " +
-                    "Current status is " + workflow.getStatus().name());
+        try {
+            waitForLock(workflowId);
```
mudit3774: Consider having a function to construct the lockId which takes the workflowId and the context into account. In the future, we might want to lock workflows for other critical sections also.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -793,6 +804,9 @@ public Task getTask(String taskId) {
      * @throws ApplicationException If there was an error - caller should retry in this case.
      */
     public boolean decide(String workflowId) {
```
naveenchlsn: @milongapau  Can you please tell me what is the code path taken when we see the race condition. 
From what I understand, A parallel execution of the same task can happen if the task is present multiple times in the task queue. How can this happen? 



I will be able to review the PR better if I can understand this and also would be helpful to me as I am thinking of picking up the implementation of your locking service.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -1319,4 +1350,34 @@ private boolean rerunWF(String workflowId, String taskId, Map<String, Object> ta
...
     }
+
+    private boolean acquireLock(String workflowId) {
+        if (config.enableWorkflowExecutionLock()) {
+            if (!lockService.acquireLock(workflowId, 2, TimeUnit.MILLISECONDS)) {
+                LOGGER.info("Thread {} failed to acquire lock to workflowId {}.", Thread.currentThread().getId(), workflowId);
+                Monitors.recordAcquireLockUnsuccessful(workflowId);
+                return false;
+            }
+            LOGGER.info("Thread {} acquired lock to workflowId {}.", Thread.currentThread().getId(), workflowId);
```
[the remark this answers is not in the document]
milongapau: yes, this should be DEBUG level.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
milongapau: typo... good catch.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
milongapau: these are meant as helper methods specifically for execution.  If we want to lock for other reasons, a different string would be passed in (to lock API) as lockId.  Will see if it makes sense to make these methods more general.

mudit3774: I understand that are specific to workflow, I was suggesting that if we want to lock the workflow for execution and workflow for completion then it would be a problem. We can consider having a function like :

`
String getWorkflowLockId(String workflowId, String lockContext) {
                return getHash(workflowId, lockContext);
}

waitForLock(getWorkflowLockId("asdxxjhkadkshd", LOCKED_FOR_EXECUTION));
waitForLock(getWorkflowLockId("asdxxjhkadkshd", LOCKED_FOR_COMPLETION));
`

core/src/main/java/com/netflix/conductor/core/utils/NoopLock.java
```diff
@@ -0,0 +1,22 @@
+package com.netflix.conductor.core.utils;
+
+import java.util.concurrent.TimeUnit;
+
+public class NoopLock implements Lock {
+    @Override
+    public void acquireLock(String lockId) {}
+
+    @Override
+    public Boolean acquireLock(String lockId, long timeToTry, TimeUnit unit) {
```
naveenchlsn: Do we need an object here?  Can we use primitive so null pointer exceptions are avoided

core/src/main/java/com/netflix/conductor/core/utils/Lock.java
```diff
@@ -0,0 +1,45 @@
...
+    void acquireLock(String lockId);
+
+    /**
+     * acquires a re-entrant lock on lockId, blocks for timeToTry duration before giving up
+     * @param lockId resource to lock on
+     * @param timeToTry blocks up to timeToTry duration in attempt to acquire the lock
+     * @param unit time unit
+     * @return
+     */
+    Boolean acquireLock(String lockId, long timeToTry, TimeUnit unit);
```
naveenchlsn: Can we have input param a container object(LockOptions) and return a LockItem object. 

Advantages are : 
1) If we want to take in a new parameter for acquiring Lock. we would not have overloaded methods with a lot of parameters. For example, if we want a Lock Implementation which supports how long a lease can be taken. 

2) If we have a LockItem object which has when the lock has been acquired and how much time is still remaining. 
He can try to reacquire the lock before proceeding.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
milongapau: @naveenchlsn see the example in analysis section of documentation, the issue is a workflow can be executed more than once resulting in inconsistency.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
milongapau: @mudit3774 workflowCompletion is part of execution, and would use the same lock.  Happy to continue this conversation, I'm merging this PR first to move forward.

PR #1162, opened 2019-05-25
added LWT when updating workflow and added a update counter on workflow

apanicker-nflx:


apanicker-nflx:
Closing since this is no longer needed with the distributed lock changes.

PR #1166, opened 2019-05-29
honor optional flag in join task

apanicker-nflx:


core/src/main/java/com/netflix/conductor/core/orchestration/ExecutionDAOFacade.java
```diff
@@ -103,18 +104,20 @@ public Workflow getWorkflowById(String workflowId, boolean includeTasks) {
...
-                try {
-                    Workflow workflow = getWorkflowById(workflowId, includeTasks);
-                    workflows.add(workflow);
-                } catch (ApplicationException e) {
-                    //This might happen when the workflow archival failed and the workflow was removed from dynomite
-                    LOGGER.error("Error getting the workflowId: {}  for correlationId: {} from Dynomite/Archival", workflowId, correlationId, e);
-                }
-            });
-            return workflows;
+            SearchResult<String> result = indexDAO.searchWorkflows("correlationId='" + correlationId + "'", "*", 0, 1000, null);
```
apanicker-nflx: This method is called from the API, so using parallel stream processing here to improve performance.

core/src/main/java/com/netflix/conductor/core/orchestration/ExecutionDAOFacade.java
```diff
@@ -103,18 +104,20 @@ public Workflow getWorkflowById(String workflowId, boolean includeTasks) {
      */
     public List<Workflow> getWorkflowsByCorrelationId(String correlationId, boolean includeTasks) {
```
apanicker-nflx: This method is called from the API, so using parallel stream processing here to improve performance.

PR #1167, opened 2019-05-29
Lock implementation using Zookeeper

milongapau:
Zookeeper lock implementation.
@kishorebanala @apanicker-nflx @cyzhao

### DIFF
--- a/zookeeper-lock/src/main/java/com/netflix/conductor/zookeeper/ZkLock.java
+++ b/zookeeper-lock/src/main/java/com/netflix/conductor/zookeeper/ZkLock.java
@@ -1,0 +1,85 @@
+package com.netflix.conductor.zookeeper;
+
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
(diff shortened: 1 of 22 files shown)


zookeeper-lock/src/main/java/com/netflix/conductor/zookeeper/ZkLock.java
```diff
@@ -0,0 +1,85 @@
...
+import org.apache.curator.framework.CuratorFrameworkFactory;
+import org.apache.curator.framework.recipes.locks.InterProcessMutex;
+import org.apache.curator.retry.ExponentialBackoffRetry;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.inject.Inject;
+import java.util.concurrent.TimeUnit;
+
+public class ZkLock implements Lock {
```
kishorebanala: ZkLock seems ambiguous. Is this is a common practice in Zookeeper? If not, it's probably better to use full names.

milongapau: It is standard shorthand, can spell it out too.  For now will merge first to setup testing.

Issue #1031, opened 2019-03-06
Construct of "goto" task.


naveenchlsn:
Hi, 
We are interested in adding loop construct to conductor

There are similar discussions in old tickets.  https://github.com/Netflix/conductor/issues/938. 

One approach would be to add goto construct. In addition, have the ability to run a task multiple times, 
saving each execution. 

@cyzhao @apanicker-nflx @v1r3n @mashurex @kishorebanala  Can you please let us know your thoughts, so we can work on this.

andrea11:
Hello @naveenchlsn,
Indeed this will be very helpful. But they could be considered as two different features.
I would be interested as well in knowing the position of @cyzhao @apanicker-nflx @v1r3n @mashurex @kishorebanala about that.

Were you thinking about a static loop or a dynamic loop (like FORK or DYNAMIC_FORK) ?

cyzhao:
@naveenchlsn contributions are always welcome, please provide details on how you plan to add the support so we can discuss here, once we all agree with the approach then you can start working on it.

naveenchlsn:
Thanks @cyzhao.

### **Proposal.**
To support goto contruct in the workflow def, we can have a field "next" in the worflowtask,
containing the taskReferenceName of the task we want to execute after the current task.

### **Rationale** 
This approach is similar to what [States Language](https://states-language.net/spec.html)  proposes.

### **Implementation.**
We can enhance the WorkflowDef.getNextTask to return the "next" Task, falling back to old logic if next field is absent.Currently, each workflowTask corresponds to one task , because task instance is uniquely identified by taskReferenceName.
In contrast, we need to make task identity independent of taskReferenceName, by using a UUID instead.
This would require changes in database layer and other places where we identify task by referenceTaskName.

### **Backward Compatability.** 
If the next field is absent we fall back to old logic,  maintaining backward compatability. 


Please let me know your thoughts.  @cyzhao @apanicker-nflx @mashurex @kishorebanala

apanicker-nflx:
Hi @naveenchlsn, thanks for the proposal. We will review this and get back to you soon with feedback.

mudit3774:
@apanicker-nflx Any update on the proposal ? @naveenchlsn I think it would make sense to add more details in the implementation section. For instance : 

1. How will we maintain history of all runs ? 
2. How will the registration change ? 
3. How will the execution change ?

kishorebanala:
[...]

PR #1169, opened 2019-05-30
Loop functionality

manan164:
This PR contains loop functionality in conductor. 
I have added two json field named loopCondition and loopOver.
loopCondition is the condition which needs to be evaluated.
loopOver is the list of tasks that we want to execute in loop. Sample json is given below.
loopCondition will be evaluated every time the loopOver task will be executed. If the condition is true then loopOver tasks will be scheduled. 

Pseudo code 
In general, we return true from execute methods when task status is changed inside execute method.
So overall we will return false till loop task is not completed.
 
```
Wait till all loop task status become terminal.
If (all executed successfully)
    Evaluate condition.
    if(condition is true)
        Schedule loop task with refname as refname__iteration.
    else
        Mark loop task completed and come out of loop.
else
    Mark loop task failed and come out of loop.
```
Sample json:- 
```
{
  "ownerApp": "Test",
  "createTime": 1562568126745,
  "createdBy": "Manan",
  "name": "TestWF",
  "description": "DoWhileWF",
  "version": 1,
  "tasks": [
    {
      "name": "dowhiletask1",
      "taskReferenceName": "dowhiletask1",
      "inputParameters": {
        "value": "${workflow.input.loop}"
      },
      "type": "DO_WHILE",
      "startDelay": 0,
      "optional": false,
      "asyncComplete": false,
      "loopCondition": "if ( ($.dowhiletask1['iteration'] < $.value) && ($.Http1['response']['body'] > 10)) { true; } else { false; }",
      "loopOver": [
        {
          "name": "Fork",
          "taskReferenceName": "Fork",
          "type": "FORK_JOIN",
          "forkTasks": [
            [
              {
                "name": "Http1",
                "taskReferenceName": "Http1",
                "inputParameters": {
                  "http_request": {
                    "method": "GET",
                    "uri": "http://localhost:8081",
                    "accept": "application/json",
                    "contentType": "application/json"
                  }
                },
                "type": "HTTP",
                "startDelay": 0,
                "optional": false
              }
            ],
            [
              {
                "name": "Http2",
                "taskReferenceName": "Http2",
                "inputParameters": {
                  "http_request": {
                    "method": "GET",
                    "uri": "http://localhost:8081",
                    "accept": "application/json",
                    "contentType": "application/json"
                  }
                },
                "type": "HTTP",
                "startDelay": 0,
                "optional": false
              }
            ]
          ]
        },
        {
          "name": "Join",
          "taskReferenceName": "Join",
          "type": "JOIN",
          "joinOn": [
            "Http1",
            "Http2"
          ]
        }
      ]
    }
  ],
  "schemaVersion": 2,
  "restartable": true,
  "workflowStatusListenerEnabled": false
}
```


core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java
```diff
@@ -0,0 +1,164 @@
...
+ * @author Manan
+ *
+ */
+public class DoWhile extends WorkflowSystemTask {
+
+	public DoWhile() {
+		super("DO_WHILE");
+	}
+
+	Logger logger = LoggerFactory.getLogger(DecisionTaskMapper.class);
```
mudit3774: Please use the logger name same as the class name.

core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java
```diff
@@ -0,0 +1,164 @@
...
+		List<String> loopOver = (List<String>) task.getInputData().get("loopOver");
+
+		int iteration = iterations.get(task.getTaskId()) == null ? 0 : iterations.get(task.getTaskId());
+
+		for(String loopOverRef : loopOver){
+			String refName = loopOverRef;
+			if (iteration > 0) {
+				refName += "_" + (iteration);
+			}
+			Task forkedTask = workflow.getTaskByRefName(refName);
```
mudit3774: Can you please help me understand why we have a forkedTask?

core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java
```diff
@@ -0,0 +1,164 @@
...
+		boolean allDone = true;
+		boolean hasFailures = false;
+		StringBuilder failureReason = new StringBuilder();
+		List<String> loopOver = (List<String>) task.getInputData().get("loopOver");
+
+		int iteration = iterations.get(task.getTaskId()) == null ? 0 : iterations.get(task.getTaskId());
+
+		for(String loopOverRef : loopOver){
+			String refName = loopOverRef;
+			if (iteration > 0) {
```
mudit3774: Consider moving this to a separate function.

Maybe 

getRefNameForIteration(refName, iteration)

core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java
```diff
@@ -0,0 +1,164 @@
...
+		}
+	}
+
+	Map<String, Object> getConditionInput(Task task, int iteration, Workflow workflow) {
+		Map<String, Object> map = new HashMap<>();
+		List<String> loopOverTask = task.getWorkflowTask().getLoopOver();
+		for (String taskName : loopOverTask) {
+			String refName = taskName;
+			if (iteration > 0) {
+				refName += "_" + (iteration);
```
mudit3774: Consider moving this to a separate function and "_" to constant.

core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java
```diff
@@ -0,0 +1,164 @@
...
+			}
+			Task temp = workflow.getTaskByRefName(refName);
+			map.putAll(temp.getOutputData());
+			logger.debug("Taskname {} output {}", temp.getReferenceTaskName(), temp.getOutputData());
+		}
+		return map;
+	}
+
+	boolean scheduleLoopTasks(Task task, Map<String, Integer> iterations, Workflow workflow, WorkflowExecutor provider) {
+		List<String> loopOver = (List<String>) task.getInputData().get("loopOver");
```
mudit3774: Consider moving string to constant.

core/src/main/java/com/netflix/conductor/validations/WorkflowTaskTypeConstraint.java
```diff
@@ -60,6 +60,9 @@ public boolean isValid(WorkflowTask workflowTask, ConstraintValidatorContext con
                 case TaskType.TASK_TYPE_DECISION:
                     valid = isDecisionTaskValid(workflowTask, context);
                     break;
+                case TaskType.TASK_TYPE_DO_WHILE:
```
mudit3774: Consider adding this as the last case.

core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java
```diff
@@ -0,0 +1,164 @@
...
+import java.util.Map;
+
+/**
+ * @author Manan
+ *
+ */
+public class DoWhile extends WorkflowSystemTask {
+
+	public DoWhile() {
+		super("DO_WHILE");
```
naveenchlsn: Please make this a constant.

core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java
```diff
@@ -0,0 +1,164 @@
...
+		Map<String, Object> taskInput = getConditionInput(task, iteration, workflow);
+		String condition = task.getWorkflowTask().getLoopCondition();
+		boolean shouldContinue = false;
+		if (condition != null) {
+			logger.debug("Case being evaluated using decision expression: {}", condition);
+			try {
+				//Evaluate the expression by using the Nashhorn based script evaluator
+				shouldContinue = ScriptEvaluator.evalBool(condition, taskInput);
+			} catch (ScriptException e) {
+				logger.error(e.getMessage(), e);
```
naveenchlsn: Consider putting a log message.

grpc/src/main/proto/model/workflowtask.proto
```diff
@@ -36,4 +36,6 @@ message WorkflowTask {
     bool rate_limited = 20;
     repeated string default_exclusive_join_task = 21;
     bool async_complete = 23;
+    string loop_condition = 24;
```
naveenchlsn: I think this and AbstractProtoMapper are autogenerated files by gradle. Consider not checking them in.

manan164:
> I see that we have taken an approach in which we have defined the task that DO_WHILE needs to loop over before the actual DO_WHILE task which is similar to a JOIN construct.
> 
> Can you please help me understand why this was chosen over a FORK like semantics in which the loop tasks could have been defined inside the task?
> 
> I believe defining the do-while like fork would make more sense because loop-body should be inside the loop def.
> 
> Thoughts?

Hi, I tried returning loopOver tasks in getMappedTask for do while task but the issue is in second time iteration. How we are going to iterate again. Also as far as this PR is concern my worry is I am calling scheduleTask from execute method, that two are different concern. Also if we are making do while like fork then we also need a way to handle input output for each loopOver task. Let me try once again.

core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java
```diff
@@ -0,0 +1,164 @@
...
+		return true;
+	}
+
+	@VisibleForTesting
+	boolean getEvaluatedCondition(Task task,int iteration, Workflow workflow) {
+		Map<String, Object> taskInput = getConditionInput(task, iteration, workflow);
+		String condition = task.getWorkflowTask().getLoopCondition();
+		boolean shouldContinue = false;
+		if (condition != null) {
+			logger.debug("Case being evaluated using decision expression: {}", condition);
```
naveenchlsn: Consider adding log for result of condition evaluation as well. Would be helpfull

core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java
```diff
@@ -0,0 +1,164 @@
...
+	@Override
+	@SuppressWarnings("unchecked")
+	public boolean execute(Workflow workflow, Task task, WorkflowExecutor provider) {
+		
+		boolean allDone = true;
+		boolean hasFailures = false;
+		StringBuilder failureReason = new StringBuilder();
+		List<String> loopOver = (List<String>) task.getInputData().get("loopOver");
+
+		int iteration = iterations.get(task.getTaskId()) == null ? 0 : iterations.get(task.getTaskId());
```
naveenchlsn: We are maintaining "iterations" map in-memory and updating the iteration number of each task. 
But since the map is in-memory, when the task is scheduled and dequeued in other JVMs, the execution history will be overwritten. Also, consider the case when a JVM goes down, we will lose execution history. 

IMO. The iteration number of each task should be a member of task instance, not DoWhile Task.

andrea11:
Hello,
Nice feature :) Could you please provide more info on how this task should be used? (maybe updating the documentation)

manan164:
> Hello,
> Nice feature 😄 Could you please provide more info on how this task should be used? (maybe updating the documentation)

Hi @andrea11 Sure will update the doc.

core/src/main/java/com/netflix/conductor/core/execution/SystemTaskType.java
```diff
@@ -21,19 +21,16 @@
 import java.util.HashSet;
 import java.util.Set;
 
-import com.netflix.conductor.core.execution.tasks.Decision;
-import com.netflix.conductor.core.execution.tasks.ExclusiveJoin;
-import com.netflix.conductor.core.execution.tasks.Fork;
-import com.netflix.conductor.core.execution.tasks.Join;
-import com.netflix.conductor.core.execution.tasks.WorkflowSystemTask;
+import com.netflix.conductor.core.execution.tasks.*;
```
shivambansal2222: Would it make more sense to import specific task instead of including all the tasks?

core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java
```diff
@@ -0,0 +1,166 @@
...
+		List<WorkflowTask> loopOver = task.getWorkflowTask().getLoopOver();
+		List<Task> taskToBeScheduled = new ArrayList<>();
+		int iteration = task.getIteration() + 1;
+		for(WorkflowTask loopOverRef : loopOver){
+			String refName = getTaskRefName(loopOverRef, task);
+			Task existingTask = workflow.getTaskByRefName(refName);
+			existingTask.setRetried(true);
+			Task newTask = provider.taskToBeRescheduled(existingTask);
+			newTask.setSeq(existingTask.getSeq());
+			newTask.setReferenceTaskName(loopOverRef.getTaskReferenceName() + "_" + iteration);
```
manan164: Scheduling the task again with refname as refname_iteration.

core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java
```diff
@@ -0,0 +1,166 @@
...
+		}
+		return refName;
+	}
+
+	Map<String, Object> getConditionInput(Task task, Workflow workflow) {
+		Map<String, Object> map = new HashMap<>();
+		List<WorkflowTask> loopOverTask = task.getWorkflowTask().getLoopOver();
+		for (WorkflowTask workflowTask : loopOverTask) {
+			Task temp = workflow.getTaskByRefName(getTaskRefName(workflowTask, task));
+			map.put(workflowTask.getTaskReferenceName(), temp.getOutputData());
```
manan164: Input to a condition will still be original taskrefname only. So we will pick latest iteration output as input to condition.

core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java
```diff
@@ -0,0 +1,168 @@
...
+			}
+		}
+		if (hasFailures) {
+			return markLoopTaskFailed(task, failureReason.toString());
+		} else if (!allDone) {
+			return false;
+		}
+		boolean shouldContinue = getEvaluatedCondition(task, workflow);
+		logger.debug("taskid {} condition evaluated to {}",task.getTaskId(), shouldContinue);
+		if (shouldContinue){
```
kishorebanala: Nit: Please use standard spacing all along. Eg: `if (shouldContinue) {`, `if (!allDone || hasFailures) {`, ```for (WorkflowTask workflowTask : loopOver) {``` from above.

core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java
```diff
@@ -0,0 +1,168 @@
...
+		Map<String, Object> taskInput = getConditionInput(task, workflow);
+		String condition = task.getWorkflowTask().getLoopCondition();
+		boolean shouldContinue = false;
+		if (condition != null) {
+			logger.debug("Condition {} is being evaluated{}", condition);
+			try {
+				//Evaluate the expression by using the Nashhorn based script evaluator
+				shouldContinue = ScriptEvaluator.evalBool(condition, taskInput);
+			} catch (ScriptException e) {
+				logger.error("Unable to evaluate condition " + condition + " Exception ", e);
```
apanicker-nflx: Consider marking the loop task as failed here instead of throwing an exception.

core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java
```diff
@@ -0,0 +1,168 @@
...
+		logger.debug("Scheduling loop tasks for taskid {} as condition {} evaluated to true",
+				task.getTaskId(), task.getWorkflowTask().getLoopCondition());
+		List<WorkflowTask> loopOver = task.getWorkflowTask().getLoopOver();
+		List<Task> taskToBeScheduled = new ArrayList<>();
+		int iteration = task.getIteration() + 1;
+		for(WorkflowTask loopOverRef : loopOver){
+			String refName = getTaskRefName(loopOverRef, task);
+			Task existingTask = workflow.getTaskByRefName(refName);
+			existingTask.setRetried(true);
+			Task newTask = provider.taskToBeRescheduled(existingTask);
```
apanicker-nflx: Correct me if I got this wrong, but this is a new iteration of the task to meet given condition and not a retry of this task. I think that mixing these up can muddle up behavior for when this task actually needs to be retried in a FAILED scenario.

Please add some integration tests to `AbstractWorkflowServiceTest` with both happy and sad paths(with retry).

core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java
```diff
@@ -0,0 +1,168 @@
...
+			newTask.setSeq(existingTask.getSeq());
```
apanicker-nflx: Why update the sequence number here, since this is auto-populated within the `scheduleTask` method?

core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java
manan164: Hi @apanicker-nflx , Yes. It can cause issue. I have to form a logic to know wether this task will ever get retry or not.

core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java
```diff
@@ -0,0 +1,168 @@
...
+			newTask.setReferenceTaskName(loopOverRef.getTaskReferenceName() + "_" + iteration);
+			newTask.setRetryCount(existingTask.getRetryCount());
+			newTask.setScheduledTime(System.currentTimeMillis());
+			taskToBeScheduled.add(newTask);
+		}
+		task.setIteration(iteration);
+		provider.scheduleTask(workflow, taskToBeScheduled);
```
manan164: I have to make scheduleTask method public so that task will be scheduled on immediate basis here. I feel this violation of separation of concern. Task execute method should not schedule another task.

So I tried different approach to come up with loop. I changed https://github.com/manan164/conductor/blob/do-while-conductor/core/src/main/java/com/netflix/conductor/core/execution/mapper/DoWhileTaskMapper.java#L48 method and evaluate condition. If that condition is true then schedule the task again and if that false do not schedule. Decision task does that.

Can you people tell what is the good way of doing this. @naveenchlsn @mudit3774 @kishorebanala @apanicker-nflx @shivambansal2222

common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowTask.java
```diff
@@ -160,6 +160,12 @@ public void setTasks(List<WorkflowTask> tasks) {
 	@ProtoField(id = 23)
 	private Boolean asyncComplete = false;
 
+	@ProtoField(id = 24)
+	private String loopCondition;
+
+	@ProtoField(id = 25)
+	private List<WorkflowTask> loopOver = new LinkedList<>();
```
shivambansal2222: I guess this file is auto generated using protocol buffers. We should not manually update this.

manan164:
Hi @apanicker-nflx @kishorebanala  Have written integration test. Please review.

core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java
```diff
@@ -0,0 +1,140 @@
...
+			if (hasFailures){
+				failureReason.append(loopOverTask.getReasonForIncompletion()).append(" ");
+			}
+			output.put(workflowTask.getTaskReferenceName(), loopOverTask.getOutputData());
+			allDone = taskStatus.isTerminal();
+			if (!allDone || hasFailures){
+				break;
+			}
+		}
+		task.getOutputData().put(String.valueOf(task.getIteration()), output);
```
manan164: @shivambansal2222  storing output of each iteration with key as iteration. See if this can be useful for you.

core/src/main/java/com/netflix/conductor/core/execution/mapper/DoWhileTaskMapper.java
```diff
@@ -0,0 +1,87 @@
...
+        loopTask.setReferenceTaskName(taskToSchedule.getTaskReferenceName());
+        loopTask.setWorkflowInstanceId(workflowInstance.getWorkflowId());
+        loopTask.setCorrelationId(workflowInstance.getCorrelationId());
+        loopTask.setWorkflowType(workflowInstance.getWorkflowName());
+        loopTask.setScheduledTime(System.currentTimeMillis());
+        loopTask.setTaskId(taskId);
+        loopTask.setStatus(Task.Status.IN_PROGRESS);
+        loopTask.setWorkflowTask(taskToSchedule);
+
+        tasksToBeScheduled.add(loopTask);
```
manan164: Add it before so while deletion we can delete the task which has higher seq number directly. It is used https://github.com/Netflix/conductor/pull/1169/files#diff-a28de3002a75934bb7c693d03b674144R1336

core/src/main/java/com/netflix/conductor/core/execution/mapper/DoWhileTaskMapper.java
```diff
@@ -0,0 +1,87 @@
...
+        Task loopTask = new Task();
+        loopTask.setTaskType(SystemTaskType.DO_WHILE.name());
+        loopTask.setTaskDefName(taskToSchedule.getName());
+        loopTask.setReferenceTaskName(taskToSchedule.getTaskReferenceName());
+        loopTask.setWorkflowInstanceId(workflowInstance.getWorkflowId());
+        loopTask.setCorrelationId(workflowInstance.getCorrelationId());
+        loopTask.setWorkflowType(workflowInstance.getWorkflowName());
+        loopTask.setScheduledTime(System.currentTimeMillis());
+        loopTask.setTaskId(taskId);
+        loopTask.setStatus(Task.Status.IN_PROGRESS);
```
shivambansal2222: @manan164 If a loopLask is simple (worker) task , it will be first "Scheduled" and then marked "IN_PROGRESS" when a worker polls for it.

docs/docs/configuration/systask.md
```diff
@@ -552,3 +552,51 @@ Task Input payload sample
 ```
 
 The task is marked as ```FAILED``` if the message could not be published to the Kafka queue. 
+
+
+## Do While Task
+
```
andrea11: It could be useful to add information about the output as well, and eventually about how to reference looped tasks inside another task.

core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java
andrea11: Could it not be preferable to store the output directly inside an Array, using the index as information about the iteration?

manan164: Initially I had thought about it storing array way, but Storing it map way has advantage because it will automatically update the value to latest. Map has inbuilt facility for that. Also output map can be filled partially depend upon which task completed and which task remaining so map is better.

docs/docs/configuration/systask.md
manan164: Sure. will add up here.

core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java
andrea11: Ok, but in this case why do you stringify the iteration? It would not be better to have as key the same type of _iteration_ attribute (integer) ?

core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java
apanicker-nflx: IMO, do not use the `READY_FOR_RERUN` status. We would like to deprecate/remove this status in an upcoming version. 
Are you storing the history of past executions of the loop task?

manan164: Hi @apanicker-nflx Thanks for the reply. I have changed my approach. Will add in this pr after sufficient dev testing. Now I am deleting execution history of each task after terminal execution and store the output of each iteration in do while task as iteration as key. The approach mentioned in this PR will not work for loopOver task like fork join and decision because I was changing task reference name. I have to maintain original taskRefName in order to get task mapper work properly in second iteration. If storing history is concern then I think while storing I can append iteration.

core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java
manan164: No specific reason for this. Will change in subsequent commit.

andrea11:
Hello @manan164,
any news on this PR? I was also thinking, that it could be useful to have a parameter to set a delay between loop. What do you think?

manan164:
Hi @andrea11 . I will update the PR in some days. For delay between loop, initially I had thought but our use case is like loop over task will have combination of fork and decision task along with join at last. So JOIN will wait for all task to get completed. So in a way it is kind of delay but yes, I can explore how we will be able to accommodate it.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -1431,4 +1433,32 @@ private boolean rerunWF(String workflowId, String taskId, Map<String, Object> ta
         }
         return false;
     }
+
+    public void removeLoopOverTasks(Task loopTask, Workflow workflow) {
+        // Remove all tasks after the "DO_WHILE" task
+        for(Task task: workflow.getTasks()) {
```
manan164: @apanicker-nflx, @kishorebanala, @mudit3774, @naveenchlsn Even after removing task after every iteration I still see that workflow has the task already. This happens because any of loopover task will get updated in executionDAO after do while task gets updated. I am scheduling task immediately after removing from executionDAO and so I am getting all the task from task mapper properly. I feel this is not the good way of doing it. We have to remove from executionDAo otherwise scheduling second iteration will be a problem. Suggestion are welcomed. On approach I see is to rename a task with iteration and store it. But in that case also original task must be removed.

docs/docs/configuration/systask.md
manan164: @andrea11 Updated the doc. Please let me know if you find it still unclear how to use do while task.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
apanicker-nflx: Have you considered an approach similar to the `FORK..JOIN` concept which constitutes two system tasks?
Essentially, have `DO...WHILE` as two separate system tasks. 
1. DO - This takes a list of tasks that need to be executed in a loop as the input and is executed only once.
2. WHILE - This takes an expression to be evaluated as input. It could also have a configurable flag to control if iteration history needs to be saved. This task could be executed multiple times.

When the workflow is run, the `DO` task moves directly to `COMPLETED` state and schedules a `WHILE` task. The `WHILE` task could then schedule the first task in the list or complete execution based on the condition evaluation. It can also decide on saving iteration history based on the flag. This approach could probably alleviate some of the hurdles. 
Let me know your thoughts.

manan164: Hi @apanicker-nflx . Thanks for the suggestion. 
1) If we go with DO...WHILE , still the issue to schedule second iteration will be problem unless I change the task reference name in every iteration and schedule it. Issue is, in second iteration even though task mapper gives correct task but it will be omitted in DeciderService.getTasksToBeScheduled because workflow will have that task with same reference name.
2) In case of changing the task reference name and scheduling it we have to maintain the list of task that we have scheduled last and wait for all of them to get completed before scheduling second set. Consider a case like DO..WHILE (HTTP1,HTTP2). In this case while scheduling second iteration, we will schedule HTTP1_2 forcefully (Create new task object). When it gets completed its next task will be null (ideally we can make HTTP2 but then we have to construct task object). So when execute of WHILE will get called we have to know that HTTP1 is completed and now time to schedule HTTP2. So we have to store list of last scheduled tasks.
To solve this problem we have to change the way DeciderService.getTasksToBeScheduled works.

manan164: Hi @apanicker-nflx Let me know your thoughts on this. Let's close this as soon as possible.

apanicker-nflx: Hi @manan164 Thanks for the response. 
1. I feel that since we are violating the DAG structure for a workflow with the concept of loops, we should introduce a new field `iteration` on the task which can be used alongwith `taskReferenceName` to ensure that duplicate tasks are not scheduled within an execution. The logic in the task mappers and `tasksToBeScheduled` would then need to be updated to account for this. 
2. This new field also enables the re-use of the current logic for finding the next task from a given task by not changing the `taskReferenceName`, but rather updating the iteration. 

Let me know if this approach sounds reasonable.

apanicker-nflx:
@manan164 Please also add some validations for the new task type to ensure that all constraints are being met.

manan164:
> @manan164 Please also add some validations for the new task type to ensure that all constraints are being met, similar to what is done [here](https://github.com/Netflix/conductor/blob/master/core/src/main/java/com/netflix/conductor/validations/WorkflowTaskTypeConstraint.java)

Hi @apanicker-nflx WorkflowTaskType validations are at :- https://github.com/Netflix/conductor/pull/1169/files#diff-d421b44afaf1173e37c7ae99613eaeb4R82.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
manan164: Hi @apanicker-nflx Iteration is already there https://github.com/Netflix/conductor/pull/1169/files#diff-5ef655e1d0e0f90d2cbc1edca630bae4R193. But currently exposed only for dowhile task. 
There are two ways to solve this.
1) While scheduling second iteration the task ref names will be like ``` taskRefName_iteration```. So while checking condition before scheduling next iteration, we will strip ```_iteration``` in case loopOver task is used in loopCondition. Also while getting the next task I will strip down ```iteration```. In this case I have to book keep what are the task scheduled currently as a part of second iteration. This is what I was mentioning in my previous comment.
2) Do not change task ref name in second iteration. Instead change the way taskMapper gives the task to DeciderService.getTasksToBeScheduled and change the executionDAO.createTasks to check iteration also. Currently it checks retry count only and that is the reason we update retryCount every time task is rescheduled. In this case also I need the book keeping mentioned in point 1.
Will create working model for point 1 so that things will be clear.

manan164: Hi @apanicker-nflx I have made changes with the approach mentioned in point 1. Used delimiter ```:__:```. So task will be scheduled as taskRefName:__:iteration. Same will be shown in task details.  I have avoided book keeping list with the help of changing the DeciderService.getNextTask. Please have a look. Some nit picks are there but will change once we finalise the approach.

apanicker-nflx: @manan164 This approach looks good and very clean. Please fix the failing tests and add sufficient unit tests to cover all edge cases. Thanks a lot for your contribution.

manan164:
Hi @apanicker-nflx Only few integration tests is pending. Please review.

core/src/main/java/com/netflix/conductor/core/execution/ParametersUtils.java
```diff
@@ -114,7 +116,7 @@ public ParametersUtils() {
                     taskParams.put("reasonForIncompletion", task.getReasonForIncompletion());
                     taskParams.put("callbackAfterSeconds", task.getCallbackAfterSeconds());
                     taskParams.put("workerId", task.getWorkerId());
-                    inputMap.put(task.getReferenceTaskName(), taskParams);
+                    inputMap.put(task.isLoopOverTask() ? DoWhileTaskMapper.getTaskDefReferenceName(task.getReferenceTaskName()) :  task.getReferenceTaskName(), taskParams);
```
manan164: This is required to support input parameters in loopOver tasks as we are changing task refname in each iteration.

manan164:
Hi @apanicker-nflx Waiting for your response in this PR.

manan164:
Hi @apanicker-nflx  Any update on this thread?

apanicker-nflx:
@manan164 Apologies for the delay, will get back to you soon regarding this.

manan164:
> @manan164 Apologies for the delay, will get back to you soon regarding this.

Hi @apanicker-nflx  Thanks for the update.

core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
```diff
@@ -331,7 +351,7 @@ private boolean checkForWorkflowCompletion(final Workflow workflow) throws Termi
             }
         }
 
-        String taskReferenceName = task.getReferenceTaskName();
+        String taskReferenceName = task.isLoopOverTask() ? DoWhileTaskMapper.getTaskDefReferenceName(task.getReferenceTaskName()) : task.getReferenceTaskName();
```
apanicker-nflx: This looks leaky to me that the specific TaskMapper implementation is explicitly invoked here.

core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
```diff
@@ -217,6 +221,22 @@ private DeciderOutcome decide(final Workflow workflow, List<Task> preScheduledTa
         return outcome;
     }
 
+
+    List<Task> filterNextLoopOverTasks(List<Task> task, Task pendingTask, Workflow workflow) {
```
apanicker-nflx: Please change variable name `task -> tasks/taskList`.

core/src/main/java/com/netflix/conductor/core/execution/ParametersUtils.java
apanicker-nflx: similar leaky abstraction as the comment above. 

Can this static method be moved out of the TaskMapper class?

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -1464,4 +1467,22 @@ private boolean rerunWF(String workflowId, String taskId, Map<String, Object> ta
         }
         return false;
     }
+
+    public void scheduleNextIteration(Task loopTask, Workflow workflow) {
+        List<Task> scheduledLoopOverTasks = deciderService.getTasksToBeScheduled(workflow, loopTask.getWorkflowTask().getLoopOver().get(0), loopTask.getRetryCount(), null);
+        scheduledLoopOverTasks.stream().forEach(t -> {
+            t.setReferenceTaskName(DoWhileTaskMapper.appendIteration(t.getReferenceTaskName(), loopTask.getIteration()));
```
apanicker-nflx: similar leak here, probably can move static methods out of TaskMapper.

core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
```diff
@@ -187,6 +188,9 @@ private DeciderOutcome decide(final Workflow workflow, List<Task> preScheduledTa
             if (!pendingTask.isExecuted() && !pendingTask.isRetried() && pendingTask.getStatus().isTerminal()) {
                 pendingTask.setExecuted(true);
                 List<Task> nextTasks = getNextTask(workflow, pendingTask);
+                if (pendingTask.isLoopOverTask() && !nextTasks.isEmpty()) {
```
manan164: Here is the core logic. @apanicker-nflx While calling getnext task I am passing original ref name so taskmapper will work and while scheduling I change task ref name. So in general if task is part of loopOver Task then task will not be scheduled with original taskrefname which is different behaviour compare to existing.

andrea11:
Hi @manan164, I did not use this task, but I am very interested in it. I looked at the documentation and it seems clear :) 
Is there any limitation to the tasks that can be used inside this task?

manan164:
> Hi @manan164, I did not use this task, but I am very interested in it. I looked at the documentation and it seems clear :)
> Is there any limitation to the tasks that can be used inside this task?

Hi @andrea11 ,
The only limitation as of now is, We can not use task type SUB_WORKFLOW as loopover task. that I am checking when onboarding a workflow. Will work on this once base PR is merged.
Other than this you can use any of the task mentioned in [TaskTypes](https://github.com/Netflix/conductor/blob/master/common/src/main/java/com/netflix/conductor/common/metadata/workflow/TaskType.java#L8) as a part of loopover task.
The only thing to keep in mind is, this is not go to task so I am not checking constraints mentioned in the [comment](https://github.com/Netflix/conductor/issues/1031#issuecomment-507501877).
The loopover task will be executed in a loop so crossing branching will not be supported. Ideally, we should avoid having such definitions.

docs/docs/configuration/systask.md
apanicker-nflx: Can you also call out the limitation here? From this [comment](https://github.com/Netflix/conductor/pull/1169#issuecomment-539589612).

apanicker-nflx:
@manan164 Great work and follow-through on this PR that has been ongoing work for a few months over few iterations. 

Please fix the conflict and update the documentation with the limitation for this task and we are good to merge.

manan164:
Hi @apanicker-nflx Updated the doc.

sharmasourabh:
Thanks @manan164 for PR.

@apanicker-nflx Could you please share in which version, you are planning to release this change?

apanicker-nflx:
@manan164 @sharmasourabh This is released as part of [v2.17.0](https://github.com/Netflix/conductor/releases/tag/v2.17.0)

PR #1221, opened 2019-07-24
allow adding task definitions for system tasks to configure timeouts etc

### LINKED ISSUE #1068: Can I set a timeout for a WAIT TASK

apanicker-nflx:
Closes #1068

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
@@ -156,6 +156,10 @@ public class DeciderService {
             }
 
             Optional<TaskDef> taskDefinition = pendingTask.getTaskDefinition();
+            if (!taskDefinition.isPresent()) {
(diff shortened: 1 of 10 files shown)


core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
```diff
@@ -156,6 +156,10 @@ private DeciderOutcome decide(final Workflow workflow, List<Task> preScheduledTa
             }
 
             Optional<TaskDef> taskDefinition = pendingTask.getTaskDefinition();
+            if (!taskDefinition.isPresent()) {
+               taskDefinition = Optional.ofNullable(workflow.getWorkflowDefinition().getTaskByRefName(pendingTask.getReferenceTaskName()))
+                       .map(WorkflowTask::getTaskDefinition);
+            }
 
             if (taskDefinition.isPresent()) {
```
kishorebanala: Is this if check still required?

apanicker-nflx: Yes, if no task definition is provided, it will be `Optional.empty`

PR #1234, opened 2019-08-02
Fix tasks removal from elasticsearch

andrey-borue:
When deleting workflow via api:
1. DB: entries from `workflows` and `tasks` are deleted.
2. Elasticsearch: only entries from the `workflow` index are deleted, but entries from the `tasks` index remains.

### DIFF
--- a/es6-persistence/src/main/java/com/netflix/conductor/dao/es6/index/ElasticSearchDAOV6.java
+++ b/es6-persistence/src/main/java/com/netflix/conductor/dao/es6/index/ElasticSearchDAOV6.java
@@ -68,6 +68,7 @@ import java.io.IOException;
 import java.text.SimpleDateFormat;
 import java.time.LocalDate;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.LinkedList;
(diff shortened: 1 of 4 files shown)


es6-persistence/src/main/java/com/netflix/conductor/dao/es6/index/ElasticSearchDAOV6.java
```diff
@@ -467,20 +468,31 @@ private void updateWithRetry(UpdateRequest request, String operationDescription)
...
+            logger.error("Failed to remove " + docType + " {} from index", docId, e);
             Monitors.error(className, "remove");
         }
     }
 
+    @Override
+    public void removeWorkflow(String workflowId) {
+        removeDocument(workflowIndexName, WORKFLOW_DOC_TYPE, workflowId);
+
+        List<String> taskIds = searchTasks("", "workflowId:\"" + workflowId + "\"", 0, 100, Collections.emptyList()).getResults();
```
kishorebanala: So, the searchTasks parameters are start=0, and count=100. What if the workflow has more than 100 tasks?

apanicker-nflx:
Closing due to inactivity, please reopen if you are able to resume working on this.

Issue #1238, opened 2019-08-07
Add header to TaskClient besides root URI


EngAhmedAdel:
Hi,

We need to create task workers, and first part of that is to create a new TaskClient and connect it with conductor server.

The api only provides a way of setting rootURI `taskClient.setRootURI(String)`, we need to provide a header to specify the `Host` along with that root URI in order for the requests to reach conductor server (as it needs to go through envoy first).

Is there a way of doing it?

davidwadden:
@EngAhmedAdel,

This can be done by using the constructor overload for [`TaskClient(ClientConfig, ClientHandler, ClientFilter...)`](https://github.com/Netflix/conductor/blob/8fc13b09347383c746fb789240114860ac945354/client/src/main/java/com/netflix/conductor/client/http/TaskClient.java#L96), then implementing a [ClientFilter](http://javadox.com/com.sun.jersey/jersey-client/1.19/com/sun/jersey/api/client/filter/ClientFilter.html) to append the header to the request.

An example can be found in the [`HTTPBasicAuthFilter.java`](https://github.com/jersey/jersey-1.x/blob/1.19.3/jersey-client/src/main/java/com/sun/jersey/api/client/filter/HTTPBasicAuthFilter.java)

David

EngAhmedAdel:
@davidwadden,

Thanks for your reply. I tried to use the `ClientFilter` as you mentioned, but the header doesn't seem to be set!

All requests to the server ends up with 404 responses!

This is how my client filter looks like:
```
import com.sun.jersey.api.client.ClientHandlerException;
import com.sun.jersey.api.client.ClientRequest;
import com.sun.jersey.api.client.ClientResponse;
import com.sun.jersey.api.client.filter.ClientFilter;

import javax.ws.rs.core.HttpHeaders;

public class HttpHostFilter extends ClientFilter {
    private static final String HOST_NAME = "my-host-name";

    @Override
    public ClientResponse handle(final ClientRequest cr) throws ClientHandlerException {
        if (!cr.getHeaders().containsKey(HttpHeaders.HOST)) {
            cr.getHeaders().add(HttpHeaders.HOST, HOST_NAME);
        }
        return getNext().handle(cr);
    }
}
```

And this is the TaskClient initialization:
```
TaskClient taskClient = new TaskClient(new DefaultClientConfig(),
                                            new DefaultConductorClientConfiguration(),
                                            null,
                                            new HttpHostFilter());
// Point the client to the server API
[...]

PR #1241, opened 2019-08-08
Add request headers in Client Base

EngAhmedAdel:
As mentioned in issue: https://github.com/Netflix/conductor/issues/1238

Root URL only is not always enough to reach conductor server if requests need to go through some proxy (i.e Envoy).

This pull request allows the client to use headers while connecting to the server.

Thanks.


EngAhmedAdel:
Using ClientFilter would solve this issue, so I'm closing this pull request as it's not needed.

PR #1247, opened 2019-08-09
Adds max.block.ms config and caching of Kafka producer creation. 

naveenchlsn:
max.block.ms  - If the Kafka is not reachable producer blocks for a minute. This is not ideal as it blocks the execution for a long time. I have added the config to enable users to specify the block time, by default the task blocks for 500ms. 

caching of Kafka producer - Bulk of the task execution time is consumed by producer creation and connection establishment. This PR enables users to cache the producer once created and increase the throughput. By default 10 producers are cached. 

@kishorebanala  @apanicker-nflx  Please consider the PR if you think this is a useful addition to the conductor.

### DIFF
--- a/contribs/build.gradle
+++ b/contribs/build.gradle
@@ -19,7 +19,6 @@ dependencies {
 	compile "io.swagger:swagger-jaxrs:${revSwagger}"
 
 	compile "org.apache.kafka:kafka-clients:2.2.0"
-	compile "com.google.guava:guava:${revGuava}"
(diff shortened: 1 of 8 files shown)


kishorebanala:
@naveenchlsn Looks good to me. Please resolve the conflicts when you have a chance.

naveenchlsn:
> @naveenchlsn Looks good to me. Please resolve the conflicts when you have a chance.

Thank you @kishorebanala. I have updated the PR. please take a look.

PR #1260, opened 2019-08-17
Remove taskIds that are in the queue without a valid task execution.

kishorebanala:


core/src/main/java/com/netflix/conductor/service/ExecutionService.java
```diff
@@ -130,6 +130,9 @@ public Task poll(String taskType, String workerId, String domain) {
 			for (String taskId : taskIds) {
 				Task task = getTask(taskId);
 				if (task == null) {
+					// Remove taskId(s) without a valid Task from the queue.
+					queueDAO.remove(queueName, taskId);
+					logger.info("Removed taskId without a valid task from queue: {}", taskId);
```
apanicker-nflx: Also log the queue name here.

Issue #1264, opened 2019-08-19
MYSQL v3 migration contains invalid MYSQL syntax(only supported by MariaDB)?


s50600822:
From:
https://github.com/Netflix/conductor/blob/master/mysql-persistence/src/main/resources/db/migration/V3__queue_add_priority.sql

```
ALTER TABLE `queue_message` ADD COLUMN IF NOT EXISTS `priority` TINYINT DEFAULT 0 AFTER `message_id`;
```
ADD COLUMN IF NOT EXISTS does not seem to be a valid ALTER TABLE option: https://dev.mysql.com/doc/refman/8.0/en/alter-table.html

As a result:
```
mysql> ALTER TABLE 'queue_message' ADD COLUMN IF NOT EXISTS 'priority' TINYINT DEFAULT 0 AFTER 'message_id';
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ''queue_message' ADD COLUMN IF NOT EXISTS 'priority' TINYINT DEFAULT 0 AFTER 'mes' at line 1
```

An interesting side effect of this is that even though the migration failed, flyway recorded this as done in schema_version
<img width="1231" alt="Screen Shot 2019-08-19 at 10 15 06 pm" src="https://user-images.githubusercontent.com/5586453/63264629-e73b8580-c2ce-11e9-9aed-6e09a02ce7b6.png">

Which fails startup later and require manual fix in database. Subsequent startup will not show the SQL syntax error.
```
delete  from schema_version where installed_rank =3;
```

Upgrading flyway fixes the issue where flyway.migrate() combine with com.netflix.conductor.mysql.MySQLDataSourceProvider.get() and Guice creates a vicious cycle of unlimited connection pool until:
```
10005 [main] ERROR com.zaxxer.hikari.pool.HikariPool  - HikariPool-21 - Exception during pool initialization.
java.sql.SQLNonTransientConnectionException: Data source rejected establishment of connection,  message from server: "Too many connections"
	at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:108)
	at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:95)
	at com.mysql.cj.jdbc.exceptions.SQLExceptionsMapping.translateException(SQLExceptionsMapping.java:122)
[...]

PR #1272, opened 2019-08-21
Fix_V3_MySQL_Migration

s50600822:
Fix migration issue when running with MySQL, for more detail see:
https://github.com/Netflix/conductor/issues/1264


s50600822:
See https://github.com/Netflix/conductor/pull/1280 instead or minimal change.

PR #1273, opened 2019-08-21
Allow compatibility with modules in go1.11+ 

tartale:
With the enabling of go modules in go 1.11+, the `GOPATH` will be going away.  The imports in this code only work if the source code is installed under `${GOPATH}/conductor`, which necessitates the need for the `install.sh` scripts.

By changing the local imports, you no longer need to run `install.sh`, and go modules users can work with the code outside of the `GOPATH`.

Existing users with imports on `conductor/*` will still need to install the source code to the `$GOPATH`, but until they migrate their imports, they won't be able to update to using the go modules feature.

I'd like the original author input as well (TIA @cquon !)


tartale:
> Can you expand on this? I _think_ I get you, but I'd like to make sure.

Absolutely - if a user of this library has already run the `install.sh` script, and has the following in their own code:
```go
import "conductor"
...
    conductorClient := conductor.NewConductorHttpClient("http://localhost:8080")
```

When they upgrade to this new version of the library, they need to revise this import to:
```go
import "github.com/netflix/conductor/client/go"
```
Additionally, if they are not using go modules, the source code will need to be moved to `$GOPATH/github.com/netflix/conductor/client/go`


> Also, how does this PR affect the instructions provided in the [README](https://github.com/Netflix/conductor/blob/master/client/go/README.md)?

Ah, good point.  These instructions would need revision; should I add this to the PR?

kchristidis:
Thanks for clarifying!

Yes, let's please update the README and those changes to this PR. Thanks!

tartale:
@kchristidis - I've updated the readme; additionally, I removed the install/uninstall scripts, as they're no longer needed

tartale:
@apanicker-nflx @kchristidis 

ping!

kchristidis:
Sorry for the delay. Thanks for following up on this! LGTM.

PR #1274, opened 2019-08-21
Update timeout description

jimmort25:
it says the `timeoutSeconds` field takes its input in terms of milliseconds, when in reality, it takes it in terms of seconds. :)


apanicker-nflx:
Thanks for pointing out the error and the fix.

PR #1280, opened 2019-08-25
wrong MySQL syntax

s50600822:


demichej:
@jvemugunta @mashurex Could y'all take a look at this? Or if you are no longer able to contribute, can you at least respond here so that we're aware if you're looking at this? Thanks.

mashurex:
@s50600822 @apanicker-nflx  @demichej I'm sorry it took so long for me to see this! Apparently Gmail has been suppressing all my Github notifications. The MySQL V3 migration script shouldn't just work w/ MariaDB as I wrote and tested it against MySQL 5.7. 

My initial worry about this change is that the user this migration runs at may not have SELECT privileges to the `INFORMATION_SCHEMA`. I'm not sure if all users get this by default or not, from memory I believe I've run in to issues with selecting from system schemas without granting extra privileges to users.

If `INFORMATION_SCHEMA` privileges aren't an issue than this looks like a more portable migration.

apanicker-nflx:
@s50600822 Please respond to the comments/make changes as suggested.

s50600822:
@mashurex your concern is reasonable.
From my experience using the fixed script(I applied the fixed last month to my forked since this took a bit long) in different env (my different locals with mysql on the host sometimes and in docker container and also on mysql aurora of AWS) it works just fine. 

https://dev.mysql.com/doc/refman/5.7/en/information-schema-introduction.html
"The same privileges apply to selecting information from INFORMATION_SCHEMA and viewing the same information through SHOW statements. In either case, you must have some privilege on an object to see information about it."

I wouldn't try to make the script or code to be that smart, that automatically fixes the migration history. Considering ppl who got hit by the issue would have failed to startup and flyway won't go pass V3 for them. Also the situation of MariaDB folks and MySQL's now are different. I'd rather being explicit about the problem and let the developer handle it once(https://github.com/Netflix/conductor/pull/1293 should cover that aspect) - if the checksum does not match it will print both and the dev will know how to fix that for once. I myself did the same for my pipeline cause the first time we tried to run up conductor that has V3 script it affected the schema_version permanently, conductor won't startup afterward.

Having certain conditions for the migration script to run is a normal thing(flyway itself required a few things ppl know it when they see it). Has the original script have used a function then we won't need to change it (and just need to let ppl know to define the function). But it's too late for that I think. 

I'd say since there's no way we can covered every scenario don't try to fixed the issue in the code but let give enough information for the developer to be able to fix up and retry. We can either include the comment in the V3 script linked to this issue or put something like that in next release change log.

mashurex:
@apanicker-nflx @s50600822 I reviewed this and everything is working against MySQL 5.7 as well as the tests passing. 

One other thought, if we're breaking the migration hash chain for this, couldn't we just change the V1 migration to include this column (as well as the V2 changes)? It would seem cleaner going forward.

apanicker-nflx:
@mashurex Thank you for reviewing and for the comments.
@s50600822 Please add comments to the V3 script detailing the change and we can include these in the release log as well for better visibility.

sp1986sp:
I having this issue with latest released jar : https://repo1.maven.org/maven2/com/netflix/conductor/conductor-server/2.22.2/

Is there any work-around?

apanicker-nflx:
@s50600822 Pinging as reminder.

RileyMichael:
@sp1986sp you can run the updated sql manually, then mark the v3 schema migration as successful (schema_version table).

cschuyle:
The following works for me on the fork that we're using:

- delete the current `V3__queue_add_priority.sql` migration file.
- create a new one called `V3__queue_add_priority_corrected.sql`

If the old migration has already been executed, the new migration will be a no-op.
If the old migrations hasn't yet been executed, it'll create the new column.

This is a minimally-invasive change.

ritu-p:
Can this fix get merged?

PR #1311, opened 2019-09-11
add GetWorkflows and GetTasks to gRPC service implementation

goldibex:
This is a small patch to add two gRPC metadata service methods, `GetWorkflows` and `GetTasks`. They are functionally equivalent to the REST API methods.

A small note: the new override in `ProtoMapper` is required because the auto-generator for `AbstractProtoMapper` does not properly implement null checking for map types. Without it you'll get a `NullPointerException` at runtime.

### DIFF
--- a/grpc-server/src/main/java/com/netflix/conductor/grpc/server/service/MetadataServiceImpl.java
+++ b/grpc-server/src/main/java/com/netflix/conductor/grpc/server/service/MetadataServiceImpl.java
@@ -34,6 +34,15 @@ public class MetadataServiceImpl extends MetadataServiceGrpc.MetadataServiceImpl
     }
 
     @Override
+    public void getWorkflows(MetadataServicePb.GetWorkflowsRequest req, StreamObserver<MetadataServicePb.GetWorkflowsResponse > response) {
+        response.onNext(
+            MetadataServicePb.GetWorkflowsResponse.newBuilder().addAllWorkflows(
+                service.getWorkflowDefs().stream().map(PROTO_MAPPER::toProto)::iterator
+            ).build());
+        response.onCompleted();
+    }
+
+    @Override
     public void createWorkflow(MetadataServicePb.CreateWorkflowRequest req, StreamObserver<MetadataServicePb.CreateWorkflowResponse> response) {
         WorkflowDef workflow = PROTO_MAPPER.fromProto(req.getWorkflow());
         service.registerWorkflowDef(workflow);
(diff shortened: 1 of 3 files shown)


kishorebanala:
Hey @josedab, This looks good to me. But, because you're one of the original authors of this module, can you help review this please.

goldibex:
ping?

grpc-server/src/main/java/com/netflix/conductor/grpc/server/service/MetadataServiceImpl.java
```diff
@@ -33,6 +33,15 @@ public MetadataServiceImpl(MetadataService service) {
         this.service = service;
     }
 
+    @Override
+    public void getWorkflows(MetadataServicePb.GetWorkflowsRequest req, StreamObserver<MetadataServicePb.GetWorkflowsResponse > response) {
```
josedab: **Suggestion**: can we use `getWorkflowDefs` or `getWorkflowDefinitions`?
Workflow objects are instances, while definitions (`WorkflowDef` objects) are within those instances.

grpc-server/src/main/java/com/netflix/conductor/grpc/server/service/MetadataServiceImpl.java
```diff
@@ -70,6 +79,15 @@ public void getWorkflow(MetadataServicePb.GetWorkflowRequest req, StreamObserver
         }
     }
 
+    @Override
+    public void getTasks(MetadataServicePb.GetTasksRequest req, StreamObserver<MetadataServicePb.GetTasksResponse > response) {
```
josedab: **Suggestion**: can we use `getTaskDefs` or `getTaskDefinitions`?
Similar concept stated with workflows but with task definitions: A task is a runtime object that contains a task definition.

grpc-server/src/main/java/com/netflix/conductor/grpc/server/service/MetadataServiceImpl.java
josedab: **Question**: I can see this working with immutable workflows (where the definition is embedded inside). Have you tried running this with a legacy workflow (where the definition is stored on db)?
In addition, If you want to get promoted this to `Netflix:dev` or production on their side, a test might help gain their confidence ;)

apanicker-nflx:
@josedab Thank you very much for reviewing.
@goldibex Please make changes as suggested and add some integration tests [here](https://github.com/Netflix/conductor/blob/master/test-harness/src/test/java/com/netflix/conductor/tests/integration/AbstractGrpcEndToEndTest.java)

apanicker-nflx:
@goldibex Are you available to continue working on this PR?

apanicker-nflx:
Closing due to inactivity. Please reopen if you want to resume working on this.

PR #1327, opened 2019-09-21
add postgres persistence

mactaggart:
This PR adds postgres persistence (Execution, Metadata, and Queue DAOs). The Postgres implementation is based heavily on the MySQL implementation with minor modifications to SQL syntax/types to support Postgres. A sql-persistence module was added that abstracts code common to both MySQL and Postgres. The sql-persistence module can be leveraged to add other SQL backends. SQL configuration properties have been renamed from "conductor.mysql.xx" to "conductor.sql.xx".


mactaggart:
@apanicker-nflx it would be great to get feedback on this implementation if there's interest in supporting postgres. Thanks!

mactaggart:
I rebased onto the latest in dev including PR #1299 which retries transactions that fail due to MySQL deadlock. I'm uncertain if Postgres will exhibit deadlock behavior. (I haven't experienced this in testing and don't know how to reproduce) Since I've extracted code common to MySQL and Postgres into the sql-persistence module, I included the retry logic in sql-persistence, however, as written only MySQL deadlocks will be caught and retried as I don't know what errors Postgres may throw. Any guidance would be appreciated.

docker/server/config/config-mysql.properties
```diff
@@ -14,8 +14,8 @@ db=mysql
 jdbc.url=jdbc:mysql://mysql:3306/conductor
 
 # Hikari pool sizes are -1 by default and prevent startup
-conductor.mysql.connection.pool.size.max=10
-conductor.mysql.connection.pool.idle.min=2
+conductor.sql.connection.pool.size.max=10
```
kishorebanala: This change would be incompatible to existing users.

kishorebanala:
@mactaggart Apologies for the delays, and thanks for your contribution. We really appreciate it. However, the changes to property names and module structure is not backward compatible, especially for users who have customized.

To expedite making Postgres module available to community, please create a separate independent module for Postgres. And in next major release, we shall revisit to splitting the module to have a shared `sql-persistence-module`. Also, please consider keeping cleanup changes in other modules in a separate PR - which are a great value addition, but makes it easy to review when they're separated.

mactaggart:
@kishorebanala thanks for taking time to review this! I'll create a separate independent module and submit another PR.

mactaggart:
a new PR [1368](https://github.com/Netflix/conductor/pull/1368) has been submitted containing requested changes.

Issue #1316, opened 2019-09-13
Make Conductor model extendable


andrea11:
Hello,
I was thinking to "change" (more precisely extend) my data model of conductor. Today I could fork the project and edit directly. But I was thinking to a more generic approach, using in the interface only generics that extend your base model. For example:

```java
public interface ExecutionDAO<T extends Task, W extends Workflow> {
    List<T> getPendingTasksByWorkflow(String taskName, String workflowId);
    W getWorkflow(String workflowId, boolean includeTasks);
```

What do you think? Is there a better approach to avoid this huge refactoring?

Jiehong:
Hello @andrea11,

I think something more in this vein might make more sense if I understand your point well. Here with the ExecutionDAO as well:

```java
public interface ExecutionDAO {
  <T extends Task> List<T> getPendingTasksByWorkflow(String var1, String var2);

  <T extends Task> List<T> getTasks(String var1, String var2, int var3);

  <T extends Task> List<T> createTasks(List<T> var1);

  <T extends Task> void updateTask(T var1);
}
```

As a `T` should always be a valid `Task` in this case, very few issues should occur I reckon, but it's a breaking change, indeed.

What do you think?

Jiehong:
but I'd be curious to see if a Task is a valid T, but that might be the case

Jiehong:
However, for now, if you implement DAOs with a T that actually extends your model, you should be fine with upcasting/downcasting, even though it's ugly.

Jiehong:
> 
> 
> but I'd be curious to see if a Task is a valid T, but that might be the case

Well, after testing, it seems it is indeed perfectly valid:

```java
public class Test {
  private <T extends String> void handle(final T input) {
    return;
  }
  
  private void start() {
    this.handle("Hello");
  }
}
```

andrea11:
@apanicker-nflx @kishorebanala @naveenchlsn Any suggestion about it? It could be something that could be accepted by/interesting for the community? Me and my team were thinking to prepare a first draft PR to give a better overview. 

Waiting for your advice 😃 
Thanks

naveenchlsn:
@andrea11  Can you please let me know what are the benefits of run-time binding of data models? 
What functionality in the service classes/api's are we aiming to achieve with this change?

apanicker-nflx:
[...]

PR #1336, opened 2019-09-30
Introducing generics at interfaces level

AlexVinogradov:
Introduced Generics in order to easily extend the current model.
This PR is linked to issue  https://github.com/Netflix/conductor/issues/1316

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/mapper/SubWorkflowTaskMapper.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/mapper/SubWorkflowTaskMapper.java
@@ -13,7 +13,9 @@
 package com.netflix.conductor.core.execution.mapper;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.netflix.conductor.common.metadata.events.EventHandler;
 import com.netflix.conductor.common.metadata.tasks.Task;
+import com.netflix.conductor.common.metadata.tasks.TaskDef;
 import com.netflix.conductor.common.metadata.workflow.SubWorkflowParams;
 import com.netflix.conductor.common.metadata.workflow.WorkflowDef;
 import com.netflix.conductor.common.metadata.workflow.WorkflowTask;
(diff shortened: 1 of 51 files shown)


core/src/main/java/com/netflix/conductor/core/execution/mapper/SubWorkflowTaskMapper.java
```diff
@@ -116,7 +116,7 @@ private Integer getSubWorkflowVersion(Map<String, Object> resolvedParams, String
                 .map(Object::toString)
                 .map(Integer::parseInt)
                 .orElseGet(
-                        () -> metadataDAO.getLatest(subWorkflowName)
+                        () -> ((Optional<WorkflowDef>)metadataDAO.getLatest(subWorkflowName))
```
Jiehong: Is it really `WorkflowDef` here or an extension?

core/src/main/java/com/netflix/conductor/core/metadata/MetadataMapperService.java
```diff
@@ -125,7 +125,7 @@ private void populateVersionForSubWorkflow(WorkflowTask workflowTask) {
         if (subworkflowParams.getVersion() == null) {
             String subWorkflowName = subworkflowParams.getName();
             Integer subWorkflowVersion =
-                    metadataDAO.getLatest(subWorkflowName)
+                ((Optional<WorkflowDef>)metadataDAO.getLatest(subWorkflowName))
```
Jiehong: Same

core/src/main/java/com/netflix/conductor/dao/ExecutionDAO.java
```diff
@@ -214,7 +214,7 @@
 	 * @return List of workflows by correlation id
 	 *  
 	 */
-	List<Workflow> getWorkflowsByCorrelationId(String correlationId, boolean includeTasks);
+	<T extends Workflow> List<T> getWorkflowsByCorrelationId(String correlationId, boolean includeTasks);
```
Jiehong: You already call it `W`, so reused it instead of using `T extends Workflow`

core/src/main/java/com/netflix/conductor/core/execution/mapper/SubWorkflowTaskMapper.java
AlexVinogradov: Actually here it does not matter, since this code will only use the parent class (in this case WorkflowDef). Tried to introduce the generics at model level but i faced a limitation coming from Protogen - it cannot map generics :(

AlexVinogradov:
Hello @apanicker-nflx, Could you kindly please review this change? Thanks!

apanicker-nflx:
@AlexVinogradov Thanks for the pull request. However, since this changes the interface contracts and introduces incompatibility, we will hold off one merging this till the next major version upgrade.

apanicker-nflx:
Closing this for now due to incompatibility.

PR #1337, opened 2019-09-30
kafka persistance module

manan164:
Sample module for understanding.

Issue #1243, opened 2019-08-08
Break indexDAO to publisherDAO and indexDAO


manan164:
Hi all,
Currently for high throughput scenario we are facing issue while indexing to ElasticSearchDAO. 
We are dropping the message so we are not able to search in conductor UI.
I was thinking to break it into publisherDAO and consumerDAO. We can use kafka to publish to any kafka topic and indexDAO which will consume from kafka topic and index to indexDAO i.e. ElasticSearch.

Let me know your thoughts on this
@apanicker-nflx @kishorebanala @mudit3774 @cyzhao

manan164:
Another approach I tried was to send bulk request while indexing. As of now we bulk only task execution logs but we can bulk indexing task workflow and messages.

apanicker-nflx:
@manan164 I would suggest keeping the indexDAO as is but you can customize the implementation that is being used underneath to tailor it to your needs.
In your implementation, you can choose to break up the producer actions to write to an intermediate kafka layer which eventually sinks to an ES index and the consumer actions to directly read from ES. Also, you can choose to do bulk indexing in this approach.

This way the provided implementation can stay generic for wider usage and do feel free to submit your implementation as a contribution in a separate module, since this can be added as a niche optimization that other users can choose to adopt.

manan164:
Hi, @apanicker-nflx I am working on this approach. 
One potential issue I am facing is since the consumer has to know about indexDAO to index into ES and producer will override indexDAO method to publish to Kafka. 
1) Add a set of (sync and async) methods in indexDAO which will be eventually called by consumer service. We have to add a method because the producer will call the overridden method so ExecutionDAOFacade need not be changed.
2) Write completely another service for this. Which will not be in scope with the conductor.

apanicker-nflx:
Let me clarify what I was proposing earlier: 
> In your implementation, you can choose to break up the producer actions to write to an intermediate kafka layer which eventually sinks to an ES index

This could be implemented in a standalone implementation of the `IndexDAO` such that all the write operations (indexWorkflow, indexTask, updateWorkflow, removeWorkflow, indexEventExecution, etc) are published to a Kafka topic. The Kafka consumer here, does not need to be aware of the indexDAO implementation, rather can read messages from the Kafka topic and index to ES. The caveat here is that the message published to the topic needs to have details about the type of action (CREATE,UPDATE,DELETE), the document type and the index name. 

>and the consumer actions to directly read from ES.

[...]

PR #1338, opened 2019-09-30
fixes for indexing layer

apanicker-nflx:


es5-persistence/src/main/java/com/netflix/conductor/dao/es5/index/ElasticSearchDAOV5.java
```diff
@@ -149,6 +150,19 @@ public ElasticSearchDAOV5(Client elasticSearchClient, ElasticSearchConfiguration
                     logger.warn("Request {} to async dao discarded in executor {}", runnable, executor);
                     Monitors.recordDiscardedIndexingCount();
                 });
+
+        corePoolSize = 1;
+        maximumPoolSize = 2;
+        keepAliveTime = 30L;
+        this.logExecutorService = new ThreadPoolExecutor(corePoolSize,
```
manan164: Hi @apanicker-nflx We need same changes in ELasticSearchRestDAOV5 and ELasticSearchRestDAOV6

apanicker-nflx:
> Hi @apanicker-nflx We need same changes in ELasticSearchRestDAOV5 and ELasticSearchRestDAOV6

Hi @manan164 I will add these same changes to ES6 module as well, as soon as I complete my testing. Currently, I am testing with ES5 and varying loads.

For some short running workflows, I am running into an issue where the ES document could be updated with COMPLETED status first and then with RUNNING status, since a workflow could start and end on 2 different instances and depending on the worker queue size in each, documents could be indexed at different periods. I am currently, looking to address this using a [ScheduledThreadPoolExecutor](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ScheduledThreadPoolExecutor.html) which would index completion of short-running workflows. Wanted to know if you encountered any similar issues and how you mitigated them. I will make these additions in the meantime and test out.

manan164:
Hi @apanicker-nflx, What we can do here is have all indexing workflow operations should be sync. Also, we should minimize the number of indexing workflow calls. I think we have discussed this in other [thread](https://github.com/Netflix/conductor/issues/1243#issuecomment-522809341) But still, I feel the same problem will be there with the task also. 
Can you please elaborate on how you are trying to solve this by ScheduledThreadPoolExecutor. By using that we only ensure that tasks will be scheduled after a particular interval. Here what we lack is ordering guarantee. 
How we are trying to solve this is, publish all index operations to Kafka which will give ordering guarantee. And consume from Kafka. I have raised a sample [kafka-persistance](https://github.com/Netflix/conductor/pull/1337)  also. The only thing I want to know in that is how should we write consumers. Please go through that once. Again this is a niche operation and not all people will be interested for this. For those who are not interested we can use sync indexing only in case of a workflow. So basically I would say the flag you have used for this will separate out high throughput and low throughput scenario.
Views?

core/src/main/java/com/netflix/conductor/core/orchestration/ExecutionDAOFacade.java
```diff
@@ -358,4 +375,22 @@ public void addMessage(String queue, Message message) {
...
+        private String workflowId;
+
+        DelayWorkflowUpdate(String workflowId) {
+            this.workflowId = workflowId;
+        }
+
+        @Override
+        public void run() {
+            try {
+                Workflow workflow = executionDAO.getWorkflow(workflowId, false);
```
manan164: Hi @apanicker-nflx , I just wanted to tell that in high throughput scenario, we will not have data in Redis also. We remove it the moment it gets complete. What we can do here is have data while creating  DelayWorkflowUpdate object and schedule it to run after 30 seconds.

apanicker-nflx: >in high throughput scenario, we will not have data in Redis also. We remove it the moment it gets complete.

In this case, are you removing the workflow synchronously when workflow enters terminal state? I would suggest using the [searchArchivableWorkflows](https://github.com/Netflix/conductor/blob/master/es5-persistence/src/main/java/com/netflix/conductor/dao/es5/index/ElasticSearchDAOV5.java#L639) method in ES to remove the workflows using a background process. 

> What we can do here is have data while creating DelayWorkflowUpdate object and schedule it to run after 30 seconds.

I intentionally am passing `workflowId` only to `DelayWorkflowUpdate` object so as to reduce the memory footprint in this case. Since the `ScheduledThreadPoolExecutor` uses an unbounded queue, if the number of short running workflows are too large this would avoid having to store the entire workflow object with each Runnable.

apanicker-nflx:
Thanks for your suggestions and pointers.

> What we can do here is have all indexing workflow operations should be sync. Also, we should minimize the number of indexing workflow calls. I think we have discussed this in other [thread](https://github.com/Netflix/conductor/issues/1243#issuecomment-522809341) But still, I feel the same problem will be there with the task also.

In-sync indexing would not be ideal since this keeps elasticsearch in the request path, but users can opt to do that if the throughputs in their use-case is quite low. This can be controlled via configuration. 

> Can you please elaborate on how you are trying to solve this by ScheduledThreadPoolExecutor. By using that we only ensure that tasks will be scheduled after a particular interval. Here what we lack is ordering guarantee.

This is correct, be delaying the update operation by a fixed interval. This does not guarantee ordering but greatly reduces the probability of this inconsistency.

> How we are trying to solve this is, publish all index operations to Kafka which will give ordering guarantee. And consume from Kafka. I have raised a sample [kafka-persistance](https://github.com/Netflix/conductor/pull/1337) also. The only thing I want to know in that is how should we write consumers. Please go through that once. Again this is a niche optimisation and not all people will be interested for this. For those who are not interested we can use sync indexing only in case of a workflow. So basically I would say the flag you have used for this will separate out high throughput and low throughput scenario.
> Views?

We will be reviewing the Kafka-persistence PR soon. Eventually, in the long term, we would also be looking to move to Kafka as an intermediate layer for indexing. 

> The only thing I want to know in that is how should we write consumers.

Have you considered writing a Flink job to consume from the Kafka topic? This job could handle de-duping and merging before indexing the document to elasticsearch.

manan164:
> Have you considered writing a Flink job to consume from the Kafka topic? This job could handle de-duping and merging before indexing the document to elasticsearch.

Hi @apanicker-nflx,
I will sure explore Flink.
I want to keep the solution end to end in the conductor scope. That was the ultimate goal. I was using ScheduledThreadPoolExecutor only to read from kafka and index to ES. The only doubt was is it good to modify the indexDAO interface or not.
As of now, I am writing a spring boot application which has good [Kafka consumer library](https://mvnrepository.com/artifact/org.springframework.kafka/spring-kafka) which provides good retrying mechanism so that I don't have to write boilerplate code.

PR #1350, opened 2019-10-12
Graceful shutdown exec service

apanicker-nflx:


es5-persistence/src/main/java/com/netflix/conductor/dao/es5/index/ElasticSearchRestDAOV5.java
```diff
@@ -769,17 +794,19 @@ private void indexObject(final String index, final String docType, final String
         }
     }
 
-    private void indexBulkRequest(String docType) {
-        indexWithRetry(bulkRequests.get(docType).getBulkRequest(), "Bulk Indexing " + docType, docType);
-        bulkRequests.put(docType,new BulkRequests(System.currentTimeMillis(), new BulkRequest()));
+    private synchronized void indexBulkRequest(String docType) {
```
kishorebanala: What is the reason behind using `synchronized` here?

apanicker-nflx: the hashmap which is reset within this method, is accessed from multiple threads and we would want to synchronize this reset such that reset only occurs after the documents within the bulk request are indexed

PR #1353, opened 2019-10-15
Upgrade Jackson dependencies to v2.10.0

kishorebanala:


common/src/main/java/com/netflix/conductor/common/utils/JsonMapperProvider.java
```diff
@@ -111,11 +112,11 @@ public Any deserialize(JsonParser p, DeserializationContext ctxt)
                 JsonNode value = root.get(JSON_VALUE);
 
                 if (type == null || !type.isTextual()) {
-                    throw ctxt.reportMappingException("invalid '@type' field when deserializing ProtoBuf Any object");
+                    ctxt.reportMappingException("invalid '@type' field when deserializing ProtoBuf Any object");
```
apanicker-nflx: why is this exception not thrown?

PR #1362, opened 2019-10-22
Update logic for processing of unacks (mysql)

jdoupe:
I believe the intent for "processUnacks" and "processAllUnacks" is to un-ack messages that have been lingering for longer than a minute (60 seconds is hardcoded in the query).  The problem is that the current logic adds that minute to the current time (which always results as a minute into the future) and tests if it is greater than the deliver_on time. This is always going to return true.  The desired effect would be garnered by SUBTRACTing the 60 seconds from current time, thus comparing deliver_on to 1 minute in the past to see if the minute has passed.

The result of the current logic is that all messages that are APPROACHING deliver_on time (within the next 60 seconds) are unacked every time one of these functions is called, regardless of whether deliver_on has been reached, let alone passed.

### DIFF
--- a/mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLQueueDAO.java
+++ b/mysql-persistence/src/main/java/com/netflix/conductor/dao/mysql/MySQLQueueDAO.java
@@ -160,13 +160,13 @@ public class MySQLQueueDAO extends MySQLBaseDAO implements QueueDAO {
         logger.trace("processAllUnacks started");
 
 
-        final String PROCESS_ALL_UNACKS = "UPDATE queue_message SET popped = false WHERE popped = true AND TIMESTAMPADD(SECOND,60,CURRENT_TIMESTAMP) > deliver_on";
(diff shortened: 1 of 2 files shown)


kishorebanala:
@s50600822 @jvemugunta @mashurex As primary contributors to the MySQL module, we would appreciate it if you could help us out with reviewing this pull request.

mashurex:
@jdoupe The current query is specifically setting popped to `false` for all `popped = true` messages that have a `deliver_on` less than 1 minute from the instant the method is run. I'm honestly not sure why it is using 1 minute in the future as the measuring line, it appears that it was originally implemented with `CURRENT_TIMESTAMP > deliver_on`, we'll have to ask @jvemugunta what the reason for the change was.

It looks like the `RedisDynoQueue` implementation is un-acking up to `now`, so maybe we'd get more consistent behavior with no offset at all and go back to using `CURRENT_TIMESTAMP > deliver_on`- I just don't know what it breaks.

apanicker-nflx:
Closing due to inactivity, please reopen if you are available to resume working on this PR.

PR #1373, opened 2019-11-04
Goto Task implementation

shivambansal2222:
Hi @apanicker-nflx , 
                              Here is the pull request for GOTO task.  The major change is in the way "task_key" was treated earlier in database. Earlier it used to be a combination of "task_ref" and "retry_count", and now it is a combination of "task_ref", "retry_count" and "iteration_count" as a task can now run multiple times  with a different iteration_count. We can understand it better with the help of an example. 

Ex: Sample "GOTO" task:
 ```  
   {
    "name": "gotoLereview",
    "taskReferenceName": "gotoLereview",
    "inputParameters": {
      "fileName": "${MrReview.output.reviewFileName}"
    },
    "type": "GOTO",
    "gotoTask": "t1"
  } 
```
This "GOTO" construct is very similar to what we have in programming languages.  The "GOTO" task would take a task reference t1 where it would take the workflow execution. Task t1 will be created again with new task id. Data-passing from "GOTO" task to the previously executed task t1 will work with superimposed new input i.e. whatever input parameters in t1 matches with input parameters of GOTO, will be replaced with GOTO task's input.

Following is the sample two stage review workflow schema which uses "GOTO" construct:
   
  
```
{
  "name": "TwoStageReviewWithGoto",
  "description": "sample review workflow using GOTO",
  "version": 1,
  "tasks": [
    {
      "name": "review",
      "description": "legal review task",
      "taskReferenceName": "LeReview",
      "inputParameters": {
        "fileName": "${Workflow.input.reviewFileName}"
      },
      "type": "SIMPLE"
    },
    {
      "name": "review",
      "taskReferenceName": "MRreview",
      "description": "marketing review task",
      "inputParameters": {
        "fileName": "${LeReview.output.reviewFileName}"
      },
      "type": "SIMPLE"
    },
    {
      "name": "gotoDecision",
      "taskReferenceName": "gotoDecision",
      "inputParameters": {
        "reviewAccepted": "{MrReview.output.reviewRejected}"
      },
      "type": "DECISION",
      "caseValueParam": "reviewRejected",
      "decisionCases": {
        "1": [
          {
            "name": "gotoLereview",
            "taskReferenceName": "gotoLereview",
            "inputParameters": {
              "fileName": "${MrReview.output.reviewFileName}"
            },
            "type": "GOTO",
            "gotoTask": "LeReview"
          }
        ]
      }
    },
    {
      "name": "search_elasticsearch",
      "taskReferenceName": "get_es_1",
      "type": "HTTP",
      "inputParameters": {
        "http_request": {
          "uri": "http://localhost:9200/conductor/_search?size=10",
          "method": "GET"
        }
      }
    }
  ],
  "outputParameters": {
    "workflowIds": "${get_es_1.output..workflowId}",
    "status": "${get_es_1.output..status}"
  },
  "schemaVersion": 2
}
```

**Explanation:** A review file is taken as workflow input. The first task is a legal review (LeReview) which starts legal review. Then it's output file is shared for marketing review (MrReview). The MRreview task will generate output "reviewRejected" and based on it's value, workflow can either continue or it can go to previously executed task "LeReview". The "GOTO" task has an input parameter "fileName" which matches with "LeReview" task's input parameter "fileName", so it will superimpose value of "fileName" with the earlier one. This approach simplifies passing data to earlier executed task.

### DIFF
--- a/common/src/main/java/com/netflix/conductor/common/metadata/tasks/Task.java
+++ b/common/src/main/java/com/netflix/conductor/common/metadata/tasks/Task.java
@@ -198,6 +198,9 @@ public class Task {
     @ProtoField(id = 40)
     private int iteration;
 
+    @ProtoField(id = 41)
+    private int iterationCount;
+	
     public Task() {
     }
 
@@ -725,6 +728,20 @@ public class Task {
     public void setWorkflowPriority(int workflowPriority) {
         this.workflowPriority = workflowPriority;
     }
+	
+	/**
+     * @return the iterationCount
+     */
+    public int getIterationCount() {
+        return iterationCount;
+    }
+
+    /**
+     * @param iterationCount the iterationCount to set
+     */
+    public void setIterationCount(int iterationCount) {
+        this.iterationCount = iterationCount;
+    }
 
     public Task copy() {
         Task copy = new Task();
@@ -756,6 +773,7 @@ public class Task {
         copy.setIteration(iteration);
         copy.setExecutionNameSpace(executionNameSpace);
         copy.setIsolationGroupId(isolationGroupId);
+		copy.setIterationCount(iterationCount);
 
         return copy;
     }
@@ -769,6 +787,7 @@ public class Task {
                 ", inputData=" + inputData +
                 ", referenceTaskName='" + referenceTaskName + '\'' +
                 ", retryCount=" + retryCount +
+                ", iterationCount=" + iterationCount +
                 ", seq=" + seq +
                 ", correlationId='" + correlationId + '\'' +
                 ", pollCount=" + pollCount +
@@ -810,6 +829,7 @@ public class Task {
         if (o == null || getClass() != o.getClass()) return false;
         Task task = (Task) o;
         return getRetryCount() == task.getRetryCount() &&
+                getIterationCount() == task.getIterationCount() &&
                 getSeq() == task.getSeq() &&
                 getPollCount() == task.getPollCount() &&
                 getScheduledTime() == task.getScheduledTime() &&
@@ -851,6 +871,6 @@ public class Task {
 
     @Override
     public int hashCode() {
-        return Objects.hash(getTaskType(), getStatus(), getInputData(), getReferenceTaskName(), getWorkflowPriority(), getRetryCount(), getSeq(), getCorrelationId(), getPollCount(), getTaskDefName(), getScheduledTime(), getStartTime(), getEndTime(), getUpdateTime(), getStartDelayInSeconds(), getRetriedTaskId(), isRetried(), isExecuted(), isCallbackFromWorker(), getResponseTimeoutSeconds(), getWorkflowInstanceId(), getWorkflowType(), getTaskId(), getReasonForIncompletion(), getCallbackAfterSeconds(), getWorkerId(), getOutputData(), getWorkflowTask(), getDomain(), getInputMessage(), getOutputMessage(), getRateLimitPerFrequency(), getRateLimitFrequencyInSeconds(), getExternalInputPayloadStoragePath(), getExternalOutputPayloadStoragePath(), getIsolationGroupId(), getExecutionNameSpace());
+        return Objects.hash(getTaskType(), getStatus(), getInputData(), getReferenceTaskName(), getWorkflowPriority(), getRetryCount(), getSeq(), getCorrelationId(), getPollCount(), getTaskDefName(), getScheduledTime(), getStartTime(), getEndTime(), getUpdateTime(), getStartDelayInSeconds(), getRetriedTaskId(), isRetried(), isExecuted(), isCallbackFromWorker(), getResponseTimeoutSeconds(), getWorkflowInstanceId(), getWorkflowType(), getTaskId(), getReasonForIncompletion(), getCallbackAfterSeconds(), getWorkerId(), getOutputData(), getWorkflowTask(), getDomain(), getInputMessage(), getOutputMessage(), getRateLimitPerFrequency(), getRateLimitFrequencyInSeconds(), getExternalInputPayloadStoragePath(), getExternalOutputPayloadStoragePath(), getIsolationGroupId(), getExecutionNameSpace(), getIterationCount());
     }
 }
--- a/core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
@@ -202,9 +202,8 @@ public class DeciderService {
             }
         }
 
-        //All the tasks that need to scheduled are added to the outcome, in case of
+
         List<Task> unScheduledTasks = tasksToBeScheduled.values().stream()
-                .filter(task -> !executedTaskRefNames.contains(task.getReferenceTaskName()))
                 .collect(Collectors.toList());
         if (!unScheduledTasks.isEmpty()) {
             LOGGER.debug("Scheduling Tasks: {} for workflow: {}", unScheduledTasks.stream()
@@ -259,7 +258,7 @@ public class DeciderService {
             }
 
             //In case of a new workflow, the first non-skippable task will be scheduled
-            return getTasksToBeScheduled(workflow, taskToSchedule, 0);
+            return getTasksToBeScheduled(workflow, taskToSchedule, 0, 0);
         }
 
         // Get the first task to schedule
@@ -344,10 +343,15 @@ public class DeciderService {
 
     private List<Task> getNextTask(Workflow workflow, Task task) {
         final WorkflowDef workflowDef = workflow.getWorkflowDefinition();
+        
+        if(TaskType.GOTO.name().equals(task.getTaskType()))
+        	{
+        	    return Collections.emptyList();
+        	}
 
         // Get the following task after the last completed task
         if (SystemTaskType.is(task.getTaskType()) && SystemTaskType.DECISION.name().equals(task.getTaskType())) {
-            if (task.getInputData().get("hasChildren") != null) {
+            if (task.getInputData().get("hasChildren") != null || task.getInputData().get("gotoActivated") != null) {
                 return Collections.emptyList();
             }
         }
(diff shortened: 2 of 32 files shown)

### CODE REVIEW

**Comment by apanicker-nflx** [comment]
Thank you @shivambansal2222 for the pull request.

Pinging @naveenchlsn @andrea11 @mudit3774 @manan164 @choubisat since this is of interest from #1031 .

**Comment by manan164** [comment]
Hi, @apanicker-nflx  Thanks for pinging. @shivambansal2222  Will review it by tomorrow.

File: common/src/main/java/com/netflix/conductor/common/metadata/tasks/Task.java [hunk 1]
```diff
@@ -198,6 +198,9 @@ public boolean isRetriable() {
     @ProtoField(id = 40)
     private int iteration;
 
+    @ProtoField(id = 41)
```
manan164: We already have iteration as a part of the Task. Please evaluate if you can reuse that property.

core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
```diff
@@ -344,10 +343,15 @@ private boolean checkForWorkflowCompletion(final Workflow workflow) throws Termi
...
+        
+        if(TaskType.GOTO.name().equals(task.getTaskType()))
+        	{
+        	    return Collections.emptyList();
+        	}
 
         // Get the following task after the last completed task
         if (SystemTaskType.is(task.getTaskType()) && SystemTaskType.DECISION.name().equals(task.getTaskType())) {
-            if (task.getInputData().get("hasChildren") != null) {
+            if (task.getInputData().get("hasChildren") != null || task.getInputData().get("gotoActivated") != null) {
```
manan164: What is the significance of gotoActivated?

core/src/main/java/com/netflix/conductor/core/execution/tasks/Goto.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+public class Goto extends WorkflowSystemTask {
+	
+	public Goto() {
+		super("GOTO");
+	}
+	
+	@Override
+	public boolean execute(Workflow workflow, Task task, WorkflowExecutor provider) {
+		task.setStatus(Status.COMPLETED);
```
manan164: We can store more information as a part of goto task output.

common/src/main/java/com/netflix/conductor/common/metadata/tasks/Task.java
shivambansal2222: Thanks Manan. I checked that "iteration" was introduced as part of do-while task. So here is the problem in re-using the property. Do-while task and "Goto" interpret differently. The iteration is limited to the iteration of tasks running inside do-while container, while GOTO task consider the global state of this iteration. I am not sure whether re-using it will solve the problem. Let me know if there is any gap in my understanding.

core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
shivambansal2222: If 'GOTO' task is part of a decision task , then it will take the control of the workflow to the next task mentioned in 'gotoTask' field. The deciderService basically schedule next task defined in DSL, whenever a task (decision task in this case) is completed. Since control is already taken elsewhere by GOTO task, we don't want task subsequent to decision task to run. This property is being added in DecisionTaskMapper, whenever there is a GOTO task in selected branch of decision.

core/src/main/java/com/netflix/conductor/core/execution/tasks/Goto.java
shivambansal2222: Please suggest what other information we can store as part of gotoTask. 
The next Task of GOTO is already defined in workflow  DSL.

manan164:
Hi @shivambansal2222 , I have an overall look over the PR.
Please answer the following questions,
1) We are changing the way task keys are generated. [Key Definition](https://github.com/Netflix/conductor/blob/master/core/src/main/java/com/netflix/conductor/dao/ExecutionDAO.java#L62) says we should use only ref name and retryCount. So if we are changing the way keys are getting generate then we should update the method documentation in the interface. Another thing is, currently we have CassandraExecutionDAO and MySQLExecutionDAO. We might need to change there. Also in the future, if we are going to have another implementation we might need to change there. I was thinking to change in case of do while but later dropped the idea considering future maintenance.
2) Instead of changing all task mappers, we can change the way tasks are getting scheduled. Please explore how we are doing this in case of [do-while](https://github.com/Netflix/conductor/pull/1169#discussion_r327177275) The changes will be very minimal because we already have changed the way task keys are getting populated. In [GoToTaskMapper](https://github.com/Netflix/conductor/pull/1373/files#diff-02cefa0c6fb9c2346a4e8459eb37329cR101) update the iteration of the gototask and when it gets completed in [DeciderService](https://github.com/Netflix/conductor/blob/master/core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java#L192) automatically it will update the iterationCount of all subsequent tasks.
3) How are we handling the case of gotoTask will be subworkflow?

shivambansal2222:
> * Instead of changing all task mappers, we can change the way tasks are getting scheduled. Please explore how we are doing this in case of [do-while](https://github.com/Netflix/conductor/pull/1169#discussion_r327177275) The changes will be very minimal because we already have changed the way task keys are getting populated. In [GoToTaskMapper](https://github.com/Netflix/conductor/pull/1373/files#diff-02cefa0c6fb9c2346a4e8459eb37329cR101) update the iteration of the gototask and when it gets completed in [DeciderService](https://github.com/Netflix/conductor/blob/master/core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java#L192) automatically it will update the iterationCount of all subsequent tasks.

1. Yes, we need to change keyDefinition. As per primary key constraint, it should be unique to be added to the database. I'll update the key definition (comment in ExecutionDao) as mentioned by you.  I have already change the key definition code in MySqlExecutionDao and RedisExecutionDao. Since task_key is generated in code itself, we need not to change DB schema manually.  Moreover, we will get the whole workflow execution history ( same task with different iterationCount are treated differently) as it is persisted in db.

2. Thanks for the suggestion. I will have a look into it and refactor it accordingly.

3. GotoTask is not supported  to be  a subWorkflow currently. We will have to think more to make it work with subWorkflow as things  are bit complicated with subworkflow. We can treat it as limitation as of now.

andrea11:
Hi @shivambansal2222,
I would like to understand if in the workflow execution we will find all the execution of a same task (in your example LeReview), or only the last one?
If all of them are stored, will they have the same taskReferenceName (and different retryCount) ?
If yes, how can you distinguish between a fail of the task and an execution demanded by a GOTO task?
If no, how can you select the last execution via taskReferenceName as variable inside the workflow?

shivambansal2222:
> Hi @shivambansal2222,
> I would like to understand if in the workflow execution we will find all the execution of a same task (in your example LeReview), or only the last one?
> If all of them are stored, will they have the same taskReferenceName (and different retryCount) ?
> If yes, how can you distinguish between a fail of the task and an execution demanded by a GOTO task?
> If no, how can you select the last execution via taskReferenceName as variable inside the workflow?

HI @andrea11 , Thanks for having a look into this PR. 
 In workflow execution, we'll get record of all the executions of same task. So the newer execution of the same task will have higher 'iterationCount' but same taskReferenceName. I am using separate property 'iterationCount' rather than relying on existing 'retryCount' to make these 2 cases mentioned by you unambiguous. Using retryCount for newer iteration of same task is semantically incorrect as we need to distinguish between failure of a task and newer Iteration of such task due to 'GOTO'. 
Earlier  'taskRefName' + 'retryCount' was used as primaryKey while storing tasks in 'task' table, now 'taskRefName' + 'retryCount' + 'IterationCount' is used for the same. 
Please let me know if that clears your doubt.

core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
```diff
@@ -588,6 +600,7 @@ private void timeoutTask(TaskDef taskDef, Task task) {
                 .withTaskToSchedule(taskToSchedule)
                 .withTaskInput(input)
                 .withRetryCount(retryCount)
+                .withIterationCount(iterationCount)
                 .withRetryTaskId(retriedTaskId)
                 .withTaskId(taskId)
                 .withDeciderService(this)
```
sharmasourabh: There is a following statement at the end of method:
```
        List<Task> tasks = taskMappers.get(taskType.name()).getMappedTasks(taskMapperContext).stream()
                .filter(task -> !tasksInWorkflow.contains(task.getReferenceTaskName()))
                .collect(Collectors.toList());
```
Task with referenceTaskName _LeReview_ is already completed. Goto task again schedule it based on `gotoTask` field value. However, due to above filter it does not get scheduled.

Am I missing something?

shivambansal2222: @sharmasourabh Thanks for pointing it out. Actually this method has been changed recently, previously the filter was based on inProgress tasks only. So, if a task is not in progress, it can be scheduled again. I missed this change during the merge. I will update the PR with the change.

kushalhalder:
Hi @manan164, Is there a roadmap which release will this PR be merged to?

shivambansal2222:
@kushalhalder Apologies. I got occupied in something else and did not get time to incorporate review comments. I will get back to it once I get some free time.

kishorebanala:
Hey @shivambansal2222, bumping this up to see if you're still planning to work on this feature.

shivambansal2222:
> Hey @shivambansal2222, bumping this up to see if you're still planning to work on this feature.
HI @kishorebanala, 
                         I am currently pretty much occupied on some other high priority tasks. But I intend to pick it in future as soon as I get some time. Sorry for the delay.

amitfin:
Hi @shivambansal2222, we would also like to use GOTO functionality. Is this work still relevant?

Issue #1371, opened 2019-11-01
inputTemplate value not override by custom value


anxiaoyi:
define task:

```javascript
[{
    "name": "task_one",
    "retryCount": 2,
    "timeoutSeconds": 600,
    "inputTemplate": {
      "a": 100,
      "b": 200
    },
    "inputKeys": [
      "c"
    ],
    "outputKeys": [
      "result"
    ],
    "timeoutPolicy": "TIME_OUT_WF",
    "retryLogic": "FIXED",
    "retryDelaySeconds": 5,
    "responseTimeoutSeconds": 60
  }]
```

define workflow:

```javascript

  {
    "name": "task_one_workflow",
    "description": "task_one_workflow",
    "version": 1,
    "tasks": [
      {
        "name": "task_one",
        "taskReferenceName": "task_one",
        "inputParameters": {
          "a": "${workflow.input.a}",
          "b": "${workflow.input.b}",
          "c": "${workflow.input.c}"
        },
        "type": "SIMPLE",
        "startDelay": 0
      }
    ],
    "schemaVersion": 2
  }
```

start workflow with:

```javascript
{
    "name": "task_one_workflow",
    "version": 1,
    "correlationId": "my_unique_correlation_id",
    "input": {
        "b": 456,
        "c": 7890    }
}
```

finally I got value from `task.getInputData()` result:
- a: 100
- b: 200
- c: 7890

I don't know why **b value is 200, not 456**, any one meet this problem yet?

apanicker-nflx:
@anxiaoyi I have labeled this as a question for now. Will run some tests and change this to a bug if required.
In the meantime, I will wait for other users to comment on if they have encountered this behavior in their workflows.

rayoy:
> @anxiaoyi I have labeled this as a question for now. Will run some tests and change this to a bug if required.
> In the meantime, I will wait for other users to comment on if they have encountered this behavior in their workflows.

@apanicker-nflx 
I got the same problem, and I didn't find the code for the workflow's _input_ to override the task's _inputParameters_ value.

https://github.com/Netflix/conductor/blob/27ab532d2f625492a07aff39130e28790e871843/core/src/main/java/com/netflix/conductor/core/execution/ParametersUtils.java#L73

apanicker-nflx:
@rayoy Thanks for flagging this. I have re-labeled this to a bug. 
Please feel free to submit a PR with the fix if you would want a fix immediately. We will try to fix this issue in an upcoming release.

[...]

PR #1418, opened 2019-11-24
fix for using inputTemplate bug (#1371).

rayoy:
Fix the bug #1371 addording to the [using inputTemplate](https://netflix.github.io/conductor/configuration/taskdef/#using-inputtemplate) document.


rayoy:
> @rayoy Thank you very much for the PR.
> As described in the issue #1371 , can you also add a test which evaluates the partial substitution of the input params map (the case where some of the values are provided in the template, while some values are not).

@apanicker-nflx done.

rayoy:
I've optimized the test case. Do you have time to review it? @apanicker-nflx

PR #1430, opened 2019-12-05
add owner email in definitions; workflow level timeouts

apanicker-nflx:
- set owner email in workflow and task def
- added workflow level timeouts


demichej:
Does this *require* an ownerEmail now on a task?

apanicker-nflx:
> Does this _require_ an ownerEmail now on a task?

@demichej Yes, we plan to enforce this as a requirement for all definitions moving forward. The idea is to aid platform teams operating Conductor to easily automate alert creation and provide tracking/observability assistance.

Issue #1424, opened 2019-11-28
Communicate with Redis through SSL/TLS


DemisB:
Hi,

First of all, thanks for open-sourcing conductor, it is a great tool!

I am trying to setup a conductor server that works with a Redis persistence model. My redis instance runs on a different server and can only be accessed through SSL/TLS encryption (to be honest, my knowledge of encryption using SSL and/or TLS is quite vague). I think it is the reason why conductor seems to be unable to communicate with my redis instance, as I read this in the conductor output (cf. the first `[ERROR]`, followed by `Unexpected end of stream`):
```
Using config file: settings.properties
[INFO ] 2019-11-28 16:33:49,098 ModulesProvider - Starting conductor server using dynomite/redis cluster.
[WARN ] 2019-11-28 16:33:49,101 ModulesProvider - Starting locking module using Noop Lock.
[INFO ] 2019-11-28 16:33:49,103 ModulesProvider - External payload storage is not configured, provided: , supported values are: [S3]
java.lang.IllegalArgumentException: No enum constant com.netflix.conductor.bootstrap.ModulesProvider.ExternalPayloadStorageType.
	at java.lang.Enum.valueOf(Enum.java:238)
	at com.netflix.conductor.bootstrap.ModulesProvider$ExternalPayloadStorageType.valueOf(ModulesProvider.java:48)
	at com.netflix.conductor.bootstrap.ModulesProvider.selectModulesToLoad(ModulesProvider.java:151)
	at com.netflix.conductor.bootstrap.ModulesProvider.get(ModulesProvider.java:59)
	at com.netflix.conductor.bootstrap.Main.main(Main.java:50)
[INFO ] 2019-11-28 16:33:49,158 SystemTaskWorkerCoordinator - Adding the queue for system task: HTTP
[INFO ] 2019-11-28 16:33:49,266 HttpTask - HttpTask initialized...
[INFO ] 2019-11-28 16:33:49,268 SystemTaskWorkerCoordinator - Adding the queue for system task: KAFKA_PUBLISH
[INFO ] 2019-11-28 16:33:49,268 KafkaPublishTask - KafkaTask initialized...
[INFO ] 2019-11-28 16:33:49,268 SystemTaskWorkerCoordinator - Adding the queue for system task: JSON_JQ_TRANSFORM
[INFO ] 2019-11-28 16:33:49,871 ConfigurationInjectingListener - Initializing ConfigurationInjectingListener
[...]

PR #1441, opened 2019-12-10
Database persistence: connection to a redis host with SSL/TLS encryption can be opted-in

DemisB:
Hi,

So following this issue https://github.com/Netflix/conductor/issues/1424 (tldr; I needed conductor to be able to connect to a redis host with TLS encryption enabled), here is a small pull request that enables encrypted communication between conductor and the configured redis host. The encryption can be opted-in by setting `workflow.dynomite.cluster.hosts.useSSL` to `true` in `settings.properties`. 

What I did is setting up an SSL context using platform default security providers and random generator (cf. [Java documentation](https://docs.oracle.com/javase/7/docs/api/javax/net/ssl/SSLContext.html#init(javax.net.ssl.KeyManager[],%20javax.net.ssl.TrustManager[],%20java.security.SecureRandom))). As I said in my issue message, I am not an expert in SSL/TLS encryption, so I would be happy to get feedbacks about that, to know if that is the recommended/safest choice or not.

The `conductor-postgres-persistence:test` tasks fails on my machine but it does not seem to fail on the latest travis build, and I do not think to have touched anything related to it (but I may be mistaken), however the `conductor-redis-persistence:test` runs fine.

Thank you for reviewing this :)

### DIFF
--- a/redis-persistence/src/main/java/com/netflix/conductor/jedis/DynomiteJedisProvider.java
+++ b/redis-persistence/src/main/java/com/netflix/conductor/jedis/DynomiteJedisProvider.java
@@ -19,13 +19,22 @@ import com.netflix.dyno.connectionpool.impl.ConnectionPoolConfigurationImpl;
 import com.netflix.dyno.jedis.DynoJedisClient;
 import javax.inject.Inject;
 import javax.inject.Provider;
+import javax.net.ssl.SSLContext;
+import java.security.NoSuchAlgorithmException;
+import java.security.KeyManagementException;
 import redis.clients.jedis.commands.JedisCommands;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 
 public class DynomiteJedisProvider implements Provider<JedisCommands> {
 
     private final HostSupplier hostSupplier;
     private final TokenMapSupplier tokenMapSupplier;
     private final DynomiteConfiguration configuration;
+
+    private static Logger logger = LoggerFactory.getLogger(DynomiteJedisProvider.class);
 
     @Inject
     public DynomiteJedisProvider(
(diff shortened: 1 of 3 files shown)


redis-persistence/src/main/java/com/netflix/conductor/jedis/DynomiteJedisProvider.java
```diff
@@ -51,11 +67,23 @@ public JedisCommands get() {
...
                 .withApplicationName(configuration.getAppId())
                 .withDynomiteClusterName(configuration.getClusterName())
-                .withCPConfig(connectionPoolConfiguration)
-                .build();
+                .withCPConfig(connectionPoolConfiguration);
+
+        if (this.configuration.shouldUseSSL()) {
+                try {
+                        final SSLContext sslContext = createAndInitSSLContext();
+                        builder.withSSLSocketFactory(sslContext.getSocketFactory());
```
kishorebanala: To make it extensible, please inject `SSLSocketFactory` through a Provider here.

redis-persistence/src/main/java/com/netflix/conductor/jedis/DynomiteJedisProvider.java
```diff
@@ -38,6 +47,13 @@ public DynomiteJedisProvider(
         this.tokenMapSupplier = tokenMapSupplier;
     }
 
+    private SSLContext createAndInitSSLContext() throws NoSuchAlgorithmException, KeyManagementException {
```
kishorebanala: This should be part of get() method of class that implements `Provider<SSLSocketFactory>`

kishorebanala:
Hey @DemisB, are you still planning to work on this feature?

DemisB:
Hello @kishorebanala, sorry for leaving this PR inactive for so long, I have been away without a computer for pretty much the whole year 2020. I still plan on working on this feature, hopefully within the upcoming weeks!

PR #1447, opened 2019-12-13
Update package-json.lock and revert LeftMenu section updating

demichej:
There are 2 changes happening here.

1) the package-json is currently broken out of the box. This updates it so that npm install and gulp watch work from the ui directory now.

2) This reverts a **part** of this commit: https://github.com/Netflix/conductor/commit/2cf4f0413f452e4f9549726a3571c336bec83ee5#diff-03e4bf1f523c6bb8e97e4eef067cce3f

The UI was no longer updating the data when you clicked a LeftMenu item. This changes it back so that it does fetch data on clicking those LeftMenu items.

Pinging @mstier-nflx because he was the author of the above commit.


demichej:
Going to open a different one. My commits are a mess.

PR #1449, opened 2019-12-14
fixed dag evaluation for cases when a decision task is within a fork join

apanicker-nflx:
- Fixed existing test cases
- Added new tests


demichej:
Very excited for this one :) Thanks @apanicker-nflx

PR #1450, opened 2019-12-15
Upgrade Mockito to v3

ParkerM:
This upgrades the Mockito test library from 1.10.9 to 3.1.0.

The version bump required modifications to many test suites, _mostly_ due package imports.  

Some tests were heavily reliant on the implicit null-matching strategies of Mockito v1, which required scattered changes to assertions and usage of argument matchers.  
I attempted to logically group these changes by commit for reviewability's sake, but can squash them if needed.


common/src/test/java/com/netflix/conductor/common/tasks/TaskTest.java
```diff
@@ -78,13 +78,15 @@ public void testTaskDefinitionIfAvailable() {
     @Test
     public void testTaskQueueWaitTime() {
         Task task = new Task();
-        task.setScheduledTime(System.currentTimeMillis() - 30_000); // 30 seconds ago
-        task.setStartTime(System.currentTimeMillis() - 25_000);
```
ParkerM: The assertion at line 85 fails if `System.currentTimeMillis()` does not return the same value on lines 81 and 82.

ParkerM:
I tacked on a small test change to fix the intermittent failure from my previous build (see the review notes above).

kishorebanala:
@ParkerM Thanks for your contribution, much appreciate it!

client/src/test/java/com/netflix/conductor/client/task/WorkflowTaskCoordinatorTests.java
```diff
@@ -95,8 +97,8 @@ public void testTaskException() {
 				.withTaskClient(client)
 				.withWorkerNamePrefix("test-worker-")
 				.build();
-		when(client.batchPollTasksInDomain(anyString(), anyString(), anyString(), anyInt(), anyInt())).thenReturn(ImmutableList.of(new Task()));
-		when(client.ack(anyString(), anyString())).thenReturn(true);
+		when(client.batchPollTasksInDomain(anyString(), isNull(), anyString(), anyInt(), anyInt())).thenReturn(ImmutableList.of(new Task()));
```
apanicker-nflx: Why is the second argument `domain` changed to isNull?

ParkerM: The behavior of `anyString()` changed so that it no longer matches `null`, and `null` is the actual value that was causing the stub to match in this case.
That conclusion was particularly difficult to arrive at, so while `any()` (which also matches `null`) would work, I instead went with `isNull()` to self-document and prevent the same confusion going forward.

apanicker-nflx: Sounds good, thanks for the explanation. Please rebase against dev and resolve conflicts and we are good to merge.

ParkerM: Rebase has been pushed 👍

PR #1454, opened 2019-12-18
Updating Http task to use injected ObjectMapper with global defaults …

kishorebanala:
…instead of custom ObjectMapper.


anupam587:
So how to create a new http task with new configuration? Can you give me an example? Since with previous configuration my http tasks are always in running state.

kishorebanala:
@anupam587 This PR only changes the Jackson ObjectMapper and not Http task's configuration. Please open a Github issue with more details if you have any questions about Http task usage.

PR #1457, opened 2019-12-18
Redis cluster fix and optimizations

mdepak:
1) Redis cluster client fix
JeidsCluser is broken in the newer versions of the conductor. `JedisCluster` client can be used instead of `Jedis` pool in case of Redis cluster.

2) Initialize index on server mode
While starting the conductor using war, the server modules do not initialize necessary index creation. To keep it consistent with process mode, `setup()` can be done in the IndexDAO instantiation. Further, `workflow.elasticsearch.index.shard.count` and `workflow.elasticsearch.index.replicas.count` properties are added to configure basic settings of elasticsearch index.

3) Optional Event message indexing
In the SimpleEventProcessor's handle() method, all the event messages are indexed in elasticsearch. For large event loads, to achieve minimal delay in event processing indexing event message becomes a bottleneck because of the i/0 involved in elastic search. Thus, the event queue poll is stalled causing the event's wait time in the queue to increase drastically. Since the event messages indexed in elasticsearch are not used anywhere, it can be made optional using a flag `worklfow.event.message.indexing.enabled`.

4) Optional task log indexing
Since task log indexing can take up a considerable amount of space and CPU of elasticsearch, indexing can be optional and controlled using flag `workflow.tasklog.indexing.enabled`.


apanicker-nflx:
@mdepak Thanks for the contribution. 
Please break this up into multiple PRs, one for each issue that you are solving. This would help with isolating and fixing issues in the future.

mdepak:
@apanicker-nflx  Thanks! I will close this PR and create multiple PRs for each issue.

PR #1463, opened 2019-12-26
Redis cluster client fix

mdepak:
JeidsCluser is broken in the current versions of the conductor. `JedisCluster` client can be used instead of Jedis pool in case of Redis cluster persistence.


skyrocknroll:
@apanicker-nflx Can you please review this pull request. We use the Redis cluster.

PR #1467, opened 2019-12-27
Refactoring daos to enable pluggability and separation of concern

apanicker-nflx:


core/src/main/java/com/netflix/conductor/dao/MetadataDAO.java
```diff
@@ -1,41 +1,34 @@
...
  * Data access layer for the workflow metadata - task definitions and workflow definitions
  */
 public interface MetadataDAO {
 
     /**
      * @param taskDef task definition to be created
-     * @return name of the task definition
      */
-    String createTaskDef(TaskDef taskDef);
+    void createTaskDef(TaskDef taskDef);
```
kishorebanala: Instead of void, should it return something confirming that TaskDef is actually created?

core/src/main/java/com/netflix/conductor/dao/MetadataDAO.java
```diff
@@ -62,73 +55,34 @@
     /**
      * @param def workflow definition
      */
-    void create(WorkflowDef def);
+    void createWorkflowDef(WorkflowDef def);
```
kishorebanala: Same as above.

core/src/main/java/com/netflix/conductor/dao/MetadataDAO.java
apanicker-nflx: This is a synchronous request from the API so if the creation fails, the request would receive a non-200 status code.

core/src/main/java/com/netflix/conductor/dao/MetadataDAO.java
apanicker-nflx: same as above, non-200 for failure and 200 for success.

PR #1469, opened 2019-12-29
Use boolean instead of String in TaskService/TaskResource

mactaggart:
This PR replaces using a string representation of a boolean with a boolean in TaskService.ackTaskReceived and TaskResource.ack.


kishorebanala:
@mactaggart Looks great, but we're afraid it would be backwards incompatible. We shall consider doing this in next major release. I'm closing this PR for now. Thanks for your contribution.

PR #1470, opened 2019-12-31
refactor java client to remove client-side queues

apanicker-nflx:
**Current State**
- Tasks are polled in bulk by the client and added to a LinkedBlockingQueue.
- StartTime is set on the task alongwith IN_PROGRESS state by the server upon polling. 
- If ack fails on the task, the task will be polled again (after unack timeout) and added to the client-side queue.

**Observation**
Server calculates timeouts on the task based on the startTime. If the ack for a task fails multiple times due to long-running tasks in the worker, task will be marked as TIMED_OUT without being processed by a worker but the task could still be enqueued within the client-side queue.

**Solution**
- Remove the client-side queues.
- Tasks will be polled and executed based on the number of available threads.
- Inherently, eventhough tasks were polled in bulk, execution was serial which meant that task execution was limited by the number of available threads.
- This PR bubbles up the responsibility to poll and execute a task only if a thread is available, thus ensuring that a task will not be started (put into IN_PROGRESS with a startTime) by the server, only when a thread is available for executing the task after polling.
- This essentially removes the reliance on `ack` since a worker will only poll for a task when it can execute it.
- The `TaskRunnerConfigurer` is the entrypoint to configure the automated polling and execution.
- The `TaskPollExecutor` contains the actual business logic of polling and executing the tasks based on the number of available threads in the configured threadpool.

**NOTE**
- Removed deprecated method in TaskClient 
```java
void updateTask(TaskResult, String)
```


kishorebanala:
Lookout for build errors.

client/src/main/java/com/netflix/conductor/client/task/WorkflowTaskCoordinator.java
```diff
@@ -23,426 +23,420 @@
...
  * Manages the Task workers thread pool and server communication (poll, task update and acknowledgement).
- *
- * @author Viren
  */
 public class WorkflowTaskCoordinator {
 
-	private static final Logger logger = LoggerFactory.getLogger(WorkflowTaskCoordinator.class);
-
-	private TaskClient taskClient;
+    private static final Logger LOGGER = LoggerFactory.getLogger(WorkflowTaskCoordinator.class);
```
pctreddy: There is a lot happening in this class, like WorkflowTaskExecution setup and the orchestration of polling for new messages and processing them. Think to break this into smaller composable classes. It would reduce complexity of the class.

client/src/main/java/com/netflix/conductor/client/task/WorkflowTaskCoordinator.java
```diff
@@ -23,426 +23,420 @@
...
+                .namingPattern(workerNamePrefix + count.getAndIncrement())
+                .uncaughtExceptionHandler(uncaughtExceptionHandler)
+                .build());
+
+        this.scheduledExecutorService = Executors.newScheduledThreadPool(workers.size());
+        workers.forEach(worker -> scheduledExecutorService.scheduleWithFixedDelay(() -> pollForTask(worker),
+            worker.getPollingInterval(), worker.getPollingInterval(), TimeUnit.MILLISECONDS));
+    }
+
+    public void shutdown() {
```
pctreddy: Javadoc on when and how it is invoked or should be invoked.

client/src/main/java/com/netflix/conductor/client/task/WorkflowTaskCoordinator.java
```diff
@@ -23,426 +23,420 @@
...
+        } finally {
+            pollingSemaphore.complete();
+        }
+        return task;
+    }
+
+    private void execute(Worker worker, Task task) {
+        String taskType = task.getTaskDefName();
+        try {
+            if (!worker.preAck(task)) {
```
pctreddy: Should this be still performed ? There should be no `ACK`  semantics since the task is in progress once polled.

client/src/main/java/com/netflix/conductor/client/task/WorkflowTaskCoordinator.java
```diff
@@ -23,426 +23,420 @@
...
-		logger.debug("Polling {}, domain={}, count = {} timeout = {} ms", worker.getTaskDefName(), domain, worker.getPollCount(), worker.getLongPollTimeoutInMS());
-
-		List<Task> tasks = Collections.emptyList();
-		try{
-            // get the remaining capacity of worker queue to prevent queue full exception
-            int realPollCount = Math.min(workerQueue.remainingCapacity(), worker.getPollCount());
-            if (realPollCount <= 0) {
-				logger.warn("All workers are busy, not polling. queue size = {}, max = {}", workerQueue.size(), workerQueueSize);
-				return;
+            if (!taskClient.ack(task.getTaskId(), worker.getIdentity())) {
```
pctreddy: Same as above.

client/src/main/java/com/netflix/conductor/client/task/WorkflowTaskCoordinator.java
```diff
@@ -23,426 +23,420 @@
...
+        }
+
+        /**
+         * Builds an instance of the WorkflowTaskCoordinator.
+         * <p>
+         * Please see {@link WorkflowTaskCoordinator#init()} method. The method must be called after this constructor
+         * for the polling to start.
+         * </p>
+         */
+        public WorkflowTaskCoordinator build() {
```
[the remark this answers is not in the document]
apanicker-nflx: fixed

client/src/main/java/com/netflix/conductor/client/task/WorkflowTaskCoordinator.java
```diff
@@ -23,426 +23,420 @@
...
+     * @param updateRetryCount number of times to retry the failed updateTask operation
+     * @param taskWorkers      workers that will be used for polling work and task execution.
+     * @param workerNamePrefix String prefix that will be used for all the workers.
+     *                         <p>
+     *                         Please see {@link #init()} method. The method must be called after this constructor for
+     *                         the polling to start.
+     *                         </p>
+     * @see Builder
+     */
+    public WorkflowTaskCoordinator(EurekaClient eurekaClient, TaskClient taskClient, int threadCount,
```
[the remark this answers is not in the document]
apanicker-nflx: Deprecated this class, made the `TaskRunnerConfigurer` constructor private and can only be initialized using the builder.

client/src/main/java/com/netflix/conductor/client/task/WorkflowTaskCoordinator.java
apanicker-nflx: Refactored into `TaskRunnerConfigurer` and `TaskPollExecutor`

client/src/main/java/com/netflix/conductor/client/task/WorkflowTaskCoordinator.java
apanicker-nflx: added javadoc

client/src/main/java/com/netflix/conductor/client/task/WorkflowTaskCoordinator.java
apanicker-nflx: `TaskPollExecutor` does not use ack anymore

client/src/main/java/com/netflix/conductor/client/automator/PollingSemaphore.java
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Signals if polling is allowed based on whether a permit can be acquired.
+     *
+     * @return {@code true} - if permit is acquired
+     *         {@code false} - if permit could not be acquired
+     */
+    boolean canPoll() {
+        boolean acquired = semaphore.tryAcquire();
+        LOGGER.debug("Trying to acquire permit: {}", acquired);
```
kishorebanala: This log itself may not be helpful much. Should we use a metric instead?

client/src/main/java/com/netflix/conductor/client/automator/TaskPollExecutor.java
```diff
@@ -0,0 +1,246 @@
...
+        if (worker.paused()) {
+            MetricsContainer.incrementTaskPausedCount(worker.getTaskDefName());
+            LOGGER.debug("Worker {} has been paused. Not polling anymore!", worker.getClass());
+            return;
+        }
+
+        Task task;
+        try {
+            if (!pollingSemaphore.canPoll()) {
+                return;
```
kishorebanala: A metric here to see how often we cannot poll?

client/src/main/java/com/netflix/conductor/client/automator/TaskPollExecutor.java
```diff
@@ -0,0 +1,246 @@
...
+        LOGGER.debug("Executing task: {} of type: {} in worker: {} at {}", task.getTaskId(), task.getTaskDefName(),
+            worker.getClass().getSimpleName(), worker.getIdentity());
+        try {
+            executeTask(worker, task);
+        } catch (Throwable t) {
+            task.setStatus(Task.Status.FAILED);
+            TaskResult result = new TaskResult(task);
+            handleException(t, result, worker, task);
+        } finally {
+            pollingSemaphore.complete();
```
kishorebanala: Apart from this finally block, I don't see `processTask` adding any value to `executeTask`. Why not just use `executeTask` and get rid of this?

client/src/main/java/com/netflix/conductor/client/http/TaskClient.java
```diff
@@ -19,7 +19,7 @@
 import com.netflix.conductor.client.config.ConductorClientConfiguration;
 import com.netflix.conductor.client.config.DefaultConductorClientConfiguration;
 import com.netflix.conductor.client.exceptions.ConductorClientException;
-import com.netflix.conductor.client.task.WorkflowTaskMetrics;
+import com.netflix.conductor.client.telemetry.MetricsContainer;
```
kishorebanala: +1

client/src/main/java/com/netflix/conductor/client/task/WorkflowTaskCoordinator.java
```diff
@@ -42,405 +43,405 @@
...
+import static com.netflix.conductor.client.telemetry.MetricsContainer.incrementTaskPollCount;
 
 /**
- * Manages the Task workers thread pool and server communication (poll, task update and acknowledgement).
- *
- * @author Viren
+ * This class has been deprecated and will be removed in a future release.
+ * Please use {@link com.netflix.conductor.client.automator.TaskRunnerConfigurer} instead.
  */
+@Deprecated
```
kishorebanala: Consider adding Deprecated since version number.

client/src/main/java/com/netflix/conductor/client/automator/PollingSemaphore.java
apanicker-nflx: This is a debug level log which is intended to be used only for debugging issues, a metric here would not give any valuable info.

client/src/main/java/com/netflix/conductor/client/automator/TaskPollExecutor.java
apanicker-nflx: Same as above, since when workers are processing tasks, the poller will not be active so metric would not give any valuable info here.

client/src/main/java/com/netflix/conductor/client/automator/TaskPollExecutor.java
apanicker-nflx: `executeTask` calls execute on the worker and handles uncaught exceptions. Then it tries to  `updateTask` with the server.
`processTask` is only responsible for handling exceptions during this update. Combining different exception cases based on the flow within the same method would be confusing.

Issue #1478, opened 2020-01-07
Lambda after loop - stuck execution


Valarin:
Execution of this WF never finishes

`{
  "name": "stuck_execution",
  "description": "Lambda after loop - stuck execution",
  "version": 1,
  "schemaVersion": 2,
  "tasks": [
    {
      "name": "LoopTask",
      "taskReferenceName": "LoopTask",
      "type": "DO_WHILE",
      "loopCondition": "($.LoopTask['iteration'] < 2)",
      "loopOver": [
        {
          "name": "LAMBDA_TASK",
          "taskReferenceName": "result_lambda_ref",
          "type": "LAMBDA",
          "inputParameters": {
            "scriptExpression": "return { status: 'STH' };"
          }
        }
      ],
      "startDelay": 0,
      "optional": false
    },
    {
      "name": "LAMBDA_TASK",
      "taskReferenceName": "result_lambda_ref",
      "type": "LAMBDA",
      "inputParameters": {
        "scriptExpression": "return { status: 'Executed' };"
      }
    }
  ]
}
`

![Screenshot from 2020-01-07 10-47-40](https://user-images.githubusercontent.com/11278978/71885737-3c234f80-313b-11ea-8bac-af14b1b2457a.png)

Lambda task that is supposed to run after loop never gets executed and is somehow partially treated as if it was inside the loop.

manan164:
Hi @Valarin 
Thanks for pointing out. I am working on a fix. Basically it is treating next task as loopOver task. 
I will raise PR in some time.

manan164:
Hi @Valarin ,
I have created a PR https://github.com/Netflix/conductor/pull/1479
Please pull this branch till the time.
Also ideally we should not use the loopover task outside the scope of DO_WHILE task as it can hamper the Loop task operation.

PR #1479, opened 2020-01-07
Loop task fix.

manan164:
Fix for [Loop -stuck](https://github.com/Netflix/conductor/issues/1478)

Things fixed in this PR.
1. We have to check that the next task we are scheduling also need to be loop over task then only change the task reference name. 
2. In the case of two sequential do while task, the next task returned will contain second do while task. In this case, also we don't need to do anything as second do while the task will take care.
3. In case we have two or more DO_WHILE tasks in the workflow, It should wait for only the corresponding loopOver task to get completed. Earlier it was unoptimized although the functionality was correct. 
4. Forming of DO_WHILE task outputMap. Earlier changed task Reference Name was being used as key. Now preserving original task Reference Name.
5. Updated documentation.
6. Integration test regarding this functionality.


common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowDef.java
```diff
@@ -241,6 +241,9 @@ public WorkflowTask getNextTask(String taskReferenceName){
 			 WorkflowTask nextTask = task.next(taskReferenceName, null);
 			 if(nextTask != null){
 				 return nextTask;
+			 } else if (TaskType.DO_WHILE.name().equals(task.getType()) && !task.getTaskReferenceName().equals(taskReferenceName) && task.has(taskReferenceName)) {
```
manan164: @apanicker-nflx , We need to return null in case all loopOver task execution completed. The issue was, it was returning the next task which was LAMBDA as per [definition](https://github.com/Netflix/conductor/issues/1478#issue-546183053). Since we are returning null it won't change task reference name. LAMBDA task will be return when DO_WHILE task will get completed. And we have added safe guard not to change taskReferenceName in such case.

core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
```diff
@@ -190,7 +190,7 @@ private DeciderOutcome decide(final Workflow workflow, List<Task> preScheduledTa
             if (!pendingTask.isExecuted() && !pendingTask.isRetried() && pendingTask.getStatus().isTerminal()) {
                 pendingTask.setExecuted(true);
                 List<Task> nextTasks = getNextTask(workflow, pendingTask);
-                if (pendingTask.isLoopOverTask() && !nextTasks.isEmpty()) {
+                if (pendingTask.isLoopOverTask() && !TaskType.DO_WHILE.name().equals(pendingTask.getTaskType()) && !nextTasks.isEmpty()) {
```
manan164: Since we have changed the way next task will be returned for last loopOver task we just need to check that in case of two sequential DO_WHILE task we should not change the task reference name of second DO_WHILE task when the first one gets completed.

core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java
```diff
@@ -65,15 +66,15 @@ public boolean execute(Workflow workflow, Task task, WorkflowExecutor workflowEx
 		StringBuilder failureReason = new StringBuilder();
 		Map<String, Object> output = new HashMap<>();
 		task.getOutputData().put("iteration", task.getIteration());
-		List<Task> loopOver = workflow.getTasks().stream().filter(t -> (t.getReferenceTaskName().endsWith(TaskUtils.getLoopOverTaskRefNameSuffix(task.getIteration()))) && t.isLoopOverTask()).collect(Collectors.toList());
+		List<Task> loopOver = workflow.getTasks().stream().filter(t -> (task.getWorkflowTask().has(TaskUtils.removeIterationFromTaskRefName(t.getReferenceTaskName())) && !task.getReferenceTaskName().equals(t.getReferenceTaskName()))).collect(Collectors.toList());
```
manan164: In case of workflow having two DO_WHILE task we should not check for all workflow task which matches the current iteration. We should only wait for all children task to get completed.

core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java
```diff
@@ -65,15 +66,15 @@ public boolean execute(Workflow workflow, Task task, WorkflowExecutor workflowEx
...
 
 		for (Task loopOverTask : loopOver) {
 			Status taskStatus = loopOverTask.getStatus();
 			hasFailures = !taskStatus.isSuccessful();
 			if (hasFailures) {
 				failureReason.append(loopOverTask.getReasonForIncompletion()).append(" ");
 			}
-			output.put(loopOverTask.getReferenceTaskName(), loopOverTask.getOutputData());
+			output.put(TaskUtils.removeIterationFromTaskRefName(loopOverTask.getReferenceTaskName()), loopOverTask.getOutputData());
```
manan164: This is just code refactoring to fill DO_WHILE task output map to preserver original taskReferneceName.

core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java
```diff
@@ -129,10 +130,10 @@ boolean getEvaluatedCondition(Workflow workflow, Task task, WorkflowExecutor wor
 		TaskDef taskDefinition = workflowExecutor.getTaskDefinition(task);
 		Map<String, Object> taskInput = parametersUtils.getTaskInputV2(task.getWorkflowTask().getInputParameters(), workflow, task.getTaskId(), taskDefinition);
 		taskInput.put(task.getReferenceTaskName(), task.getOutputData());
-		List<Task> loopOver = workflow.getTasks().stream().filter(t -> (t.getReferenceTaskName().endsWith(TaskUtils.getLoopOverTaskRefNameSuffix(task.getIteration()))) && t.isLoopOverTask()).collect(Collectors.toList());
+		List<Task> loopOver = workflow.getTasks().stream().filter(t -> (task.getWorkflowTask().has(TaskUtils.removeIterationFromTaskRefName(t.getReferenceTaskName())) && !task.getReferenceTaskName().equals(t.getReferenceTaskName()))).collect(Collectors.toList());
```
manan164: Same as https://github.com/Netflix/conductor/pull/1479/files#r367092996. No need to fill the taskInput Map unnecessarily.

core/src/test/java/com/netflix/conductor/core/execution/tasks/DoWhileTest.java
```diff
@@ -107,6 +107,7 @@ public void setup() {
         loopTask.setIteration(1);
         loopWorkflowTask = new WorkflowTask();
         loopWorkflowTask.setTaskReferenceName("loopTask");
+        loopWorkflowTask.setType(TaskType.DO_WHILE.name());
```
apanicker-nflx: why was the task type removed?

docs/docs/configuration/systask.md
```diff
@@ -569,7 +569,7 @@ Each iteration of loop over task will be scheduled as taskRefname__iteration. It
 Do while task output number of iterations with iteration as key and value as number of iterations. Each iteration's output will be stored as, iteration as key and loopover task's output as value
 Taskname which contains arithmetic operator must not be used in loopCondition. Any of loopOver task can be reference outside do while task same way other tasks are referenced.
 To reference specific iteration's output, ```$.LoopTask['iteration]['first_task']```
-Do while task does NOT support domain or isolation group execution.
+Do while task does NOT support domain or isolation group execution. Nesting of DO_WHILE task is not supported. Loopover task must not be reused in neither workflow nor another DO_WHILE task.
```
apanicker-nflx: > Loopover task must not be reused in neither workflow nor another DO_WHILE task.

Please provide more details about this. Does this mean that if a task is used as a loopover task, it can be used again within the same workflow?

manan164: No. It must not be reused within the same workflow once the task is used as loopOver task.

core/src/test/java/com/netflix/conductor/core/execution/tasks/DoWhileTest.java
manan164: It was never added. The tests were running fine earlier because I was directly using DoWhile class. Now we need because I have changed the way loopOvertask is being calculated. Earlier the approach was find all the task which has taskReferenceName as ```___iteration``` but now approach is the DO_WHILE task must ```has()``` that task.

manan164:
HI, @apanicker-nflx I wrote an integration test for this with two DO_WHILE tasks as consecutive. For the WorkflowLegacyMigration test, I have to add one http task between two do while task because that http task was not being added to workflow so seconds task was getting completed. But that for WorkflowServiceTest they are consecutive.

manan164:
Hi @apanicker-nflx , Please review.

PR #1502, opened 2020-01-22
Update DECISION task documentation

Jiehong:
In particular, add the type of each parameter for a DECISION, and document its output.


manan164:
Hi @Jiehong ,
I would also like to add that task names with arithmetic operators should not be used in caseExpression. This has created lot of confusion while I was working on DO_WHILE task.
I have added that [here](https://github.com/Netflix/conductor/blob/master/docs/docs/configuration/systask.md#do-while-task)

PR #1507, opened 2020-01-23
Small updated Task.java

oanaema:
Added missing fields on copy method from Task model.


apanicker-nflx:
@oanaema Please provide a description as to why do you think these fields need to be part of the copy method in the Task object?
Currently, the main purpose of the `copy` method is to help copy over required fields during retry operation during which the startTime, scheduledTime, endTime should not be copied over.

oanaema:
Hello,
This PR was created to fix the issue we have on the Task level with some of the fields that are missing from copy method. We should be able to get scheduledTime, startTime and endTime parameters from WorkflowTasks in a workflow. Instead of getting such dates, we are always getting 0 as a value, because in the copy method multiple fields are reset.
For instance I saw that seq and startTime fields need to be reset because they are taken into account in some business logic.
Can you please advice what we should do in this case, since we cannot touch startTime because we are breaking other logic?

Issue #1500, opened 2020-01-17
Readme page with indication for external tools


flaviostutz:
I've created some tools based on Netflix Conductor to help on CRON like tasks or Backup management jobs.

* [Schellar](https://github.com/flaviostutz/schellar) is a scheduler tool for instantiating Conductor workflows from time to time
* [Backtor](https://github.com/flaviostutz/backtor) is a backup scheduler tool that uses Conductor workers to handle backup operations)

Do you think it is worth we create a link on the main README page to a page with "external" contribution" links? I've found it a bit hard to find this kind of tools over the Internet.

kishorebanala:
@flaviostutz These look awesome. We'll definitely consider adding pointers in Conductor documentation. Thanks for your contributions.

flaviostutz:
Great!
Please check PR https://github.com/Netflix/conductor/pull/1508

PR #1508, opened 2020-01-23
Adding RELATED page with some community projects related to Conductor

flaviostutz:
The objective of this page is to collect some Conductor related projects that could help people using Conductor in real operations.


flaviostutz:
Closes #1500

PR #1513, opened 2020-01-26
update doc

manan164:
HI @apanicker-nflx , I would like to update documentation of DYNAMIC_FORK_JOIN task.
We recently had a use case where the dynamic_fork_join task was not workflow's first task. We used LAMBDA task to form a json using previous task's otuput. We have passed that to DYNAMIC_FORK_JOIN task. I can share the LAMBDA task definition also required.


apanicker-nflx:
@manan164 There is no restriction on dynamic fork join such that it needs to be the first task of a workflow. We have several cases of workflow definitions where our users use the dynamic fork join at multiple different places within the workflow.
Please do share the LAMBDA task definition as that may be really helpful to other users if you feel that it is generic enough to be re-usable. Thanks for your contribution.

manan164:
Hi, @apanicker-nflx I feel it is not too generic since we have to build the JSON according to tasks and their respective inputs. Closing this PR.

PR #1516, opened 2020-01-30
Fixed deep copy of task model.

oanaema:
This PR provides a quick fix for deep copy method of Workflow. Copy method of Workflow is using the copy method from the Task model in a wrong way, since currently, the main purpose of the copy method is to help copy over required fields during retry operation.


apanicker-nflx:
@oanaema Thanks for the contribution. Please rebase and squash commits into single commit.

oanaema:
@apanicker-nflx  thank you too.
Done the rebase and squash.

PR #1519, opened 2020-01-31
Update ElasticSearch when a task is CANCELED

rickfish:
When a workflow is TERMINATED, all tasks are added to ElasticSearch in their current state, even if they are in SCHEDULED state because they have not yet been polled. After that, SCHEDULED tasks have their status changed to CANCELED and the database is updated. But ElasticSearch is not updated (indexDAO.indexTask() is never called), so the database is out of sync with ElasticSearch. This change will index the task after the database is updated so that both are in sync.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
@@ -26,6 +26,23 @@ import static com.netflix.conductor.core.execution.ApplicationException.Code.INV
 import static com.netflix.conductor.core.execution.ApplicationException.Code.NOT_FOUND;
 import static com.netflix.conductor.core.execution.tasks.SubWorkflow.SUB_WORKFLOW_ID;
 
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+import javax.inject.Inject;
+
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.netflix.conductor.annotations.Trace;
@@ -52,24 +69,11 @@ import com.netflix.conductor.core.metadata.MetadataMapperService;
 import com.netflix.conductor.core.orchestration.ExecutionDAOFacade;
 import com.netflix.conductor.core.utils.IDGenerator;
 import com.netflix.conductor.core.utils.QueueUtils;
+import com.netflix.conductor.dao.IndexDAO;
 import com.netflix.conductor.dao.MetadataDAO;
 import com.netflix.conductor.dao.QueueDAO;
 import com.netflix.conductor.metrics.Monitors;
 import com.netflix.conductor.service.ExecutionLockService;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.function.Predicate;
-import java.util.stream.Collectors;
-import javax.inject.Inject;
-import org.apache.commons.lang3.StringUtils;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * @author Viren Workflow services provider interface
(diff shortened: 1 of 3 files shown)


apanicker-nflx:
@rickfish Thank you very much for the contribution.
The task document is indexed in Elasticsearch only upon the completion of the workflow - https://github.com/Netflix/conductor/blob/master/core/src/main/java/com/netflix/conductor/core/orchestration/ExecutionDAOFacade.java#L218. So the task document would be created with CANCELED status in ES once the workflow reaches a terminal state. 

> When a workflow is TERMINATED, all tasks are added to ElasticSearch in their current state, even if they are in SCHEDULED state because they have not yet been polled. After that, SCHEDULED tasks have their status changed to CANCELED and the database is updated.

This seems odd, when a workflow is TERMINATED, all tasks within the workflow should also be in a terminal state. All the SCHEDULED state tasks should be marked as CANCELED at this stage.

rickfish:
> This seems odd, when a workflow is TERMINATED, all tasks within the workflow should also be in a terminal state. All the SCHEDULED state tasks should be marked as CANCELED at this stage.

They *are* CANCELED but only in the database.

The problem lies in WorkflowExecutor.terminateWorkflow (https://github.com/Netflix/conductor/blob/bd32fe164020781468b70f44e7dc76dc24714d49/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java#L717). 
 
The use case I used to reproduce this is starting a workflow that has a fork that has on both sides a SIMPLE task where each is SCHEDULED (so not yet polled). Then terminate the workflow using DELETE /workflow/<workflowId>. The terminateWorkflow() method referenced above is then called. You can see that when executionDaoFacade.updateWorkflow() is called in terminateWorkflow() all the SCHEDULED tasks (which are not usually indexed into ES) are indexed, When that call returns, the SCHEDULED tasks are CANCELED. If you follow the link above, you will see that the task status is set to CANCELED and updateTask() is called but there is no call to indexDAO.indexTask(). I'm sure that this is a strange use case that isn't usually executed. One of our users is causing this use case several times a day; that is how I noticed it.

apanicker-nflx:
Thanks for the context. This is a valid bug and is indeed an edge-case. Looking at the linked code, I feel that it would be easier to move ```executionDAOFacade.updateWorkflow(workflow);```  after line 721. This would prevent having to add a dependency to IndexDAO in the WorkflowExecutor. Would need to ensure that the workflow object is updated with the updated tasks (new status).

Also, we are looking into indexing task documents during updateTask to enable task level searching for in progress workflows. Once this is implemented, it would solve for this issue as well. Thoughts?

rickfish:
To me it looks like either of your solutions is better than doing a one-off index of the task when needed as I did but I admit that I don't know the reasoning behind not putting the indexDAO.indexTask() call in updateTask() in the first place. Can you explain? I'm sure there is a reason and I don't want to start changing code that is integral to the functioning of Conductor unless I understand all the implications.

The problem with not calling indexDAO.indexTask() during the updateTask() method from my perspective is the /tasks/search endpoint. If you use that endpoint and your freeText is, for example, status:CANCELED then you get inaccurate results because that endpoint does not find the CANCELED tasks as ElasticSearch is used to do the initial query. I think the same issue surfaces for other statuses but I haven't yet looked into it further.

rickfish:
I think the reason that updateWorkflow() is called before the updating of the tasks (during terminateWorkflow()) is because we don't want to have a task update exception thwart the updating of the workflow. Please correct me if you disagree. I therefore like your second solution of adding a call to indexDAO.indexTask() in the updateTask() method. That will create a ton more volume for ElasticSearch but I would rather have both ElasticSearch and the database in sync with more volume than worry about volume.

The question is: are there situations where adding something to ElasticSearch (where it currently is not added) would cause a problem for another use case? I don't know the answer to that...

apanicker-nflx:
> Are there situations where adding something to ElasticSearch (where it currently is not added) would cause a problem for another use case?

Your observation is correct that this will create a large volume of requests to ES and in case of synchronous indexing, this can introduce a latency cost. However, I agree that it is better to have both Elasticsearch and primary datastore in sync in such situations. 

>  I therefore like your second solution of adding a call to indexDAO.indexTask() in the updateTask() method.

This change can be added such that it is performed only in cases where async indexing is not being used. Otherwise with the current in-memory queue async indexing model, there is a high possibility of indexing requests being dropped.

rickfish:
> This change can be added such that it is performed only in cases where async indexing is not being used. Otherwise with the current in-memory queue async indexing model, there is a high possibility of indexing requests being dropped.

So, if I add this code to ExecutionDAOFacade.updateTask(Task task):

`
if(!config.enableAsyncIndexing()) {
     indexDAO.indexTask(task);
}
`
then I would need to bracket every existing call to indexDAO.indexTask() with this:

`
if(config.enableAsyncIndexing()) {
     indexDAO.indexTask(task);
}
`
Is that acceptable to you? It could get messy I am thinking.

rickfish:
I am going to make the changes to index all task updates if async indexing is not enabled but I will put it in another pull request as I created a different branch (index-all-task-updates). Sorry if this is a problem. The new pull request is #1526. I will close this one.

techyragu:
@rickfish - Is this issue fixed? I don't see relevant code changes in the new pull request is #1526.

rickfish:
@techyragu sorry, the pull request was #1546, not 1526.

PR #1520, opened 2020-01-31
Create Correlation id index on workflow table

rickfish:
Currently queries on the workflow table by correlation_id are slow in large databases. This PR adds an index to the workflow table to fix this for both Postgres and MySql persistence modules


kishorebanala:
@mactaggart @jvemugunta @gorzell Can you help review this please?

mactaggart:
Looks good

demichej:
We just hit this as well. Thanks for fixing it.

Issue #1501, opened 2020-01-21
/tasks/{taskId}/ack always return false


balloonio:
I'm following the add-idents developer lab [here](https://netflix.github.io/conductor/labs/beginner/#ack-the-task) and I don't seem to get the ack task step working.

**Question 1** - call to `/tasks/{taskId}/ack` with correct task id always returns `false`. Why?
**Question 2** - (probably because I always get `false` as the response) I don't see any changes or effect on Workflow UI being reflected. Is that expected?
**Question 3** - What does this ack step do exactly? I ended up just skipping this step entirely and the workflow worked fine. The task just changes from SCHEDULED to IN_PROGRESS after polling, and from IN_PROGRESS to COMPLETED after the mocked worker return response. Therefore, I'm confused if the ack call in the middle does anything at all.

For example, here is my response from task polling in [this step](https://netflix.github.io/conductor/labs/beginner/#poll-for-worker-task)
```
{
    "taskType": "add_idents",
    "status": "IN_PROGRESS",
    "inputData": {
        "identType": "animation",
        "contentId": "my_unique_content_id"
    },
    "referenceTaskName": "add_idents_by_type",
    "retryCount": 0,
    "seq": 3,
    "pollCount": 1,
    "taskDefName": "add_idents",
    "scheduledTime": 1579564011353,
    "startTime": 1579564721830,
    "endTime": 0,
    "updateTime": 1579564721830,
    "startDelayInSeconds": 0,
    "retried": false,
    "executed": false,
    "callbackFromWorker": true,
    "responseTimeoutSeconds": 180,
    "workflowInstanceId": "eaa3f523-9f0d-4426-b3f8-36d45a5d78cf",
    "workflowType": "add_netflix_identation",
    "taskId": "8452fcf4-f016-4da3-96a2-ea797068af4e",
    "callbackAfterSeconds": 0,
    "outputData": {},
    "workflowTask": {
        "name": "add_idents",
        "taskReferenceName": "add_idents_by_type",
        "inputParameters": {
            "identType": "${workflow.input.identType}",
            "contentId": "${workflow.input.contentId}"
        },
        "type": "SIMPLE",
        "decisionCases": {},
        "defaultCase": [],
        "forkTasks": [],
        "startDelay": 0,
        "joinOn": [],
        "optional": false,
        "taskDefinition": {
            "createTime": 1579563756480,
            "createdBy": "",
            "name": "add_idents",
            "retryCount": 3,
[...]

PR #1534, opened 2020-02-14
Remove /tasks/{taskId}/ack endpoint

balloonio:
Discussed with @kishorebanala in #1501 , the `/tasks/{taskId}/ack` endpoint is no longer needed and to be removed.

In this PR, I only removed the endpoint on the API layer, as I didn't want to prematurely remove anything from the core. Please let me know if more removal or deprecation should be included in this PR.

### DIFF
--- a/jersey/src/main/java/com/netflix/conductor/server/resources/TaskResource.java
+++ b/jersey/src/main/java/com/netflix/conductor/server/resources/TaskResource.java
@@ -108,15 +108,6 @@ public class TaskResource {
 	}
 
 	@POST
-	@Path("/{taskId}/ack")
-	@ApiOperation("Ack Task is received")
-	@Consumes({MediaType.WILDCARD})
-	public String ack(@PathParam("taskId") String taskId,
-					  @QueryParam("workerid") String workerId) {
-		return taskService.ackTaskReceived(taskId, workerId);
-	}
-
-	@POST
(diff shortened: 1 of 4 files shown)


balloonio:
@kishorebanala  Bumping up the thread

kishorebanala:
Hey @balloonio, Thanks for your contribution. As I was mentioning in #1501, we cannot remove this API in minor version upgrade, as this would break several clients, especially the previous versions which still calls the ack API.

For now, we shall mark this API deprecated, and do the same in client implementation as well.

christianscheuer:
A quick question as a new user to the current state of things with the /ack endpoint.
I'm seeing the following odd behaviors right now:
- /ack endpoint always returns false
- /poll endpoint returns the same task more than once when called concurrently from different worker threads (from C# client for example)
- Python client calls ack and only proceeds if ack returns true, so it never perform any actions
- C# client runs the ack endpoint but always proceeds regardless of the ack result.

So in other words - right now the C# client ends up processing the task more than once and the python client ends up not processing anything at all.

Since the /ack endpoint is deprecated, shouldn't the /poll endpoint always only return the task exactly once?
And shouldn't /ack always return true for backwards compatibility with clients like the python client that will only proceed if it returns true?

kishorebanala:
Closing due to inactivity, please reopen if you are able to resume working on this.

PR #1543, opened 2020-02-23
Replaced Hibernate Validator with Apache Bval

cyzhao:


core/src/main/resources/conductor-validation-mapping.xml
```diff
@@ -0,0 +1,15 @@
...
+<constraint-mappings
+        xmlns="http://jboss.org/xml/ns/javax/validation/mapping"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation=
+                "http://jboss.org/xml/ns/javax/validation/mapping validation-mapping-1.0.xsd">
+    <default-package>com.netflix.conductor.common.metadata.workflow</default-package>
+
+    <bean class="WorkflowTask" ignore-annotations="false">
+        <class ignore-annotations="false">
+            <constraint annotation="com.netflix.conductor.validations.WorkflowTaskTypeConstraint"/>
```
kishorebanala: Is it possible to define these constraints in a class?

cyzhao: Unfortunately, no.

PR #1546, opened 2020-02-24
index all task updates, not just when workflows are in terminal state

rickfish:
All task updates used to be indexed until async indexing was added in release 2.16.0. In order to accommodate async indexing, index requests were cached until it was time to commit their block. Since a system failure would mean that the uncommitted block(s) would be lost, it was determined to only index tasks when the workflow was set to terminal state (COMPLETED, FAILED, TERMINATED, etc).

This causes a problem for the /tasks/search API because not all tasks are returned, especially if they are in progress.

This pull request will index all task updates if async indexing is disabled; otherwise tasks will only be indexed when the workflows are put in terminal state.


core/src/main/java/com/netflix/conductor/core/orchestration/ExecutionDAOFacade.java
```diff
@@ -230,7 +230,11 @@ public String updateWorkflow(Workflow workflow) {
                 workflow.getTasks().forEach(indexDAO::asyncIndexTask);
             } else {
                 indexDAO.indexWorkflow(workflow);
-                workflow.getTasks().forEach(indexDAO::indexTask);
+                /*
```
apanicker-nflx: Can remove this comment section here since the comments below in `updateTask` covers this.

apanicker-nflx:
@rickfish Thank you very much for the PR. One minor comment but otherwise LGTM.

core/src/main/java/com/netflix/conductor/core/orchestration/ExecutionDAOFacade.java
rickfish: @apanicker-nflx, I removed the comments in my repo.

Issue #1553, opened 2020-03-02
Race condition: NPE in org.elasticsearch.action.bulk.BulkRequest.validate()


skissane:
### Reproduce steps

1. Checkout latest conductor master (I tested with `9d44897f37450fb7ad602bd016ff9a5813bee98c`, but same issue occurs with older versions)
2. Run `./gradlew build -x test`
3. `java -jar ./server/build/libs/conductor-server-2.26.0-SNAPSHOT-all.jar`
4. Run three copies of the `conductor-test.py` script (given below) simultaneously

### Results

After a while, the test scripts start to hang and this error happens in log:
```
[ERROR] 2020-03-02 12:21:40,531 RetryUtil - Operation 'indexObject:Bulk Indexing task' failed after retrying 3 times, retry limit 3
java.lang.NullPointerException
	at org.elasticsearch.action.bulk.BulkRequest.validate(BulkRequest.java:604)
	at org.elasticsearch.action.TransportActionNodeProxy.execute(TransportActionNodeProxy.java:46)
	at org.elasticsearch.client.transport.TransportProxyClient.lambda$execute$0(TransportProxyClient.java:59)
	at org.elasticsearch.client.transport.TransportClientNodesService.execute(TransportClientNodesService.java:250)
	at org.elasticsearch.client.transport.TransportProxyClient.execute(TransportProxyClient.java:59)
	at org.elasticsearch.client.transport.TransportClient.doExecute(TransportClient.java:366)
	at org.elasticsearch.client.support.AbstractClient.execute(AbstractClient.java:408)
	at org.elasticsearch.action.ActionRequestBuilder.execute(ActionRequestBuilder.java:80)
	at org.elasticsearch.action.ActionRequestBuilder.execute(ActionRequestBuilder.java:54)
	at com.netflix.conductor.dao.es5.index.ElasticSearchDAOV5.lambda$updateWithRetry$10(ElasticSearchDAOV5.java:570)
	at com.github.rholder.retry.AttemptTimeLimiters$NoAttemptTimeLimit.call(AttemptTimeLimiters.java:78)
	at com.github.rholder.retry.Retryer.call(Retryer.java:160)
	at com.netflix.conductor.common.utils.RetryUtil.retryOnException(RetryUtil.java:117)
	at com.netflix.conductor.dao.es5.index.ElasticSearchDAOV5.updateWithRetry(ElasticSearchDAOV5.java:569)
[...]

PR #1554, opened 2020-03-02
Fixes issue #1553

skissane:
Fixes issue #1553. The most obvious thing I could think of is to wrap access to the BulkRequestBuilder object in `synchronized(){}`. In my own testing, this resolves the issue both for ES5 and ES6.

### DIFF
--- a/es5-persistence/src/main/java/com/netflix/conductor/dao/es5/index/BulkRequestBuilderWrapper.java
+++ b/es5-persistence/src/main/java/com/netflix/conductor/dao/es5/index/BulkRequestBuilderWrapper.java
@@ -1,0 +1,45 @@
+package com.netflix.conductor.dao.es5.index;
+
+import org.elasticsearch.action.ActionFuture;
+import org.elasticsearch.action.bulk.BulkRequestBuilder;
+import org.elasticsearch.action.bulk.BulkResponse;
+import org.elasticsearch.action.index.IndexRequest;
+import org.elasticsearch.action.update.UpdateRequest;
+
+import javax.annotation.Nonnull;
+import java.util.Objects;
+
+/**
+ * Thread-safe wrapper for {@link BulkRequestBuilder}.
+ */
+public class BulkRequestBuilderWrapper {
+	private final BulkRequestBuilder bulkRequestBuilder;
+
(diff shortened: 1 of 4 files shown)


es5-persistence/src/main/java/com/netflix/conductor/dao/es5/index/BulkRequestBuilderWrapper.java
```diff
@@ -0,0 +1,45 @@
+package com.netflix.conductor.dao.es5.index;
```
apanicker-nflx: Please add license headers to all the files.

apanicker-nflx:
@skissane Thank you for your contribution.
 Please also add the fixes to the ElasticsearchRestDAOV5 and ElasticsearchRestDAOV6 classes.

es5-persistence/src/main/java/com/netflix/conductor/dao/es5/index/BulkRequestBuilderWrapper.java
skissane: @apanicker-nflx Thanks, added license header as requested

skissane:
@apanicker-nflx Regarding the ElasticsearchRestDAOV5 and ElasticsearchRestDAOV6 classes, I don't think this issue is relevant to them. They don't use the `BulkRequestBuilder` class, so thread-safety issues in that class don't affect them. I ran my test script for both ES5 and ES6 using the REST configuration and the issue didn't reproduce for them.

apanicker-nflx:
Thank you verifying the fix against the REST configurations.

es5-persistence/src/main/java/com/netflix/conductor/dao/es5/index/BulkRequestBuilderWrapper.java
apanicker-nflx: @skissane Since the project is open-sourced under Netflix, I believe the license header should use the same name. I understand your concern regarding working on this on company-time, but eventually since this is OSS, it would benefit the community as a whole.

skissane: @apanicker-nflx Can we discuss this on the project chat? I posted a message about it on https://gitter.im/netflix-conductor/community Thanks

PR #1555, opened 2020-03-02
Make use of ElasticSearch optional

skissane:
Adds a configuration parameter which, when set, completely disables use of ElasticSearch. Note that search operations which require ElasticSearch will silently fail by returning empty results.

The new configuration parameter is `workflow.indexing.enabled` with a default value of true. Setting it to false will completely disable use of ElasticSearch.

(Of course, with this configuration, significant parts of the Conductor UI will no longer function correctly. However, not all users of Conductor actually use the Conductor UI, so for such users that will not be an issue.)

### DIFF
--- a/server/src/main/java/com/netflix/conductor/noopindex/NoopEmbeddedElasticSearchProvider.java
+++ b/server/src/main/java/com/netflix/conductor/noopindex/NoopEmbeddedElasticSearchProvider.java
@@ -1,5 +1,5 @@
-/**
- * Copyright 2020 Netflix, Inc.
+/*
+ * Copyright 2020 Medallia, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


server/src/main/java/com/netflix/conductor/noopindex/NoopEmbeddedElasticSearchProvider.java
```diff
@@ -0,0 +1,16 @@
+package com.netflix.conductor.noopindex;
```
apanicker-nflx: Please add license headers to new files.

apanicker-nflx:
@skissane This is a fantastic addition and we really appreciate your contribution.
Please resolve the minor comment and we can merge this.

server/src/main/java/com/netflix/conductor/noopindex/NoopEmbeddedElasticSearchProvider.java
skissane: @apanicker-nflx Thanks. I added the license header using `./gradlew licenseFormat`

On the added files, I put my employer's name in the copyright notice since I made this change on work time.

kensou97:
when will this feature be released?😄

server/src/main/java/com/netflix/conductor/noopindex/NoopEmbeddedElasticSearchProvider.java
apanicker-nflx: I understand your concern about making the changes on work time, however, since the project is open sourced under Netflix, the header would need to say Netflix. Thanks for your contribution.

Issue #407, opened 2018-01-20
Task never polled by worker and timeout


nbraquart:
Hello,

I would have thought the timeoutSeconds attribute set at task definition level would start from the moment the task goes in SCHEDULED state. However, it appears that the timeout runs from the time task has been changed to IN_PROGRESS status.

Thus, if a task definition has no corresponding worker, or if the worker has stopped polling for any reason, the workflows relying on this task will be stuck forever.

The documentation is clear on this point, so I believe this is expected. I was just wondering what were the rationale behind this? And could we maybe imagine a way to handle the case explained above?

Thanks!

lucas-pinheiro-ipremios:
Same problem here

v1r3n:
@nbraquart @lucas-pinheiro-ipremios The solution will be to introduce `poll timeout`.  This should be configurable in Task Definition.  If the task is not polled within the timeout specified, it would get timed out.

We have plans to implement this not prioritized.  Happy to accept a PR if you think you can contribute this.

nbraquart:
Hello @apanicker-nflx or @kishorebanala ,

Do you think this issue could be reopened? I am not sure to understand why it was closed by Viren. Or is there another way to track this potential feature?

Thanks!

PR #1562, opened 2020-03-05
Add a new task def attribute to support task poll timeout

nbraquart:
As explained here: https://github.com/Netflix/conductor/issues/407

### DIFF
--- a/common/src/main/java/com/netflix/conductor/common/metadata/tasks/TaskDef.java
+++ b/common/src/main/java/com/netflix/conductor/common/metadata/tasks/TaskDef.java
@@ -113,6 +113,10 @@ public class TaskDef extends Auditable {
 	@Email(message = "ownerEmail should be valid email address")
 	private String ownerEmail;
 
+	@ProtoField(id = 19)
+	@Min(value = 0, message = "TaskDef pollTimeoutSeconds: {value} must be >= 0")
+	private Integer pollTimeoutSeconds;
+
 	public TaskDef() {
 	}
 
@@ -386,6 +390,20 @@ public class TaskDef extends Auditable {
 	 */
 	public void setOwnerEmail(String ownerEmail) {
 		this.ownerEmail = ownerEmail;
+	}
+
+	/**
+	 * @param pollTimeoutSeconds the poll timeout to set
+	 */
+	public void setPollTimeoutSeconds(Integer pollTimeoutSeconds) {
+		this.pollTimeoutSeconds = pollTimeoutSeconds;
+	}
+
+	/**
+	 * @return the poll timeout of this task definition
+	 */
+	public Integer getPollTimeoutSeconds() {
+		return pollTimeoutSeconds;
 	}
 
 	@Override
--- a/core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
@@ -167,6 +167,7 @@ public class DeciderService {
 
             if (taskDefinition.isPresent()) {
                 checkTaskTimeout(taskDefinition.get(), pendingTask);
+                checkTaskPollTimeout(taskDefinition.get(), pendingTask);
                 // If the task has not been updated for "responseTimeoutSeconds" then mark task as TIMED_OUT
                 if (isResponseTimedOut(taskDefinition.get(), pendingTask)) {
                     timeoutTask(taskDefinition.get(), pendingTask);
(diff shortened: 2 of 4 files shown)


core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
```diff
@@ -524,23 +524,38 @@ void checkTaskTimeout(TaskDef taskDef, Task task) {
...
-        long timeout = 1000L * taskDef.getTimeoutSeconds();
-        long now = System.currentTimeMillis();
-        long elapsedTime = now - (task.getStartTime() + ((long) task.getStartDelayInSeconds() * 1000L));
+        String reason = null;
+        final long now = System.currentTimeMillis();
+        final long startDelay = 1000L * task.getStartDelayInSeconds();
 
-        if (elapsedTime < timeout) {
-            return;
+        if (task.getStatus().equals(SCHEDULED) && taskDef.getPollTimeoutSeconds() > 0) {
```
Jiehong: if `taskDef.getPollTimeoutSeconds()` is `null` (the case for all tasks defined before that functionality has been merged, and it's optional), this check is a bit weird.

Perhaps only checking it's not null is enough instead.

common/src/main/java/com/netflix/conductor/common/metadata/tasks/TaskDef.java
```diff
@@ -113,6 +113,9 @@
 	@Email(message = "ownerEmail should be valid email address")
 	private String ownerEmail;
 
+	@ProtoField(id = 19)
+	private Integer pollTimeoutSeconds;
```
Jiehong: This field shouldn't be allowed to be negative. Could you please add a `@Min` like for `responseTimeoutSeconds` ?

core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
andrea11: I think both need to be checked. We can assume (to write down on the doc) that if pollTimeoutSeconds is set to 0, no checkTaskTimeout should be performed in this case

core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
```diff
@@ -524,23 +524,38 @@ void checkTaskTimeout(TaskDef taskDef, Task task) {
...
         }
-        if (task.getStatus().isTerminal() || taskDef.getTimeoutSeconds() <= 0 || task.getStartTime() <= 0) {
+        if (task.getStatus().isTerminal()) {
             return;
         }
 
-        long timeout = 1000L * taskDef.getTimeoutSeconds();
-        long now = System.currentTimeMillis();
-        long elapsedTime = now - (task.getStartTime() + ((long) task.getStartDelayInSeconds() * 1000L));
+        String reason = null;
```
andrea11: maybe use Optional?

nbraquart: No longer applicable

common/src/main/java/com/netflix/conductor/common/metadata/tasks/TaskDef.java
nbraquart: Done

core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
nbraquart: Done

core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
```diff
@@ -536,8 +537,36 @@ void checkTaskTimeout(TaskDef taskDef, Task task) {
...
+            LOGGER.warn("Missing task definition for task:{}/{} in workflow:{}", task.getTaskId(), task.getTaskDefName(), task.getWorkflowInstanceId());
+            return;
+        }
+        if (taskDef.getPollTimeoutSeconds() == null || taskDef.getPollTimeoutSeconds() <= 0 || !task.getStatus().equals(SCHEDULED)) {
+            return;
+        }
+
+        final long pollTimeout = 1000L * taskDef.getPollTimeoutSeconds();
+        final long now = System.currentTimeMillis();
+        final long pollElapsedTime = now - (task.getScheduledTime() + ((long) task.getStartDelayInSeconds() * 1000L));
```
apanicker-nflx: When a task is sent back from the worker with `callbackAfterSeconds` set and `IN_PROGRESS` status, Conductor will mark this task as SCHEDULED and set it to invisible state in the queue to be available for polling after the specified callbackAfterSeconds. [Reference](https://github.com/Netflix/conductor/blob/master/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java#L816-L819)

In these cases, the `pollElapsedTime` calculation here would wrongly mark this task to be poll timed out. This calculation would need to take into account, this scenario of callbackAfterSeconds and invisibility period on the queues. A similar time calculation can be found in the `isResponseTimedOut` method in this class for reference.

nbraquart: Thanks for the feedback. I have added the callbackAfterSeconds to the timeout for a correct computation as per your suggestion.

Issue #1536, opened 2020-02-14
Making event execution persistence optional


mdepak:
In `EventProcessor` event execution details are [persisted](https://github.com/Netflix/conductor/blob/master/core/src/main/java/com/netflix/conductor/core/events/SimpleEventProcessor.java#L235) and there is no way to clean up these event details in the conductor. Additionally, these event execution details are not used for any logic and there is no way of accessing status without querying the database.

Further, there is no way of removing the event execution indexes as well. So we are trying to make it optional.

Condition on `addEventExecution()` is used in `SimpleEventProcessor` to identify the potential duplicate messages from the event queues. Provided that conductor uses the two queues including Redis queues ( sorted set which does not allow/overrides duplicate message) and SQS (which handles duplicates messages within deduplication window), Is it ok to make addEventExecution() optional to ensure event execution messages are not persisted. 

In the `RedisExecutionDAO`, each event message is saved as a separate key in the Redis's hash and leads to an exponentially larger number of keys and it becomes inefficient to scan the keys and get the event messages. So, shall we use HSETNX with the same key for a particular event type with event IDs as properties of the hash?

apanicker-nflx:
@mdepak Thank you for the suggestions and observations.
We have also been evaluating the persistence of event execution and are looking into ways to avoid persisting this data/cleaning up after execution. 

> shall we use HSETNX with the same key for a particular event type with event IDs as properties of the hash?

This might be helpful to reduce the number of records at the expense of larger hashes. If you can test this out and create a PR for this, we would really appreciate that.

mdepak:
Closing this as PR is merged.

PR #1572, opened 2020-03-10
Event execution deduplication interval

mdepak:
As per https://github.com/Netflix/conductor/issues/1536, in order to clean up the event execution data from Redis, we can set the time to live for those records. 

Because of the dyno-queue implementation of reading pre-fetched ids from the local shards, there can be a chance that the duplicate event messages be delivered to the event processors in case of a multi-node conductor deployment. So `workflow.event.execution.peristence.ttl.seconds` property is introduced and the event execution keys are expired from Redis beyond this provided time interval.  This property can act as a deduplication window in subscribers analogous to the SQS deduplication window on the publisher side.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/orchestration/ExecutionDAOFacade.java
+++ b/core/src/main/java/com/netflix/conductor/core/orchestration/ExecutionDAOFacade.java
@@ -386,7 +386,7 @@ public class ExecutionDAOFacade {
      * @return true if save succeeds, false otherwise.
      */
     public boolean addEventExecution(EventExecution eventExecution) {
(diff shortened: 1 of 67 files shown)


core/src/main/java/com/netflix/conductor/core/orchestration/ExecutionDAOFacade.java
```diff
@@ -387,18 +387,21 @@ public void updateTaskLastPoll(String taskName, String domain, String workerId)
...
-            }
+            addEventExecutionIndex(eventExecution);
         }
+
         return added;
     }
 
     public void updateEventExecution(EventExecution eventExecution) {
         executionDAO.updateEventExecution(eventExecution);
+        addEventExecutionIndex(eventExecution);
```
apanicker-nflx: Please update method name to `indexEventExecution` to avoid confusion.

apanicker-nflx:
@mdepak Thank you for the contribution, please resolve conflicts and make the suggested change and we are good to merge.

PR #1574, opened 2020-03-11
allow input to FORK_JOIN_DYNAMIC task to be empty

rickfish:
If the input to a FORK_JOIN_DYNAMIC task is empty (i.e. empty or non-existent dynamicForkTasksParam), ForkJoinDynamicTaskMapper.java gets a NullPointerException. This Pull Request fixes that by simulating an empty array of tasks.


apanicker-nflx:
Please fix failing tests.

rickfish:
How do I know what the failure was? There doesn't seem to be a way for me to see what the errors were. It built fine for me.

apanicker-nflx:
This is the [build failure](https://travis-ci.com/github/Netflix/conductor/builds/152834351) for this pull request.

rickfish:
Ok, I changed it so the test case would succeed. The fix is still good in that a NullPointerException will not happen, it will be a TerminateWorkflowException. I think it would be better to not throw an exception but rather treat it like there are no tasks to process but either way works.

PR #1575, opened 2020-03-11
add flyway files used to add priority to the combo_queue_message index

rickfish:
When there is a large volume of scheduled tasks of a specific type in the queue_message table, the queries slow down significantly during polling because priority is not indexed and the ORDER BY clause uses priority. In our specific case there were 2.3 million tasks of a type and the  poll query for that type was taking 17 seconds. Adding priority to the combo_queue_message index brought that down to 4 seconds.


apanicker-nflx:
@mashurex @mactaggart @s50600822 Please help review this PR since you are other primary contributors/users to this module.

PR #1579, opened 2020-03-13
added resiliency during task poll

apanicker-nflx:
When polling for a task, if a db operation fails on a dequeued task, this task will now be lost and the corresponding workflow will be unable to progress.
This PR addresses this issue by re-enqueueing this task with a delay so that the workflow will be unblocked.
Note that the priority of this task is set to 0 since this task need not wait for other tasks with different priority to be dequeued (as it was already dequeued once).

### DIFF
--- a/core/src/main/java/com/netflix/conductor/service/ExecutionService.java
+++ b/core/src/main/java/com/netflix/conductor/service/ExecutionService.java
@@ -123,10 +123,19 @@ public class ExecutionService {
 		}
 		String queueName = QueueUtils.getQueueName(taskType, domain, null,null);
 
+		List<String> taskIds = new LinkedList<>();
 		List<Task> tasks = new LinkedList<>();
 		try {
-			List<String> taskIds = queueDAO.pop(queueName, count, timeoutInMilliSecond);
-			for (String taskId : taskIds) {
+			taskIds = queueDAO.pop(queueName, count, timeoutInMilliSecond);
+		} catch (Exception e) {
+			logger.error("Error polling for task: {} from worker: {} in domain: {}, count: {}", taskType, workerId,
+				domain, count, e);
+			Monitors.error(this.getClass().getCanonicalName(), "taskPoll");
+			Monitors.recordTaskPollError(taskType, domain, e.getClass().getSimpleName());
+		}
+
+		for (String taskId : taskIds) {
+			try {
 				Task task = getTask(taskId);
 				if (task == null || task.getStatus().isTerminal()) {
 					// Remove taskId(s) without a valid Task/terminal state task from the queue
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/netflix/conductor/service/ExecutionService.java
```diff
@@ -151,13 +159,14 @@ public Task poll(String taskType, String workerId, String domain) {
 				task.setPollCount(task.getPollCount() + 1);
 				executionDAOFacade.updateTask(task);
 				tasks.add(task);
+			} catch (Exception e) {
+				// db operation failed for dequeued message, re-enqueue with a delay
+				logger.warn("DB operation failed for task: {}, postponing task in queue", taskId, e);
+				queueDAO.postpone(queueName, taskId, 0, queueTaskMessagePostponeSeconds);
```
kishorebanala: It'd be nice to have a metric for this too, which helps in tracing poll failures due to db operations and find out before the clients notice poll delays.

core/src/main/java/com/netflix/conductor/service/ExecutionService.java
```diff
@@ -123,10 +123,18 @@ public Task poll(String taskType, String workerId, String domain) {
...
+		List<String> taskIds = new LinkedList<>();
 		List<Task> tasks = new LinkedList<>();
 		try {
-			List<String> taskIds = queueDAO.pop(queueName, count, timeoutInMilliSecond);
-			for (String taskId : taskIds) {
+			taskIds = queueDAO.pop(queueName, count, timeoutInMilliSecond);
+		} catch (Exception e) {
+			logger.error("Error polling for task: {} from worker: {} in domain: {}, count: {}", taskType, workerId,
+				domain, count, e);
+			Monitors.error(this.getClass().getCanonicalName(), "taskPoll");
```
kishorebanala: How about surfacing this exception to client? Also, should we update `updateTaskLastPoll` if the pop fails?

apanicker-nflx: Yes, this helps to track polling from workers irrespective of whether a task was returned in the poll.

apanicker-nflx: This exception was not surfaced to the client in the previous implementation either. Since polling is a highly-frequent activity, it would clutter up client-side logs in case of a temporary dependency glitch on server-side.

core/src/main/java/com/netflix/conductor/service/ExecutionService.java
apanicker-nflx: Good idea on the metric, will add.

Issue #1585, opened 2020-03-15
Please reconsider decision to make ownerEmail mandatory


skissane:
In our use case, we do not have the concept of a workflow being owned by an individual user.

We are embedding Conductor in a SaaS app. A workflow will belong to a customer of our SaaS app – in some cases to a specific org/department/team/etc within one of our customers – but not necessarily to a specific person. In many cases, the workflow will initially be created by a professional services team member, partner or implementation consultant, who will not be personally responsible for maintaining it in the long-term. We simply don't have the concept of configuration objects being owned by specific individuals, and we don't want to introduce that.

We store user email in our account database. We don't want to store it anywhere else, because then if the user changes their email we only have to change it in our account database. But if we store it in the workflow definition, we could potentially have to change it across many workflows. In our use case, workflow cloning will be common, and so the same email could end up spread across many workflow definitions, which means many points at which we have to change the email.

We don't want Conductor to send emails to anyone directly. We have our own internal problem notification system. We would be happy to have an extension point in Conductor we could use to connect Conductor to our notification system. Our notification system may, in some cases, end up sending emails to some people, but we want to keep the knowledge of whether to send an email, and to whom to send it, centralised in our notification system, not put it inside the workflow definition.

We basically have two options:
1. Set a dummy email everywhere, e.g. `noemail@domain.invalid`
2. Make the ownerEmail optional in our local forked copy of Conductor

We've gone with (2) for now, but that goes against our objective of trying to keep our internal version of Conductor as close to upstream as possible. Option (1) is ugly, and could at some point in the future result in lots of noise in the logs due to attempts to send email to a non-existent address.

apanicker-nflx:
@skissane Thanks for the detailed explanation and we agree with your reasoning. In our case, the `ownerEmail` field  is being populated by our users to be the team DL's to which alerts/notifications will be sent out. 
Thanks for creating the PR to solve this in an elegant way.

PR #1586, opened 2020-03-16
Add config setting to control whether ownerEmail is mandatory or not

skissane:
Addresses issue #1585 – adds a config setting `workflow.owner.email.mandatory` (with a default value of true) which controls whether ownerEmail attribute is mandatory on task definitions and workflow definitions. With the default value of true, the behaviour is the same as at present. If you change the value to false, then ownerEmail becomes an optional attribute.

Implementation is by replacing the `@NotEmpty(message = "ownerEmail cannot be empty")` annotation on ownerEmail with a new custom annotation `@OwnerEmailMandatoryConstraint`. We push the value of the config setting into the annotation validator; when the setting is true, it ensures the value is neither null nor an empty string; when the setting is false, the validation always succeeds.

### DIFF
--- a/common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowDef.java
+++ b/common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowDef.java
@@ -19,6 +19,7 @@ import com.github.vmg.protogen.annotations.ProtoEnum;
 import com.github.vmg.protogen.annotations.ProtoField;
 import com.github.vmg.protogen.annotations.ProtoMessage;
 import com.netflix.conductor.common.constraints.NoSemiColonConstraint;
+import com.netflix.conductor.common.constraints.OwnerEmailMandatoryConstraint;
 import com.netflix.conductor.common.constraints.TaskReferenceNameUniqueConstraint;
 import com.netflix.conductor.common.metadata.Auditable;
 import java.util.HashMap;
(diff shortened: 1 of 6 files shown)


common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowDef.java
```diff
@@ -84,7 +85,7 @@
 	private boolean workflowStatusListenerEnabled = false;
 
 	@ProtoField(id = 11)
-	@NotEmpty(message = "ownerEmail cannot be empty")
+	@OwnerEmailMandatoryConstraint
```
flavioschuindt: Hi, @skissane.

Nice PR. Thanks for working on it.

I would just rename this for ```OwnerEmailMandatory```. We already know that this is a Constraint because of the word ```mandatory```. What do you think?

skissane: Well, if you look at the other annotations in `com.netflix.conductor.common.constraints` package, they've all got `Constraint` at the end of their name as well. So, even though I agree the word is a bit superfluous, it is consistent with the others.

flavioschuindt: Makes sense, @skissane. Thank you.

PR #1590, opened 2020-03-17
Document MySQL lack of rate limiting support

scarlac:
The lack of support for time-based rate limiting for a MySQL-backed Conductor seems completely undocumented. If it's documented, it's fairly hidden (I was not able to find it by specifically Googling "conductor rate limit"). It should be clearly documented in context.

### DIFF
--- a/docs/docs/configuration/taskdef.md
+++ b/docs/docs/configuration/taskdef.md
@@ -68,7 +68,8 @@ If you have 1000 task executions waiting in the queue, and 1000 workers polling 
 * `rateLimitPerFrequency`defines the number of Tasks that can be given to Workers per given "frequency window".  
 **Example:**  
 Let's set `rateLimitFrequencyInSeconds = 5`, and `rateLimitPerFrequency = 12`. This means, our frequency window is of 5 seconds duration, and for each frequency window, Conductor would only give 12 tasks to workers. So, in a given minute, Conductor would only give 12*(60/5) = 144 tasks to workers irrespective of the number of workers that are polling for the task.  
-Note that unlike `concurrentExecLimit`, rate limiting doesn't take into account tasks already in progress/completed. Even if all the previous tasks are executed within 1 sec, or would take a few days, the new tasks are still given to workers at configured frequency, 144 tasks per minute in above example.
+Note that unlike `concurrentExecLimit`, rate limiting doesn't take into account tasks already in progress/completed. Even if all the previous tasks are executed within 1 sec, or would take a few days, the new tasks are still given to workers at configured frequency, 144 tasks per minute in above example.   
+Note: Rate limiting is only supported for the Redis-persistence module and is not available with other persistence layers.
 
 ### Using inputKeys and outputKeys
 


docs/docs/configuration/taskdef.md
```diff
@@ -68,7 +68,8 @@ If you have 1000 task executions waiting in the queue, and 1000 workers polling
...
 **Example:**  
 Let's set `rateLimitFrequencyInSeconds = 5`, and `rateLimitPerFrequency = 12`. This means, our frequency window is of 5 seconds duration, and for each frequency window, Conductor would only give 12 tasks to workers. So, in a given minute, Conductor would only give 12*(60/5) = 144 tasks to workers irrespective of the number of workers that are polling for the task.  
-Note that unlike `concurrentExecLimit`, rate limiting doesn't take into account tasks already in progress/completed. Even if all the previous tasks are executed within 1 sec, or would take a few days, the new tasks are still given to workers at configured frequency, 144 tasks per minute in above example.
+Note that unlike `concurrentExecLimit`, rate limiting doesn't take into account tasks already in progress/completed. Even if all the previous tasks are executed within 1 sec, or would take a few days, the new tasks are still given to workers at configured frequency, 144 tasks per minute in above example.   
+Note: MySQL does not currently support `rateLimitPerFrequency` and `rateLimitFrequencyInSeconds`. Please use `concurrentExecLimit` instead.
```
apanicker-nflx: @scarlac Thanks for pointing this out and updating the documentation accordingly. The implementation of the `RateLimitingDAO` is provided in the redis-persistence module only and is not available with other persistence layers. Could you please modify this documentation accordingly.

docs/docs/configuration/taskdef.md
```diff
@@ -68,7 +68,9 @@ If you have 1000 task executions waiting in the queue, and 1000 workers polling
...
-Note that unlike `concurrentExecLimit`, rate limiting doesn't take into account tasks already in progress/completed. Even if all the previous tasks are executed within 1 sec, or would take a few days, the new tasks are still given to workers at configured frequency, 144 tasks per minute in above example.
+Note that unlike `concurrentExecLimit`, rate limiting doesn't take into account tasks already in progress/completed. Even if all the previous tasks are executed within 1 sec, or would take a few days, the new tasks are still given to workers at configured frequency, 144 tasks per minute in above example.   
+The implementation of the RateLimitingDAO is provided in the redis-persistence module only and is not available with other persistence layers.
+Note: Rate limiting is only supported for the Redis-persistence module and is not available with other persistence layers. Consider using `concurrentExecLimit` instead if you are using a different persistence module.
```
scarlac: @apanicker-nflx Changed the wording. How does this sound?

apanicker-nflx: Looks like lines 72 and 73 are identical. Was this intentional?

apanicker-nflx: Also, the usage of `concurrentExecLimit` is slightly different as compared to rate limiting as explained in line 71. Not sure if we want to suggest this as a workaround.

scarlac: Def not intentional. Will fix tomorrow. Will remove the reference to concurrentExecLimit. Can users expect it to work for all persistence modules?

apanicker-nflx: The `concurrentExecLimit` functionality is currently implemented in all the persistence modules and should work correctly.
For rate limiting however, this feature may not be possible to be implemented in all the persistence layers, this was the primary motivation behind extracting this out of the `ExecutionDAO` and moving this into its own `RateLimitingDAO`. This enables users to implement the rate limiting functionality as needed, even using a custom rate limiter that may already be deployed in their environments.

scarlac: @apanicker-nflx There we go. How's that?

apanicker-nflx: Awesome, thanks a lot.

PR #1592, opened 2020-03-18
Reset SubworkflowId on task retry.

kishorebanala:


apanicker-nflx:
Similar change would need to be made in [WorkflowExecutor](https://github.com/Netflix/conductor/blob/master/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java#L514) as well. 
Also please add an integration test to the test-harness to cover this case.

kishorebanala:
> Similar change would need to be made in [WorkflowExecutor](https://github.com/Netflix/conductor/blob/master/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java#L514) as well.
> Also please add an integration test to the test-harness to cover this case.

Good catch. I've updated the PR, please check.

PR #1608, opened 2020-03-27
change nonterminal task statuses to SKIPPED when a TERMINATE task executes

rickfish:
Currently when a TERMINATE task executes, it terminates the workflow early but leaves any tasks that were SCHEDULED or IN_PROGRESS as is; therefore they might still be pollable. This change will modify those tasks and change their statuses to SKIPPED. For some reason, when WAIT task status is changed from  IN_PROGRESS to SKIPPED, updating the task happens and it gets reverted back to IN_PROGRESS but that does not really cause any issues.


core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -985,6 +987,26 @@ public boolean decide(String workflowId) {
...
+                                 * The TERMINATE task completes the workflow but does not do anything with SCHEDULED or IN_PROGRESS tasks to complete them
+                                 */
+                                for(Task workflowTask : workflow.getTasks()) {
+                                	if(workflowTask != task && !workflowTask.getStatus().isTerminal()) {
+                                		workflowTask.setStatus(SKIPPED);
+                                		tasksToBeUpdated.add(workflowTask);
+                                	}
+                                }
+                                /*
+                                 * Now find nested subworkflows that also need to have their tasks skipped
```
apanicker-nflx, line 1000: In the case of sub workflows, would it be better to terminate the sub workflow instead? The reason could be that the parent workflow was terminated using TERMINATE task.

apanicker-nflx:
@rickfish Please add integration tests for this change similar to https://github.com/Netflix/conductor/blob/master/test-harness/src/test/java/com/netflix/conductor/tests/integration/AbstractWorkflowServiceTest.java#L5274.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
rickfish: There are a couple of issues that make terminating the subworkflows problematic (I think). Firstly, I think if the terminate task specifies COMPLETED as the terminationStatus and the subworkflow is terminated, the termination of the subworkflow would actually cause the ultimate status of the parent workflow to be FAILED which wouldn't honor the contract of the terminate task.

Also, I tried the below code and there is a StackOverflowError because WorkflowExecutor.decide() would call terminateWorkflow() which would call decide() to terminate the parent workflow, etc.

```
if (TERMINATE.name().equals(task.getTaskType())) {
	workflow.setStatus(workflowInstance.getStatus());
	workflow.setOutput(workflowInstance.getOutput());
	deciderService.externalizeWorkflowData(workflow);
	/*
	 * The TERMINATE task completes the workflow but does not do anything with SCHEDULED or IN_PROGRESS tasks to complete them
	 */
	for(Task workflowTask : workflow.getTasks()) {
		if(workflowTask != task && !workflowTask.getStatus().isTerminal()) {
			workflowTask.setStatus(SKIPPED);
			tasksToBeUpdated.add(workflowTask);
		}
	}
	/*
	 * Now find nested subworkflows that also need to have their tasks skipped
	 */
	for(Task workflowTask : workflow.getTasks()) {
		if(TaskType.SUB_WORKFLOW.name().equals(workflowTask.getTaskType()) && StringUtils.isNotBlank(workflowTask.getSubWorkflowId())) {
			terminateWorkflow(workflowTask.getSubWorkflowId(), "Parent workflow was terminated with a TERMINATE task");
		/*                                		
			Workflow subWorkflow = executionDAOFacade.getWorkflowById(workflowTask.getSubWorkflowId(), true);
			if(subWorkflow != null) {
				skipTasksAffectedByTerminateTask(subWorkflow);                                		
			}
		*/                                		
		}
	}
}

```
That said, I could probably make it work but that would involve a lot of changes that might create unforeseen problems elsewhere.

What are your thoughts?

rickfish:
I added additional testing to the existing test cases to cover subworkflows and tasks being set to SKIPPED.

test-harness/src/test/java/com/netflix/conductor/tests/integration/AbstractWorkflowServiceTest.java
```diff
@@ -5272,64 +5274,43 @@ public void testLambda() {
...
-        lambdaWorkflowTask.setTaskReferenceName("lambda0");
+    @Test
+    public void testTerminateTaskWithFailedStatus() {
+    	testTerminateTask(WorkflowStatus.FAILED);
+    }
 
-        Map<String, Object> terminateTaskInputParams = new HashMap<>();
-        terminateTaskInputParams.put(Terminate.getTerminationStatusParameter(), "COMPLETED");
-        terminateTaskInputParams.put(Terminate.getTerminationWorkflowOutputParameter(), "${lambda0.output}");
+    private WorkflowDef createSubworkflowDefWithWaitTask() {
```
apanicker-nflx, line 5285: Since this is a workflow with wait task, this method can be named `createWorkflowWithWaitTask()` for better clarity.

test-harness/src/test/java/com/netflix/conductor/tests/integration/AbstractWorkflowServiceTest.java
```diff
@@ -5350,11 +5337,28 @@ public void testTerminateTaskWithFailedStatus() {
...
+        subWorkflow.setTaskReferenceName("sw1");
+
+        fanoutTask.getForkTasks().add(Arrays.asList(subWorkflow));
+        fanoutTask.getForkTasks().add(Arrays.asList(lambdaWorkflowTask, waitWorkflowTask, terminateWorkflowTask));
+
+        // join task
+        WorkflowTask joinTask = new WorkflowTask();
+        joinTask.setType(TaskType.JOIN.name());
+        joinTask.setTaskReferenceName("fanouttask_join");
+        joinTask.setJoinOn(Arrays.asList("sw1", "leftfork-wait0"));
```
apanicker-nflx, line 5355: The last task in the left fork is the terminate task, however, the join is set to an intermediate task (wait). This looks incorrect.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
apanicker-nflx: >f the terminate task specifies COMPLETED as the terminationStatus and the subworkflow is terminated, the termination of the subworkflow would actually cause the ultimate status of the parent workflow to be FAILED

This observation is correct. Agreed with the reasoning and implementation.

test-harness/src/test/java/com/netflix/conductor/tests/integration/AbstractWorkflowServiceTest.java
apanicker-nflx: Also, please add another task after the wait task in this workflow, and check that this task is not scheduled when parent workflow is completed using terminate.

rickfish: made the requested changes

test-harness/src/test/java/com/netflix/conductor/tests/integration/AbstractWorkflowServiceTest.java
rickfish: made the requested change

apanicker-nflx:
Please resolve conflicts and we are good to merge.

rickfish:
I am not sure how to reconcile my changes with the change that was just made to create the conflict (PR #1658). They seem to be 2 ways to do the same thing except that change does not seem to allow for letting the TERMINATE task to set the workflow status to COMPLETED as an end result. Confused...

rickfish:
Actually, after looking at it further - maybe PR #1658 is a better, cleaner solution than mine. I am thinking about closing this PR. Thoughts?

rickfish:
Sorry for the multiple comments. I just tested the PR #1658 change and it doesn't handle terminating the subworkflow if there is one. So I am stuck trying to figure out what that PR does that mine does not. 

@kishorebanala can you elaborate? do we need both PR's? I don't know what mine is missing that causes yours to be needed?

kishorebanala:
Hey @rickfish, Sorry for delays. Can you provide more details about ```PR #1658 change and it doesn't handle terminating the subworkflow if there is one.``` please. An example would help too.

If the current workflow's terminate task is invoked, and if it happens to also have a parent workflow, #1658 takes care of calling decide on parent workflow, and updating the parent status accordingly. LMK if we're not on the same page.

rickfish:
Your use case is opposite of mine. If the parent workflow has a terminate task and it has a sub-workflow, that sub-workflow is not handled properly, i.e. tasks in it are still SCHEDULED. I am going to attach a sub-workflow (SIO_ETD_TERMINATE_SUBWF) and a parent workflow (SIO_ETD_TERMINATE_WF). If you start the parent workflow, then complete the 'basicJava0' task you will see the results. My change takes care of handling the sub-workflow, not sure if it is the best solution but it works. 

LMK if you see differently and we will compare notes.

Actually, what is the best way to upload JSON files? I tried to attach and I guess you can only attach images?

rickfish:
Is this ever going to be merged? This is the second time, while waiting for it to be merged, some other change came in and made me do everything all over because it changed the underlying code. If it is going to be merged i will make my changes again. If not, please let me know so i don't have to do it again. I do believe this is a necessary change.

kishorebanala:
@rickfish Sorry, it slipped through. I'll try to reproduce the use-case you mentioned above and verify if the fix is still valid. Please send me over the example at kishore.mizzou [@] gmail [.] com. Alternatively, you can copy them over to https://gist.github.com/ to make it accessible to this issue for anyone else interested.

rickfish:
@kishorebanala thanks, I sent them to your email as well as adding them as a gist under rickfish/SIO_ETD_TERMINATE_SUBWF.json.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -985,29 +986,56 @@ public boolean decide(String workflowId) {
...
+                        } else if (SUB_WORKFLOW.name().equals(task.getTaskType()) && task.getStatus().equals(IN_PROGRESS)) {
+                            // Verifies and updates the task inplace, based on the Subworkflow and parent Workflow state,
+                            // and continues with the current decide.
+                            if (updateParentWorkflow(task, workflow)) {
+                                tasksToBeUpdated.add(task);
+                                stateChanged = true;
+                            }
                         }
+                    } catch (Exception e) {
+                        throw new ApplicationException(Code.INTERNAL_ERROR, String.format("Unable to start system task: %s", workflowSystemTask.getName()), e);
```
kishorebanala, line 1038: This exception message: ```Unable to start system task:``` is not relevant, as we're not starting any system tasks here.

kishorebanala:
Hey @rickfish , this looks good to me, except for the comment above about exception message.

Issue #1315, opened 2019-09-13
Best practice : Cleanup of redis and elasticsearch database (and synchronization)


csperli:
Hi all,

I'm searching for a best practice solution regarding cleanup and synchronization of Redis and ES databases.
We have configured conductor with db=dynomite.
Using dynomite cluster with 3 nodes and underlying redis db.

After a certain time of working with conductor, redis db is full.

- Configuring redis without eviction, we are facing some redis memory errors at this time, which then causes errors in conductor. (Which is clear)

- Configuring redis with eviction, the redis memory errors are gone. But eviction not only removes old workflows but also our custom workflow and task definitions. So again we are facing problems in conductor. At this scenario we also facing the problem of inconsistency between redis db and elasticsearch db. So the workflow in redis may be removed but they are still there in elasticsearch.

So my question is, is there any proposal how to solve this database problems ?
What is the best practice to do, when we see redis database is reaching its limit ?
How can we cleanup redis database and also get synchronization to elasticsearch database ?
Do we have to implement own cleanup strategy using the rest api /workflow/{workflowid}/remove with archiveWorkflow=false or are there any other proposals ?
Is there a possibilty to use redis eviction for that, without getting workflow defs removed and also synchonization with elasticsearch database ?

A lot of questions, I know.
I would be glad if someone could help us with these questions.

Thank you in advance

Chris

manan164:
Hi @csperli . We are in same boat. I have asked this [question](https://github.com/Netflix/conductor/issues/1243) earlier.
1) For high throughput workflow we are using archive workflow true so it will be removed from redis on completion. 
2) We are using async indexing in ES with queue size 15k, which is also not sufficient during peak hours. And after the queue is full, we drop the indexes.  During non peak hours it is fine.

So potentially there are two issues,
1) Redis is getting piled up.
2) Workflow are not searchable from UI because it is getting dropped after queue is full.

We will not be able solve 1st problem without solving 2nd one otherwise till the workflow is in redis we will be able to search.
We are trying to solve the second problem in phases.
These are the approaches we have tried.
1) Introduce batch indexing while indexing into ES. It is mentioned in the PR:- https://github.com/Netflix/conductor/pull/1269. We saw much improvement with the batch size of 100. This number is not result of any hypothesis. You are free to use any.
2) Limit the number of fields that are being indexed. We did not see much improvement here.
[...]

PR #1609, opened 2020-03-27
add workflow cleaner to clean expired workflows periodically

kensou97:
Hi, I ran into a problem while using redis persistence, which has already been discussed  in some issues like #1315.  Redis memory usage kept growing and random data including some metadata got evicted.
So I add a workflow cleaner in RedisExecutionDAO. It records the timestamp and workflowId in zset (timestamp as score, workflowId as value) when a workflow is created. A background thread will find and clean the expired workflows (indicated by config **workflow.cleaner.expire.seconds**) periodically. The period is indicated by config **workflow.cleaner.period.seconds**. If there are too many expired workflows, cleaner will process them batch by batch. The batch size is indicated by config **workflow.cleaner.batch.size**.
The cleaner is not meant to clean data in IndexDAO or other db implementation, because the disk-based storage may not have this problem. It works well in my environment and I hope this would help someone else.


muralicmr:
@kensou97 - was there a reason this PR was not merged. We were looking at the archival of workflows once execution was complete. This PR seems to have sorted the problem.

PR #1565, opened 2020-03-06
Default workflow archiving module

pavelhalabala:
Implements issue https://github.com/Netflix/conductor/issues/1563

PR #1610, opened 2020-03-27
add workflow cleaner to clean expired workflows periodically

kensou97:
Hi, I ran into a problem while using redis persistence, which has already been discussed  in some issues like #1315.  Redis memory usage kept growing and random data including some metadata got evicted.
So I add a workflow cleaner in RedisExecutionDAO. It records the timestamp and workflowId in zset (timestamp as score, workflowId as value) when a workflow is created. A background thread will find and clean the expired workflows (indicated by config **workflow.cleaner.expire.seconds**) periodically. The period is indicated by config **workflow.cleaner.period.seconds**. If there are too many expired workflows, cleaner will process them batch by batch. The batch size is indicated by config **workflow.cleaner.batch.size**.
The cleaner is not meant to clean data in IndexDAO or other db implementation, because the disk-based storage may not have this problem. It works well in my environment and I hope this would help someone else.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/config/Configuration.java
+++ b/core/src/main/java/com/netflix/conductor/core/config/Configuration.java
@@ -82,6 +82,15 @@ public interface Configuration {
     String INDEXING_ENABLED_PROPERTY_NAME = "workflow.indexing.enabled";
     boolean INDEXING_ENABLED_DEFAULT_VALUE = true;
 
+    String WORKFLOW_CLEANER_EXPIRE_SECONDS_PROPERTY_NAME = "workflow.cleaner.expire.seconds";
+    long WORKFLOW_CLEANER_EXPIRE_SECONDS_DEFAULT_VALUE = -1;
+
+    String WORKFLOW_CLEANER_PERIOD_SECONDS_PROPERTY_NAME = "workflow.cleaner.period.seconds";
+    long WORKFLOW_CLEANER_PERIOD_SECONDS_DEFAULT_VALUE = 1800;
+
+    String WORKFLOW_CLEANER_BATCH_SIZE_PROPERTY_NAME = "workflow.cleaner.batch.size";
+    long WORKFLOW_CLEANER_BATCH_SIZE_DEFAULT_VALUE = 200;
+
     String TASK_DEF_REFRESH_TIME_SECS_PROPERTY_NAME = "conductor.taskdef.cache.refresh.time.seconds";
     int TASK_DEF_REFRESH_TIME_SECS_DEFAULT_VALUE = 60;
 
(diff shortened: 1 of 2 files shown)


manan164:
Hi @kensou97 ,
Really nice work. Even we are facing this issue in our environment.
Have you tried considering [WorkflowSweeper](https://github.com/Netflix/conductor/blob/master/core/src/main/java/com/netflix/conductor/core/execution/WorkflowSweeper.java)?
It is not helpful for use because while doing redis.pop we are facing [issue](https://github.com/Netflix/conductor/issues/1595).
Also, I feel in case of the environment having short-lived workflows ~ seconds and long-lived workflows ~days, this approach will clean both together. Instead, I feel we should add the variable **workflow.cleaner.expire.seconds** in the workflow definition   
@mudit3774 . This is important for us.

kensou97:
> Hi @kensou97 ,
> Really nice work. Even we are facing this issue in our environment.
> Have you tried considering [WorkflowSweeper](https://github.com/Netflix/conductor/blob/master/core/src/main/java/com/netflix/conductor/core/execution/WorkflowSweeper.java)?
> It is not helpful for use because while doing redis.pop we are facing [issue](https://github.com/Netflix/conductor/issues/1595).
> Also, I feel in case of the environment having short-lived workflows ~ seconds and long-lived workflows ~days, this approach will clean both together. Instead, I feel we should add the variable **workflow.cleaner.expire.seconds** in the workflow definition
> @mudit3774 . This is important for us.

hi @manan164 , thanks for your reply.  Because sweeper is enabled by default and my memory usage still kept growing before, so i think sweeper doesn't work as I expected. And as you mentioned, if there is a "expire" property in workflow def, we can have more flexible and accurate control with different workflows. Although current commit has already met my needs, I can patch a commit if necessary.

manan164:
> > Hi @kensou97 ,
> > Really nice work. Even we are facing this issue in our environment.
> > Have you tried considering [WorkflowSweeper](https://github.com/Netflix/conductor/blob/master/core/src/main/java/com/netflix/conductor/core/execution/WorkflowSweeper.java)?
> > It is not helpful for use because while doing redis.pop we are facing [issue](https://github.com/Netflix/conductor/issues/1595).
> > Also, I feel in case of the environment having short-lived workflows ~ seconds and long-lived workflows ~days, this approach will clean both together. Instead, I feel we should add the variable **workflow.cleaner.expire.seconds** in the workflow definition
> > @mudit3774 . This is important for us.
> 
> hi @manan164 , thanks for your reply. Because sweeper is enabled by default and my memory usage still kept growing before, so I think sweeper doesn't work as I expected. And as you mentioned, if there is a "expire" property in workflow def, we can have more flexible and accurate control with different workflows. Although current commit has already met my needs, I can patch a commit if necessary.

Hi @kensou97 The reason for that might be you have not implemented any archival strategy as such. We can implement workflowStatusListener and decide what should be done when workflow is completed or terminated. That will be a cleaner approach I guess. Please check here:- [Archival Module](https://github.com/Netflix/conductor/pull/1565)

kensou97:
> > > Hi @kensou97 ,
> > > Really nice work. Even we are facing this issue in our environment.
> > > Have you tried considering [WorkflowSweeper](https://github.com/Netflix/conductor/blob/master/core/src/main/java/com/netflix/conductor/core/execution/WorkflowSweeper.java)?
> > > It is not helpful for use because while doing redis.pop we are facing [issue](https://github.com/Netflix/conductor/issues/1595).
> > > Also, I feel in case of the environment having short-lived workflows ~ seconds and long-lived workflows ~days, this approach will clean both together. Instead, I feel we should add the variable **workflow.cleaner.expire.seconds** in the workflow definition
> > > @mudit3774 . This is important for us.
> > 
> > 
> > hi @manan164 , thanks for your reply. Because sweeper is enabled by default and my memory usage still kept growing before, so I think sweeper doesn't work as I expected. And as you mentioned, if there is a "expire" property in workflow def, we can have more flexible and accurate control with different workflows. Although current commit has already met my needs, I can patch a commit if necessary.
> 
> Hi @kensou97 The reason for that might be you have not implemented any archival strategy as such. We can implement workflowStatusListener and decide what should be done when workflow is completed or terminated. That will be a cleaner approach I guess. Please check here:- [Archival Module](https://github.com/Netflix/conductor/pull/1565)


I didn't see this feature before, this should be what I need . Thank you😄

manan164:
I would consider this approach better in a way that we won't be facing the issue I have mentioned in the [Archival Module](https://github.com/Netflix/conductor/pull/1565). 
If we have sub-workflow and parent workflows need its output at a later point of time and it is not present in both redis and ES then parent workflow will never be able to complete. This will only occur in a high throughput scenario.

PR #1626, opened 2020-04-04
ElasticSearch Rest  BulkIndex NPE fix for Race condition

mdepak:
As per https://github.com/Netflix/conductor/issues/1553, there is a race condition in `BulkRequestBuilder` class of ElasticSearch which is used in ElasticSearchDAO transport client. In our setup, this NullPointerException happens in case of ElasticSearchRestDAO as well. ElasticSearchRestDAO uses `BulkRequest` object which in turn uses `ArrayList` to maintain the index requests is not synchronized. Hence this problem occurs in Rest implementation as well. So using an approach similar to https://github.com/Netflix/conductor/pull/1554 to use thread-safe wrapper for Rest implementation as well. Verified that in high throughput scenario the NullPointerExceptions are happening no more. Thanks, @skissane for the initial solution.

### DIFF
--- a/es5-persistence/src/main/java/com/netflix/conductor/dao/es5/index/BulkRequestWrapper.java
+++ b/es5-persistence/src/main/java/com/netflix/conductor/dao/es5/index/BulkRequestWrapper.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package com.netflix.conductor.dao.es5.index;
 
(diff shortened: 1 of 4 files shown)


apanicker-nflx:
@skissane Referencing your [comment](https://github.com/Netflix/conductor/pull/1554#issuecomment-596289037) from #1554 , can you please comment on this pull request.

es5-persistence/src/main/java/com/netflix/conductor/dao/es5/index/BulkRequestWrapper.java
```diff
@@ -0,0 +1,43 @@
+package com.netflix.conductor.dao.es5.index;
```
apanicker-nflx, line 1: Please add a license header.

es6-persistence/src/main/java/com/netflix/conductor/dao/es6/index/BulkRequestWrapper.java
```diff
@@ -0,0 +1,43 @@
+package com.netflix.conductor.dao.es6.index;
```
apanicker-nflx, line 1: Please add a license header.

es5-persistence/src/main/java/com/netflix/conductor/dao/es5/index/BulkRequestWrapper.java
mdepak: Added license header.

es6-persistence/src/main/java/com/netflix/conductor/dao/es6/index/BulkRequestWrapper.java
mdepak: Added license header.

PR #1630, opened 2020-04-08
tasklog limit - allow more than 10 entries

s50600822:
When query elasticsearch if size is not specified ES server will serve only 10 first entries by default, this caused lost of task_log on conductor-ui

Assuming we have lots of task logs entries:
```
TASK_ID=$1

for i in {1..20};
do curl -X POST --header 'Content-Type: application/json' --header 'Accept: application/json' "http://localhost:9000/api/tasks/$1/log" -d "$i";
done
```
Currently we can only see:
before

After the change, started with `task_elasticsearch_query_size=20`
```
docker run --rm --name conductor -p 9000:8080 -p 5000:5000  -edb=memory -eworkflow.elasticsearch.instanceType=memory -etask_elasticsearch_query_size=20 s50600822/conductor:all-tasklog-limit-4
```
after-startingwith-limit-at-20

### DIFF
--- a/es5-persistence/src/main/java/com/netflix/conductor/elasticsearch/ElasticSearchConfiguration.java
+++ b/es5-persistence/src/main/java/com/netflix/conductor/elasticsearch/ElasticSearchConfiguration.java
@@ -76,6 +76,9 @@ public interface ElasticSearchConfiguration extends Configuration {
     String ELASTIC_SEARCH_INDEX_REPLICAS_COUNT_PROPERTY_NAME = "workflow.elasticsearch.index.replicas.count";
     int ELASTIC_SEARCH_INDEX_REPLICAS_COUNT_DEFAULT_VALUE = 1;
 
+    String ELASTIC_SEARCH_TASK_LOG_RESULT_LIMIT = "task.elasticsearch.query.size";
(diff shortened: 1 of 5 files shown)


es5-persistence/src/main/java/com/netflix/conductor/elasticsearch/ElasticSearchConfiguration.java
```diff
@@ -76,6 +76,9 @@
     String ELASTIC_SEARCH_INDEX_REPLICAS_COUNT_PROPERTY_NAME = "workflow.elasticsearch.index.replicas.count";
     int ELASTIC_SEARCH_INDEX_REPLICAS_COUNT_DEFAULT_VALUE = 1;
 
+    String ELASTIC_SEARCH_TASK_LOG_RESULT_LIMIT = "task.elasticsearch.query.size";
```
apanicker-nflx, line 79: could be renamed to `tasklog.elasticsearch.query.size` for clarity.

apanicker-nflx:
@s50600822 This is a valuable addition and we really appreciate your contribution. 
Please also make the corresponding change to the es6-persistence module.

PR #1631, opened 2020-04-08
Update elasticsearch when workflows get paused/resumed

rickfish:
When workflows get paused or resumed, indexDAO.indexWorkflow() is not invoked so the workflow status stays at RUNNING even if it is PAUSED. Therefore you cannot get an accurate list of workflows when using workflow/search with freeText of status:PAUSED. This calls indexWorkflow() when the workflow is paused or resumed.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/orchestration/ExecutionDAOFacade.java
+++ b/core/src/main/java/com/netflix/conductor/core/orchestration/ExecutionDAOFacade.java
@@ -225,7 +225,9 @@ public class ExecutionDAOFacade {
             } else {
                 indexDAO.asyncIndexWorkflow(workflow);
             }
-            workflow.getTasks().forEach(indexDAO::asyncIndexTask);
+			if (workflow.getStatus().isTerminal()) {
+				workflow.getTasks().forEach(indexDAO::asyncIndexTask);
+			}
         } else {
             indexDAO.indexWorkflow(workflow);
         }
--- a/core/src/main/java/com/netflix/conductor/core/orchestration/ExecutionDAOFacade.java
+++ b/core/src/main/java/com/netflix/conductor/core/orchestration/ExecutionDAOFacade.java
@@ -210,48 +210,25 @@ public class ExecutionDAOFacade {
      * @return the id of the updated workflow
      */
     public String updateWorkflow(Workflow workflow) {
-    	return updateWorkflow(workflow, false);
-    }
-
-    /**
-     * Updates the given workflow in the data store
-     *
-     * @param workflow the workflow tp be updated
-     * @param indexRegardless whether or not to index the workflow no matter what - note that terminal workflows will already be indexed anyway
-     * @return the id of the updated workflow
-     */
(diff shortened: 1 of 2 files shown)


apanicker-nflx:
Continuing the conversation from #1546 which addresses task indexing, I believe this is a valuable change to the workflow indexing. 
In a similar vein, the workflow indexing was changed to be asynchronous and only to be indexed when workflow starts and workflow ends. All intermediate workflow updates were being skipped. I believe, we can change this such that this behavior is used only if async indexing is enabled. If indexing is synchronous, then all workflow updates would be indexed. This would remove the need for the `indexRegardless` boolean check.

rickfish:
Not sure I completely understand what you were saying. I think regardless of whether indexing is async, all statuses should be recorded in ES. I made changes based on what I think you meant but I might be wrong. That said, I am not completely sure I understand why short-running workflows are indexed with a DelayWorkflowUpdate object. Confusing to me.

apanicker-nflx:
> why short-running workflows are indexed with a DelayWorkflowUpdate object.

Consider a scenario where multiple instances of the Conductor server are deployed. In this case, it is possible that a workflow was started on one node but completed execution on another server node. When async indexing is used, the indexing operation is performed off of an in-memory queue. Depending on the depth of this queue on each node, it is possible that these actions are performed out of order which means that the start workflow indexing operation (with RUNNING status) from node A could be performed after the update workflow indexing operation (with COMPLETED status) from node B. By introducing a delay in the update indexing operation, we minimize the possibility of this occurrence.

core/src/main/java/com/netflix/conductor/core/orchestration/ExecutionDAOFacade.java
```diff
@@ -215,21 +215,19 @@ public String updateWorkflow(Workflow workflow) {
...
+                final String workflowId = workflow.getWorkflowId();
+                DelayWorkflowUpdate delayWorkflowUpdate = new DelayWorkflowUpdate(workflowId);
+                LOGGER.debug("Delayed updating workflow: {} in the index by {} seconds", workflowId, config.getAsyncUpdateDelay());
+                scheduledThreadPoolExecutor.schedule(delayWorkflowUpdate, config.getAsyncUpdateDelay(), TimeUnit.SECONDS);
+                Monitors.recordWorkerQueueSize("delayQueue", scheduledThreadPoolExecutor.getQueue().size());
             } else {
-                indexDAO.indexWorkflow(workflow);
+                indexDAO.asyncIndexWorkflow(workflow);
             }
+            workflow.getTasks().forEach(indexDAO::asyncIndexTask);
```
apanicker-nflx, line 228: As part of async indexing, the contract is that task data for a workflow will only be indexed when a workflow reaches a terminal state. Please move this line to within the if block above.

rickfish: Oh, now I understand what the intent was...thanks for that. I think I covered that now. I took out the isTerminal() check earlier so I added it later.

PR #1632, opened 2020-04-09
use semaphore instead of inmemory queue for processing system tasks

apanicker-nflx:
Instead of relying on in memory LinkedBlockingQueue to store dequeued task ids and execute them asynchronously using a threadpool, this PR provides an implementation to use a semaphore to control when a task id will be dequeued from a given system task queue based on if a thread is available for processing the task. The processing of the task will still happen asynchronously.

In the case of isolated system task queues, instead of using a separate LinkedBlockingQueue to feed the threadpool, each isolated task queue uses a separate semaphore to gate the processing of tasks in its threadpool.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/tasks/SystemTaskExecutor.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/tasks/SystemTaskExecutor.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.conductor.core.execution.tasks;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.netflix.conductor.core.config.Configuration;
+import com.netflix.conductor.core.execution.WorkflowExecutor;
+import com.netflix.conductor.core.utils.QueueUtils;
+import com.netflix.conductor.core.utils.SemaphoreUtil;
+import com.netflix.conductor.dao.QueueDAO;
+import com.netflix.conductor.metrics.Monitors;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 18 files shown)


core/src/main/java/com/netflix/conductor/core/execution/tasks/SystemTaskExecutor.java
```diff
@@ -0,0 +1,114 @@
...
+
+    void pollAndExecute(String queueName) {
+        // get the remaining capacity of worker queue to prevent queue full exception
+        ExecutionConfig executionConfig = getExecutionConfig(queueName);
+        SemaphoreUtil semaphoreUtil = executionConfig.getSemaphoreUtil();
+        ExecutorService executorService = executionConfig.getExecutorService();
+        String taskName = QueueUtils.getTaskType(queueName);
+
+        if (!semaphoreUtil.canProcess()) {
+            // no available permits, do not poll
```
kishorebanala, line 71: Consider monitoring this to see how often we run into this and potentially leverage this metric to scale out system task workers.

core/src/main/java/com/netflix/conductor/core/execution/tasks/SystemTaskExecutor.java
```diff
@@ -0,0 +1,114 @@
...
+
+                // release permit after processing is complete
+                taskCompletableFuture.whenComplete((r, e) -> semaphoreUtil.completeProcessing());
+            } else {
+                // no task polled, release permit
+                semaphoreUtil.completeProcessing();
+            }
+        } catch (Exception e) {
+            // release the permit if exception is thrown during polling, because the thread would not be busy
+            semaphoreUtil.completeProcessing();
```
kishorebanala, line 95: `semaphoreUtil.completeProcessing();` this can be part of finally block.

core/src/main/java/com/netflix/conductor/core/execution/tasks/SystemTaskWorkerCoordinator.java
```diff
@@ -13,220 +13,116 @@
...
+        return StringUtils.equals(queueExecutionNameSpace, executionNameSpace);
+    }
+
+    private boolean shouldListen(String workflowSystemTaskQueueName) {
+        return isFromCoordinatorExecutionNameSpace(workflowSystemTaskQueueName)
+            && isAsyncSystemTask(workflowSystemTaskQueueName);
+    }
+
+    @VisibleForTesting
+    boolean isAsyncSystemTask(String queue) {
```
kishorebanala, line 120: +1

core/src/main/java/com/netflix/conductor/core/execution/tasks/SystemTaskExecutor.java
apanicker-nflx: If this is put in a finally block, permits will be released twice in case of successful system task execution, once at line 88 (after the future is completed) and then again in finally block.

PR #1646, opened 2020-04-16
Hotfix subworkflow large payload changes

kishorebanala:


core/src/main/java/com/netflix/conductor/core/execution/tasks/SubWorkflow.java
```diff
@@ -40,6 +40,7 @@
 
 	private static final Logger logger = LoggerFactory.getLogger(SubWorkflow.class);
 	public static final String NAME = "SUB_WORKFLOW";
+	public static final String SUB_WORKFLOW_ID = "subWorkflowId";
```
apanicker-nflx, line 43: this can be private

PR #1648, opened 2020-04-20
Check if Cassandra session is closed before using

mactaggart:
In teardown of CassandraDAOTest "session is closed" errors occur. This happens because CassandraMetadataDAO and CassandraEventHandlerDAO launch threads that use session w/o checking if its closed. This PR adds a check to ensure session is not closed before using.

### DIFF
--- a/cassandra-persistence/src/test/java/com/netflix/conductor/dao/cassandra/CassandraDAOTest.java
+++ b/cassandra-persistence/src/test/java/com/netflix/conductor/dao/cassandra/CassandraDAOTest.java
@@ -74,7 +74,6 @@ public class CassandraDAOTest {
 
     @Before
     public void setUp() {
-        session = embeddedCassandra.getSession();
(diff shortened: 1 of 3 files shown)


cassandra-persistence/src/test/java/com/netflix/conductor/dao/cassandra/CassandraDAOTest.java
```diff
@@ -69,6 +69,7 @@
     @BeforeClass
     public static void init() throws Exception {
         embeddedCassandra = new EmbeddedCassandra();
+        session = embeddedCassandra.getSession();
     }
 
     @Before
```
apanicker-nflx, line 75: Please remove `session = embeddedCassandra.getSession();` from the setUp block.

mactaggart: Ok; removed.

Issue #1655, opened 2020-04-22
if loopOver task within DO_WHILE times out and DO_WHILE has retries, doesn't work


rickfish:
If a loopOver task within a DO_WHILE times out, the DO_WHILE fails (FAILED status) - from the documentation, that is expected. Unfortunately if the DO_WHILE has retries available, it is retried but the retried DO_WHILE task immediately comes back as FAILED and its loopOver task gets CANCELED (so not pollable). That is what happens the first time it is retried. After that the additional retries of the DO_WHILE just get set to FAILED with no retry of the loopOver and eventually the workflow fails.

I am trying to debug what is causing this behavior but it is pretty hairy with all the twists and turns. Has anyone seen this before?

rickfish:
I am thinking about creating a pull request that adds a retryCount match when filtering the loopOver tasks to review. The problem seems to be that the current DO_WHILE execute() method looks at all past tasks as well as the current one and it shouldn't. Does anyone have any issues with me changing the stream filter to the below (basically adding a match on retry count)?

``		List<Task> loopOver = workflow.getTasks().stream().filter(t -> 
			(task.getWorkflowTask().has(TaskUtils.removeIterationFromTaskRefName(t.getReferenceTaskName())) 
			&& !task.getReferenceTaskName().equals(t.getReferenceTaskName()) 
			&& task.getRetryCount() == t.getRetryCount())).collect(Collectors.toList());
``

manan164:
Hi @rickfish , Thanks for pointing out.
The idea was when any loopOver task failed we should retry complete iteration from first loopOver task. If this is not happening then its bug. The reason is, if we start from the loopOver task which failed then the condition inside the execute method will be complex. We have to take care of all loopOver tasks retryCount also. Please create PR for the fix.
Also, DO_WHILE task itself is very slow because of applying the filter on all the workflow tasks. Please change the condition if you want.

rickfish:
Thanks @manan164. I just created PR #1657 for this fix. It is really difficult to tell if this will cause any side effects but I don't think it will. DO_WHILE is a tough one to test all scenarios - thanks for taking this on - I'm sure it wasn't an easy one to implement.

PR #1657, opened 2020-04-24
add retrycount to filtering criteria for dowhile task

rickfish:
This is a fix for issue #1655. Retries for the DO_WHILE task itself do not work if one of its loopOver tasks time out and the DO_WHILE fails. This fix only checks the condition of loopOver tasks that have the same retry count as the DO_WHILE task.


apanicker-nflx:
@manan164 Please help reviewing this pull request.

apanicker-nflx:
@rickfish Thank you for the contribution. Please add an integration test to cover this case.

rickfish:
This would be a difficult test case to add as it involves having one of the loopover SIMPLE tasks time out and then polling and completing that task only after it has timed out the first time in the first iteration. Do you think it is worth that? Since it is a bug fix, wouldn't the fact that it doesn't break the existing do-while test cases be enough?

apanicker-nflx:
@rickfish I am merging this PR for now. I am working on converting the DO_WHILE integration tests into scenario based tests and will add a test to cover this case (when a loopOver task is unsuccessful).

rickfish:
@apanicker-nflx thank you. If I wanted to attach the workflow/subworkflow metadata to this PR so you know the specific scenario I use to test this issue, how would I do that? I have tried to use the 'Attach files by dragging : dropping, selecting or pasting them' link but it always gives me an error.

simarpal:
Hi @rickfish,
I have tried your PR.
Do while's output is not showing proper output for each iteration. It is showing the output of the first iteration for every iteration.

rickfish:
@simarpal sorry, I just provided another PR (#1701 ) that should fix this issue.

simarpal:
Thanks @rickfish for quick response, it worked with PR (#1701 ).

simarpal:
I have one more issue (https://github.com/Netflix/conductor/issues/1696). Which is not related to your change. Let me know if you any additional info about it.

rickfish:
So that issue ( #1696 )is all about the UI, right? The DO_WHILE task is working as expected - just not being displayed on the UI, right?

simarpal:
Right, Do while is working as expected just not showing one of the looped tasks properly in UI.

rickfish:
@simarpal sorry, I don't know the UI and I actually don't know the do_while task except where I found issues and tried to fix them. You might want to create an issue that has 'Conductor UI' in the subject and maybe a UI guy can look into it.

simarpal:
Ok
Thanks @rickfish

PR #1664, opened 2020-04-30
Workflow decider not properly scheduling next tasks after a decision

tjasko:
In the event the workflow sweeper runs before tasks get queued by result of a decision task, it's possible for the sweeper to mark the workflow as completed too early. This issue surfaces itself more under high-load, due to the nature of the bug.

Given the following workflow:
```json
[
  {
    "schemaVersion": 2,
    "name": "test-workflow",
    "version": 3,
    "tasks": [
      {
        "type": "SIMPLE",
        "name": "t1",
        "taskReferenceName": "t1"
      },
      {
        "type": "DECISION",
        "name": "d1",
        "taskReferenceName": "has_something",
        "inputParameters": {
          "count": "${t1.output.count}"
        },
        "caseExpression": "$.count > 0 ? 'true' : 'false'",
        "decisionCases": {
          "true": [
            {
              "type": "SIMPLE",
              "name": "t2",
              "taskReferenceName": "t2"
            }
          ]
        }
      }
    ]
  }
]
```

Assuming task `t1` and decision `d1` have already been marked as completed, if the workflow sweeper runs before the decider can queue up task `t2`, the workflow will be marked as completed.

After looking at the code, this makes sense why. The [decision task mapper](https://github.com/Netflix/conductor/blob/v2.26.1/core/src/main/java/com/netflix/conductor/core/execution/mapper/DecisionTaskMapper.java#L108) handles this, however that's never executed when `DeciderService.decide()` runs. This is partly due to `DeciderService.isNonPendingTask` not ever picking up a completed decision task, so it sees that there are no tasks to schedule. After correcting this conditional, additional logic needed to be added to call the `DeciderService.getTasksToBeScheduled()` function, as that's what ultimately computes the next tasks from a decision task.

I've also went ahead and added tests for this case. This was tested locally by slamming the living heck out of Conductor to ensure no more issues.

Please let me know if you need anything else regarding this PR.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
@@ -74,7 +74,8 @@ public class DeciderService {
 
     private final Map<String, TaskMapper> taskMappers;
 
-    private final Predicate<Task> isNonPendingTask = task -> !task.isRetried() && !task.getStatus().equals(SKIPPED) && !task.isExecuted() ||  task.getWorkflowTask() != null && task.getWorkflowTask().getType() == TaskType.DECISION.name();
+    private final Predicate<Task> isNonPendingTask = task -> (!task.isRetried() && !task.getStatus().equals(SKIPPED) && !task.isExecuted()) ||
+            (task.getWorkflowTask() != null && task.getWorkflowTask().getType().equals(TaskType.DECISION.name()));
 
     private static final String PENDING_TASK_TIME_THRESHOLD_PROPERTY_NAME = "workflow.task.pending.time.threshold.minutes";
 
--- a/core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
@@ -74,7 +74,7 @@ public class DeciderService {
 
     private final Map<String, TaskMapper> taskMappers;
 
-    private final Predicate<Task> isNonPendingTask = task -> !task.isRetried() && !task.getStatus().equals(SKIPPED) && !task.isExecuted();
+    private final Predicate<Task> isNonPendingTask = task -> !task.isRetried() && !task.getStatus().equals(SKIPPED) && !task.isExecuted() ||  task.getWorkflowTask() != null && task.getWorkflowTask().getType() == TaskType.DECISION.name();
 
     private static final String PENDING_TASK_TIME_THRESHOLD_PROPERTY_NAME = "workflow.task.pending.time.threshold.minutes";
 
@@ -174,12 +174,20 @@ public class DeciderService {
                 }
             }
 
-            if (!pendingTask.getStatus().isSuccessful()) {
-                WorkflowTask workflowTask = pendingTask.getWorkflowTask();
-                if (workflowTask == null) {
-                    workflowTask = workflow.getWorkflowDefinition().getTaskByRefName(pendingTask.getReferenceTaskName());
-                }
+            WorkflowTask workflowTask = pendingTask.getWorkflowTask();
+            if (workflowTask == null) {
+                workflowTask = workflow.getWorkflowDefinition().getTaskByRefName(pendingTask.getReferenceTaskName());
+            }
 
+            if (workflowTask.getType().equals(TaskType.DECISION.name()) && pendingTask.getStatus() == Status.COMPLETED) {
+                getTasksToBeScheduled(workflow, pendingTask.getWorkflowTask(), 0).forEach(nextTask -> {
+                    tasksToBeScheduled.putIfAbsent(nextTask.getReferenceTaskName(), nextTask);
+                });
+
+                continue;
+            }
+
+            if (!pendingTask.getStatus().isSuccessful()) {
                 Optional<Task> retryTask = retry(taskDefinition.orElse(null), workflowTask, pendingTask, workflow);
                 if (retryTask.isPresent()) {
                     tasksToBeScheduled.put(retryTask.get().getReferenceTaskName(), retryTask.get());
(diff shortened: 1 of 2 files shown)


tjasko:
Not exactly sure why `com.netflix.conductor.contribs.queue.sqs.TestQueueManager > testWithTaskId` is failing, that test is fine locally. 😕

apanicker-nflx:
@tjasko Thanks for the detailed explanation and the pull request.
The build seems to be failing at 
```
com.netflix.conductor.core.execution.TestDeciderService > testFork FAILED
com.netflix.conductor.core.execution.TerminateWorkflowException at testDeciderService.java:775
```

tjasko:
@apanicker-nflx After some investigation, it's failing because that test is doing some odd things. The `test.json` file first needs the workflow task definition:
```
      {
        "name": "perf_task_4",
        "taskReferenceName": "perf_task_4",
        "inputParameters": {
          "mod": "${perf_task_3.output.mod}",
          "oddEven": "${perf_task_3.output.oddEven}"
        },
        "type": "SIMPLE",
        "startDelay": 0,
        "taskDefinition": {
          "name": "perf_task_4",
          "retryCount": 0,
          "timeoutSeconds": 3600,
          "timeoutPolicy": "TIME_OUT_WF",
          "retryLogic": "FIXED",
          "retryDelaySeconds": 0,
          "responseTimeoutSeconds": 3000
        }
      },
```

And `SimpleTaskMapper` needs to be able to pull in the workflow task def:
```
        TaskDef taskDefinition = Optional.ofNullable(taskToSchedule.getTaskDefinition())
            .orElseGet(() -> Optional.ofNullable(workflowInstance.getWorkflowDefinition()
                .getTaskByRefName(taskToSchedule.getTaskReferenceName())
                .getTaskDefinition())
                .orElseThrow(() -> {
                    String reason = String.format("Invalid task. Task %s does not have a definition", taskToSchedule.getName());
                    return new TerminateWorkflowException(reason);
                }));
```

However, when doing that, the test isn't passing as all of the tasks are marked as completed:
```
java.lang.AssertionError: expected:<5> but was:<0>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at org.junit.Assert.assertEquals(Assert.java:631)
	at com.netflix.conductor.core.execution.TestDeciderService.testFork(TestDeciderService.java:713)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
```

I'm pretty sure this wasn't a valid test in the first place, and the lacking of the workflow definition was being improperly this condition. If you checkout the current codebase and only add the updated JSON task def I linked, the test fails.

Would appreciate your input here.

apanicker-nflx:
Thanks for testing this out, I will verify/fix tests in the `dev` branch.

techyragu:
@apanicker-nflx - Is this issue fixed and merged to master? I am also seeing this issue.  I merged the changes from Netflix master on 01-Sep-2020.

techyragu:
@tjasko Have you added these changes in your code base?

tjasko:
@techyragu Yes, been running these changes in production for some time now without issue. The actual commits aren't ready to be merged due to the tests failing, but frankly it appears they were partially incorrect in the first place anyways. I asked the Netflix team to look into a bit for input, but as you can see, that's where this pull request got stalled. I haven't given the changes any more attention since then.

apanicker-nflx:
@tjasko This PR slipped through the cracks. Apologies.

We will look into merging this PR soon.

tjasko:
@apanicker-nflx Appreciate it.

apanicker-nflx:
> If you checkout the current codebase and only add the updated JSON task def I linked, the test fails. 
```json 
"taskDefinition": {
          "name": "perf_task_4",
          "retryCount": 0,
          "timeoutSeconds": 3600,
          "timeoutPolicy": "TIME_OUT_WF",
          "retryLogic": "FIXED",
          "retryDelaySeconds": 0,
          "responseTimeoutSeconds": 3000
        }
```

@tjasko I am unable to reproduce this test failure locally.
Moreover, I stepped through the test execution and this test is correct. I extracted the workflow from `test.json` and ran it in our test environment and was able to verify that the workflow runs as expected. 

>And SimpleTaskMapper needs to be able to pull in the workflow task def

This is not needed, since the mapped task to schedule that is provided to the simple task mapper should already contain the task definition at this point. 

In this particular case however, the forked tasks are all `SUB_WORKFLOW` tasks and do not invoke the SimpleTaskMapper.

apanicker-nflx:
I tried reproducing the issue with the DECISION task in our test environment and am unable to do so. Can you provide some details about your setup - persistence, queue, lock implementation etc. ? Thank you.

apanicker-nflx:
@tjasko Can you provide some details about your setup - persistence, queue, lock implementation etc.  that would help with reproducing this scenario.

tjasko:
@apanicker-nflx The issue when I originally opened this was able to be found with using the in-memory DB. However, in production we run with MySQL persistence. 

In regards to queue/local implementation, that's all whatever's default.

I can't give much input on this issue due to how long it's been, but I remember it happens under high-load circumstances.

apanicker-nflx:
By default, Conductor server does not use a workflow execution lock which seems to be causing this issue. We strongly recommend enabling the workflow execution lock to ensure consistency and prevent any race conditions.  
Please let us know if you continue to run into this issue after enabling the lock.

tjasko:
@apanicker-nflx That's honestly a bit unfortunate, but that's a bit interesting as you can reproduce this issue under just one Conductor server instance. While locking does indeed help solve problems, it really shouldn't be a problem under a single instance.

apanicker-nflx:
>While locking does indeed help solve problems, it really shouldn't be a problem under a single instance.

This is true when it comes to serving requests from multiple instances. But background processes like the sweeper, system task worker, etc do modify the state of the workflow in separate threads which can lead to inconsistencies. These threads can be running on the same instance or separate instances for any given workflow and race conditions could manifest in high-load situations. This was the motivation for the workflow execution lock implementation.

PR #1667, opened 2020-05-03
Fix kitchensink example

zotvent:
Fix kitchensink example:

- set output of the task with `outputData` key
- set `ownerEmail` property to workflow and task definitions since it is mandatory


mactaggart:
It's worth noting that there's a property called workflow.owner.email.mandatory that controls if email address is mandatory for task and workflow definitions

zotvent:
@mactaggart I didn't know about this property. Thank you for telling me about it! From now on I will try to use it.

PR #1671, opened 2020-05-06
Fix the issue with ParametersUtils where input map is mutated

cyzhao:


core/src/main/java/com/netflix/conductor/core/execution/ParametersUtils.java
```diff
@@ -161,23 +161,23 @@ public Object replace(String paramString) {
...
     private Map<String, Object> replace(Map<String, Object> input, DocumentContext documentContext, String taskId) {
+        Map<String, Object> result = new HashMap<>();
         for (Entry<String, Object> e : input.entrySet()) {
+            Object newValue;
             Object value = e.getValue();
             if (value instanceof String) {
-                Object replaced = replaceVariables(value.toString(), documentContext, taskId);
-                e.setValue(replaced);
+                newValue = replaceVariables(value.toString(), documentContext, taskId);
             } else if (value instanceof Map) {
```
cyzhao, line 170: This block of code is where the previous fix falls apart, the new test I added fails without the new fix.

kishorebanala:
Rebased against release `v2.26.2-beta` to expedite for hotfix release.

PR #1675, opened 2020-05-11
Add configurations to enable external index management

apanicker-nflx:
Shade ES6 dependency and added configurations to enable external index management in ES6.

### DIFF
--- a/es6-persistence/src/main/java/com/netflix/conductor/dao/es6/index/ElasticSearchBaseDAO.java
+++ b/es6-persistence/src/main/java/com/netflix/conductor/dao/es6/index/ElasticSearchBaseDAO.java
@@ -1,18 +1,29 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
 package com.netflix.conductor.dao.es6.index;
 
 import com.netflix.conductor.dao.IndexDAO;
 import com.netflix.conductor.dao.es6.index.query.parser.Expression;
 import com.netflix.conductor.elasticsearch.query.parser.ParserException;
+import java.io.IOException;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.elasticsearch.index.query.BoolQueryBuilder;
 import org.elasticsearch.index.query.QueryBuilder;
 import org.elasticsearch.index.query.QueryBuilders;
 import org.elasticsearch.index.query.QueryStringQueryBuilder;
-
-import java.io.IOException;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 abstract class ElasticSearchBaseDAO implements IndexDAO {
 
@@ -45,8 +56,7 @@ abstract class ElasticSearchBaseDAO implements IndexDAO {
         return QueryBuilders.boolQuery().must(stringQuery).must(filterQuery);
     }
 
-    String indexName(String documentType) {
+    protected String getIndexName(String documentType) {
         return indexPrefix + "_" + documentType;
     }
-
 }
--- a/es6-persistence/src/main/java/com/netflix/conductor/dao/es6/index/ElasticSearchDAOV6.java
+++ b/es6-persistence/src/main/java/com/netflix/conductor/dao/es6/index/ElasticSearchDAOV6.java
@@ -1,15 +1,14 @@
 /*
- * Copyright 2016 Netflix, Inc.
+ * Copyright 2020 Netflix, Inc.
  *
- * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
- * in compliance with the License. You may obtain a copy of the License at
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
  *
  * http://www.apache.org/licenses/LICENSE-2.0
  *
- * Unless required by applicable law or agreed to in writing, software distributed under the License
- * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
- * or implied. See the License for the specific language governing permissions and limitations under
- * the License.
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
  */
 package com.netflix.conductor.dao.es6.index;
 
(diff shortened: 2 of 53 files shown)


es6-persistence/src/main/java/com/netflix/conductor/dao/es6/index/ElasticSearchBaseDAO.java
```diff
@@ -45,8 +56,7 @@ BoolQueryBuilder boolQueryBuilder(String expression, String queryString) throws
         return QueryBuilders.boolQuery().must(stringQuery).must(filterQuery);
     }
 
-    String indexName(String documentType) {
+    String getIndexName(String documentType) {
```
AlejandroJorgePerez, line 59: Do you use @Nullable annotations? What is the expected response for bad input?

es6-persistence/src/main/java/com/netflix/conductor/dao/es6/index/ElasticSearchDAOV6.java
```diff
@@ -91,62 +91,45 @@
...
     private static final long KEEP_ALIVE_TIME = 1L;
-
     private static final int UPDATE_REQUEST_RETRY_COUNT = 5;
 
-    private String workflowIndexName;
-
-    private String taskIndexName;
-
-    private String eventIndexPrefix;
+    private static final String className = ElasticSearchDAOV6.class.getSimpleName();
```
AlejandroJorgePerez, line 107: private static final constants, should they be capitalized? I think a better question is do we have code guidelines for Conductor published somewhere?

es6-persistence/src/main/java/com/netflix/conductor/dao/es6/index/ElasticSearchDAOV6.java
```diff
@@ -91,62 +91,45 @@
...
 
+    private final String workflowIndexName;
+    private final String taskIndexName;
+    private final String eventIndexPrefix;
     private String eventIndexName;
-
-    private String messageIndexPrefix;
-
+    private final String messageIndexPrefix;
```
AlejandroJorgePerez, line 113: I see the structure of the indices is more complicated than I suspected. Is that documented somewhere?

es6-persistence/src/main/java/com/netflix/conductor/dao/es6/index/ElasticSearchDAOV6.java
```diff
@@ -156,8 +139,8 @@ public ElasticSearchDAOV6(Client elasticSearchClient, ElasticSearchConfiguration
         this.objectMapper = objectMapper;
         this.elasticSearchClient = elasticSearchClient;
         this.indexPrefix = config.getIndexName();
-        this.workflowIndexName = indexName(WORKFLOW_DOC_TYPE);
-        this.taskIndexName = indexName(TASK_DOC_TYPE);
+        this.workflowIndexName = getIndexName(WORKFLOW_DOC_TYPE);
```
AlejandroJorgePerez, line 142: I suppose there're some legacy reasons behind this, but, shouldn't this doc types be an enum instead of a series of Strings?

es6-persistence/src/main/java/com/netflix/conductor/dao/es6/index/ElasticSearchDAOV6.java
```diff
@@ -187,17 +177,16 @@ public ElasticSearchDAOV6(Client elasticSearchClient, ElasticSearchConfiguration
...
             });
 
         Executors.newSingleThreadScheduledExecutor().scheduleAtFixedRate(this::flushBulkRequests, 60, 30, TimeUnit.SECONDS);
-
     }
 
     @PreDestroy
     private void shutdown() {
-        logger.info("Gracefully shutdown executor service");
+        LOGGER.info("Gracefully shutdown executor service");
```
AlejandroJorgePerez, line 189: Nitpicking on phrasing "Starting graceful shutdown"?

es6-persistence/src/main/java/com/netflix/conductor/dao/es6/index/ElasticSearchDAOV6.java
```diff
@@ -206,13 +195,13 @@ private void shutdownExecutorService(ExecutorService execService) {
...
             } else {
-                logger.warn("Forcing shutdown after waiting for 30 seconds");
+                LOGGER.warn("Forcing shutdown after waiting for 30 seconds");
                 execService.shutdownNow();
             }
         } catch (InterruptedException ie) {
-            logger.warn("Shutdown interrupted, invoking shutdownNow on scheduledThreadPoolExecutor for delay queue");
+            LOGGER.warn("Shutdown interrupted, invoking shutdownNow on scheduledThreadPoolExecutor for delay queue");
             execService.shutdownNow();
             Thread.currentThread().interrupt();
```
AlejandroJorgePerez, line 206: Why do you explicitly call the current thread's interrupt, instead of letting the exception float up? At the very least you're loosing the context of the original exception.

es6-persistence/src/main/java/com/netflix/conductor/dao/es6/index/ElasticSearchDAOV6.java
```diff
@@ -222,12 +211,11 @@ private void shutdownExecutorService(ExecutorService execService) {
     public void setup() throws Exception {
         waitForHealthyCluster();
 
-        createIndexesTemplates();
-
-        createWorkflowIndex();
-
-        createTaskIndex();
-
+        if (config.isElasticSearchAutoIndexManagementEnabled()) {
```
AlejandroJorgePerez, line 214: (Following up on our conversation)
I think this could be refactored to use an in-memory queue and an indexer class when index management is enabled. That way we'll have comparable architecture in both cases (producer -> queue -> indexer)

All this logic would remain the same, but isolated in it's own class, and that class can be re-used from the indexer service, when you decide your deployment requires an independent service.

es6-persistence/src/main/java/com/netflix/conductor/dao/es6/index/ElasticSearchDAOV6.java
```diff
@@ -310,20 +298,20 @@ private void createIndex(String indexName) {
...
         GetMappingsResponse getMappingsResponse = elasticSearchClient.admin().indices().prepareGetMappings(indexName).addTypes(type).execute().actionGet();
         if (getMappingsResponse.mappings().isEmpty()) {
-            logger.info("Adding the {} type mappings", indexName);
+            LOGGER.info("Adding the {} type mappings", indexName);
             try {
                 String source = loadTypeMappingSource(sourcePath);
                 elasticSearchClient.admin().indices().preparePutMapping(indexName).setType(type).setSource(source, XContentType.JSON).execute().actionGet();
             } catch (Exception e) {
-                logger.error("Failed to init index " + indexName + " mappings", e);
+                LOGGER.error("Failed to init index " + indexName + " mappings", e);
```
AlejandroJorgePerez, line 314: When you get this error, doesn't the whole system end up in a bad state? Here you're just logging and continuing on, but something really bad happened. Does it make sense to let the process go on?

es6-persistence/src/main/java/com/netflix/conductor/elasticsearch/SystemPropertiesElasticSearchConfiguration.java
```diff
@@ -1,7 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
 package com.netflix.conductor.elasticsearch;
 
 import com.netflix.conductor.core.config.SystemPropertiesConfiguration;
 
 public class SystemPropertiesElasticSearchConfiguration
```
AlejandroJorgePerez, line 17: Javadoc? :)

es6-persistence/src/main/java/com/netflix/conductor/dao/es6/index/ElasticSearchDAOV6.java
apanicker-nflx: In ES6, each index corresponds to one document type.

es6-persistence/src/main/java/com/netflix/conductor/dao/es6/index/ElasticSearchDAOV6.java
apanicker-nflx: will refactor in a subsequent PR.

PR #1676, opened 2020-05-11
Initialize workflow monitor

mdepak:
Monitor for calculating metrics `com.netflix.conductor.service.WorkflowMonitor`is not initialized and so metrics like `task_queue_depth`, `workflow_running` are not captured. So initializing it in `WorkflowExecutorModule`.


core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutorModule.java
```diff
@@ -30,5 +31,6 @@ protected void configure() {
         bind(TaskService.class).to(TaskServiceImpl.class);
         bind(EventService.class).to(EventServiceImpl.class);
         bind(MetadataService.class).to(MetadataServiceImpl.class);
+        bind(WorkflowMonitor.class).asEagerSingleton();
```
apanicker-nflx, line 34: Please move this binding to the `ServerModule` instead.

PR #1674, opened 2020-05-08
doc: update and synchronise doc for Event, Http and Sub workflow tasks

Jiehong:

PR #1686, opened 2020-05-15
doc: remove workflowInput from Sub workflow tasks documentation

Jiehong:
@elisherer: thanks for spotting it in https://github.com/Netflix/conductor/pull/1674.


elisherer:
@Jiehong, No problem,
Also notice the table seems to be broken
image
https://netflix.github.io/conductor/configuration/systask/#sub-workflow
(seems ok on the markdown though)

Jiehong:
@elisherer : thanks for seeing it!
Fixed!

PR #1692, opened 2020-05-19
Decision integration test

pctreddy:


test-harness/src/test/groovy/com/netflix/counductor/integration/test/DecisionTaskSpec.groovy
```diff
@@ -0,0 +1,365 @@
+package com.netflix.counductor.integration.test
```
apanicker-nflx, line 1: Please add license header

PR #1719, opened 2020-06-02
Support to batch poll system tasks

mdepak:
Using batch polling for system tasks from the queues. It will be very helpful for high throughput scenarios where large number of system tasks are executed and can possibly reduce the i/o (number of lookups in queues). It can be controlled by `workflow.system.task.queue.pollCount` configuration.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/utils/SemaphoreUtil.java
+++ b/core/src/main/java/com/netflix/conductor/core/utils/SemaphoreUtil.java
@@ -62,4 +62,17 @@ public class SemaphoreUtil {
         LOGGER.debug("Number of available permits: {}", available);
         return available;
     }
+
+    public boolean acquireSlots(int numSlots)
+    {
+        return semaphore.tryAcquire(numSlots);
+    }
+
+    /**
+     * Signals that processing is complete and the permit can be released.
+     */
+    public void completeProcessing(int numSlots) {
+        LOGGER.debug("Completed execution; releasing permit");
+        semaphore.release(numSlots);
+    }
 }
--- a/core/src/main/java/com/netflix/conductor/core/utils/SemaphoreUtil.java
+++ b/core/src/main/java/com/netflix/conductor/core/utils/SemaphoreUtil.java
@@ -16,6 +16,7 @@
 package com.netflix.conductor.core.utils;
 
 import java.util.concurrent.Semaphore;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/netflix/conductor/core/utils/SemaphoreUtil.java
```diff
@@ -62,4 +63,23 @@ public int availableSlots() {
...
     }
+
+    /**
+     * Signals if processing is allowed based on whether specified number of permits can be acquired.
+     *
+     * @param numSlots the number of permits to acquire
+     * @return {@code true} - if permit is acquired
+     * {@code false} - if permit could not be acquired
+     */
+    public boolean acquireSlots(int numSlots) {
```
apanicker-nflx, line 74: Does this need to be a separate method? This could be rolled into the methods defined above where numSlots=1.

core/src/main/java/com/netflix/conductor/core/utils/SemaphoreUtil.java
```diff
@@ -62,4 +63,23 @@ public int availableSlots() {
...
+        return semaphore.tryAcquire(numSlots);
+    }
+
+    /**
+     * Signals that processing is complete and the specified number of permits can be released.
+     */
+    public void completeProcessing(int numSlots) {
+        LOGGER.debug("Completed execution; releasing permit");
```
apanicker-nflx, line 82: same as above.

core/src/main/java/com/netflix/conductor/core/utils/SemaphoreUtil.java
```diff
@@ -62,4 +63,23 @@ public int availableSlots() {
...
+     * {@code false} - if permit could not be acquired
+     */
+    public boolean acquireSlots(int numSlots) {
+        return semaphore.tryAcquire(numSlots);
+    }
+
+    /**
+     * Signals that processing is complete and the specified number of permits can be released.
+     */
+    public void completeProcessing(int numSlots) {
```
apanicker-nflx, line 81: same as above.

apanicker-nflx:
Please add a test to cover this.

core/src/main/java/com/netflix/conductor/core/utils/SemaphoreUtil.java
mdepak: Thanks for the suggestion. Using the extended method with numSlots=1 now.

core/src/main/java/com/netflix/conductor/core/utils/SemaphoreUtil.java
mdepak: Fixed it.

mdepak:
Added a unit test to verify the batch poll scenario.

PR #1720, opened 2020-06-03
Optional event execution indexing

mdepak:
Please consider making event execution indexing optional. Since it is not used for any logic directly, it can be made optional and can help reduce i/o.


apanicker-nflx:
Please fix the build failure.

mdepak:
Build failure is fixed and please check when you get a chance.

PR #1722, opened 2020-06-03
Monitor additional metrics

mdepak:
Please consider adding these additional metrics for monitoring the number of workflows/tasks/system tasks completed as it will be helpful to monitor the performance.


core/src/main/java/com/netflix/conductor/metrics/Monitors.java
```diff
@@ -306,7 +306,19 @@ public static void recordAcquireLockFailure(String exceptionClassName) {
...
-	public static void recordSystemTaskWorkerPollingLimited(String queueName) {
+	public static void recordWorkflowCompleted(String workflowType) {
+		counter(classQualifier, "workflow_completed", "workflowName", workflowType);
+	}
+
+	public static void recordTaskCompleted(String taskType) {
+		counter(classQualifier, "task_completed", "taskType", taskType);
+	}
+
+	public static void recordSystemTaskCompleted(String taskType) {
```
kishorebanala, line 317: How about adding tags to `task_completed` metric instead?

core/src/main/java/com/netflix/conductor/metrics/Monitors.java
```diff
@@ -306,7 +306,19 @@ public static void recordAcquireLockFailure(String exceptionClassName) {
 		counter(classQualifier, "acquire_lock_failure", "exceptionType", exceptionClassName);
 	}
 
-	public static void recordSystemTaskWorkerPollingLimited(String queueName) {
+	public static void recordWorkflowCompleted(String workflowType) {
```
kishorebanala, line 309: We already have a Timer metric with `recordWorkflowCompletion`, which we traditionally convert to `counter` metric on the fly when needed. Can you provide more details about the need for a `counter` metric along with `timer` metric please.

kishorebanala:
Hey @mdepak , bumping this up to see if you're still planning to work on this feature.

core/src/main/java/com/netflix/conductor/metrics/Monitors.java
mdepak: Adding tasks as separate tags will be beneficial for aggregation purposes as well.

mdepak:
@kishorebanala  I am still planning to work on this feature and interested in these metrics. `Timer `(`workflow_execution`)'s count expansion does not exactly match the proposed `Counter` (`workflow_completed `) and still wondering how can these values differ by considerable numbers.

kishorebanala:
Hey @mdepak, Just to confirm, are the `workflow_execution` and `workflow_completed `'s count metrics being compared at same cadence? Say per second?

mdepak:
In our setup, Codahale metrics 3.0.1 registry collects metrics and [datadag metrics reporter](https://github.com/coursera/metrics-datadog) is used for uploading metrics to Datadog. I use `count` expansion from the `Timer` meter to track the counts at the same cadence but there seem to be some inconsistencies. I suspect the problem could be due to the old version of the Codahale library we use or the metrics collection system. In this PR, I will remove `workflow_completed` and continue my investigation related to metric discrepency in the background.

kishorebanala:
@mdepak Sounds good, please report your findings as you learn more.

mdepak:
@kishorebanala Updated the PR to just include task-related metrics and I will update my learnings soon.

kishorebanala:
I'm marking this as closed, but please feel free to reopen if this is still relevant.

PR #1723, opened 2020-06-03
Extend archive module to support TTL and delay

mdepak:
Added new workflow status listener called `ArchivingWithTTLWorkflowStatusListener` that supports delayed archiving and setting time-to-live for the documents in Redis for now (can be extended to Cassandra later). 

`Delayed Archiving`: It is required in cases where short workflows are indexed after a delay and `updateIndex` will fail in those cases since the document might not exist. Further, when using async indexing, in case of high throughput scenarios, an asynchronous create document index request could be pending in the `indexQueue`, and the synchronous update request will fail due to missing document. 

`Time-to-live`: It is helpful in the case of `TERMINATED` workflows. If the parent workflow is terminated and then sub-workflow is canceled. In those cases, the task worker could still be running the task and when it completes and try to update the status of the task, the task/workflow will be missing in persistence if the workflow is archived immediately after the termination/completion.  So TTL can help in this case.  Also if the parent-workflow need to lookup output from sub-workflow, using TTL can help the data to be available in the memory(Redis case) rather than disk lookup in elasticsearch.

### DIFF
--- a/contribs/src/main/java/com/netflix/conductor/contribs/listener/ArchivingWithTTLWorkflowStatusListener.java
+++ b/contribs/src/main/java/com/netflix/conductor/contribs/listener/ArchivingWithTTLWorkflowStatusListener.java
@@ -91,23 +91,27 @@ public class ArchivingWithTTLWorkflowStatusListener implements WorkflowStatusLis
     }
 
     private class DelayArchiveWorkflow implements Runnable {
-        private final Workflow workflow;
+        private final String workflowId;
+        private final String workflowName;
+        private final Workflow.WorkflowStatus status;
         private final ExecutionDAOFacade executionDAOFacade;
 
         DelayArchiveWorkflow(Workflow workflow, ExecutionDAOFacade executionDAOFacade) {
-            this.workflow = workflow;
+            this.workflowId = workflow.getWorkflowId();
+            this.workflowName = workflow.getWorkflowName();
+            this.status = workflow.getStatus();
             this.executionDAOFacade = executionDAOFacade;
         }
 
         @Override
         public void run() {
             try {
-                this.executionDAOFacade.removeWorkflowWithExpiry(workflow.getWorkflowId(), true, archiveTTLSeconds);
-                LOGGER.info("Archived workflow {}", workflow.getWorkflowId());
-                Monitors.recordWorkflowArchived(workflow.getWorkflowName(), workflow.getStatus());
+                this.executionDAOFacade.removeWorkflowWithExpiry(workflowId, true, archiveTTLSeconds);
+                LOGGER.info("Archived workflow {}", workflowId);
+                Monitors.recordWorkflowArchived(workflowName, status);
                 Monitors.recordArchivalDelayQueueSize(scheduledThreadPoolExecutor.getQueue().size());
             } catch (Exception e) {
-                LOGGER.error("Unable to archive workflow: {}", workflow.getWorkflowId(), e);
+                LOGGER.error("Unable to archive workflow: {}", workflowId, e);
             }
         }
     }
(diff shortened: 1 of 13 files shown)


mdepak:
@apanicker-nflx Can you please review this? This PR is expected to solve problems related to a couple of use cases in archiving workflows. Thanks for your help!

core/src/main/java/com/netflix/conductor/core/config/Configuration.java
```diff
@@ -105,6 +105,14 @@
     String EVENT_EXECUTION_PERSISTENCE_TTL_SECS_PROPERTY_NAME = "workflow.event.execution.persistence.ttl.seconds";
     int EVENT_EXECUTION_PERSISTENCE_TTL_SECS_DEFAULT_VALUE = 0;
 
+    String WORKFLOW_ARCHIVAL_TTL_SECS_PROPERTY_NAME = "workflow.archival.ttl.seconds";
+    int WORKFLOW_ARCHIVAL_TTL_SECS_DEFAULT_VALUE = 0;
+
+    String WORKFLOW_ARCHIVAL_DELAY_SECS_PROPERTY_NAME = "workflow.archival.delay.seconds";
```
apanicker-nflx, line 111: Exposing a configuration without an ability to actually configure sounds very counter-intuitive. I believe that it would be better to just use the async update delay property directly.

core/src/main/java/com/netflix/conductor/core/config/Configuration.java
```diff
@@ -317,6 +325,28 @@ default String getElasticSearchDocumentTypeOverride() {
...
+     * @return The time to live in seconds for workflow archiving module. Currently, only RedisExecutionDAO supports it.
+     */
+    default int getWorkflowArchivalTTL() {
+        return getIntProperty(WORKFLOW_ARCHIVAL_TTL_SECS_PROPERTY_NAME, WORKFLOW_ARCHIVAL_TTL_SECS_DEFAULT_VALUE);
+    }
+
+    /**
+     * @return the time to delay the archival of workflow
+     */
+    default int getWorkflowArchivalDelay() {
```
apanicker-nflx, line 338: Consider using `getAsyncUpdateDelay` directly over this indirection to reduce confusion.

contribs/src/main/java/com/netflix/conductor/contribs/listener/ArchivingWithTTLWorkflowStatusListener.java
```diff
@@ -0,0 +1,114 @@
...
+        if (delayArchiveSeconds > 0) {
+            scheduledThreadPoolExecutor.schedule(new DelayArchiveWorkflow(workflow, executionDAOFacade), delayArchiveSeconds, TimeUnit.SECONDS);
+        } else {
+            this.executionDAOFacade.removeWorkflowWithExpiry(workflow.getWorkflowId(), true, archiveTTLSeconds);
+            Monitors.recordWorkflowArchived(workflow.getWorkflowName(), workflow.getStatus());
+        }
+    }
+
+    private class DelayArchiveWorkflow implements Runnable {
+        private final Workflow workflow;
```
apanicker-nflx, line 94: Since the removal of the workflow only uses the workflowId in the `run` method below in line 105, the object size can be reduced tremendously by only saving the worklfowId as part of the instance.

mdepak: Thanks for the suggestion. This will significantly reduce the memory consumption of the archival delay queue.

core/src/main/java/com/netflix/conductor/core/config/Configuration.java
mdepak: Property `workflow.archival.delay.seconds` can be configured and it takes `getAsyncUpdateDelay()` as default value. In an ideal scenario, where the index delay queue size is near zero, this property can be same value as `getAsyncUpdateDelay()`. When the index delay queue increases due to the unavailability of worker threads/ES throttles then indexes might not be created for updating and so in those cases, this property can provide some buffer time for delays.

PR #1728, opened 2020-06-05
LocalOnly sharding strategy for dyno queues

marosmars:
Allow no distribution of workload in dyno queues.
By default tasks are distributed across all conductor instances using
dyno queue shards.
Now it is possible to only queue workfload for conductor-server that
started the workflow.

More in docs.


marosmars:
A little background:

We are developing a simple active-standby model for conductor, where only active conductor will run workflows and we also wanted only active conductor and attached workers to be able to run tasks.

kishorebanala:
@marosmars Copying over my comment from Gitter for context:

> Thanks for the great notes @marosmars . However, concurrent decide can still happen in same node. From your example, if both the Http task and Sweeper were to run on same instance at about same time (assuming both the queues are in shard ".c"), we'd still run into inconsistencies with concurrent decides. LMK if you have any questions or comments.

marosmars:
@kishorebanala  thx, copying my reply

> Hey @kishorebanala , thx for the review. The reason for the above is that there is only a NoopLocking strategy utilized by default (in ExecutionLockService). Correct ?
>
> Which means that the solution I proposed still requires a real locking strategy. However it would be sufficient to have a localOnly locking strategy (something based on pure Java semaphore) that does not require persistence... As long as I use the localOnly distribution strategy with it.
>
> What I will do is:
>
> - Add a localOnly locking provider (that can handle locking inside a single node)
> - Update the docs
>
> What do you think ?

I didn't realize the default locking is implemented by NoopLock...

redis-persistence/src/main/java/com/netflix/conductor/dao/dynomite/queue/DynoQueueDAO.java
```diff
@@ -24,6 +24,8 @@
 import com.netflix.dyno.queues.ShardSupplier;
 import com.netflix.dyno.queues.redis.RedisDynoQueue;
 import com.netflix.dyno.queues.redis.RedisQueues;
+import com.netflix.dyno.queues.redis.sharding.RoundRobinStrategy;
```
kishorebanala, line 27: Unused imports?

redis-persistence/src/main/java/com/netflix/conductor/dyno/RedisQueuesShardingStrategyProvider.java
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright 2016 Netflix, Inc.
```
kishorebanala, line 2: Nit: Please use the latest copyright.

kishorebanala:
Hey @marosmars,  Looks good to me. Please address the minor comments, and we should be good to merge this PR.

marosmars:
@kishorebanala rebased and fixed the comments. 

One additional question: Should we want make the LocalOnlyLock default ? Instead of Noop ?

core/src/main/java/com/netflix/conductor/core/utils/LocalOnlyLock.java
```diff
@@ -0,0 +1,97 @@
+/*
+ * Copyright (c) 2019 Netflix, Inc.
```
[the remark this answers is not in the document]
marosmars: Used the latest 2020 copyright header I found

redis-persistence/src/main/java/com/netflix/conductor/dao/dynomite/queue/DynoQueueDAO.java
marosmars: Done

redis-persistence/src/main/java/com/netflix/conductor/dyno/RedisQueuesShardingStrategyProvider.java
marosmars: Used the latest 2020 copyright header I found

kishorebanala:
> @kishorebanala rebased and fixed the comments.
> 
> One additional question: Should we want make the LocalOnlyLock default ? Instead of Noop ?

@marosmars Thank you! Actually, let's keep the Noop lock default, as locking is opt-in.

PR #1733, opened 2020-06-09
Fixed bug in gulp file, that results in docker build failures.

kishorebanala:


ui/package-lock.json
```diff
@@ -1726,7 +1726,7 @@
     },
     "babel-register": {
       "version": "6.26.0",
-      "resolved": "https://registry.npmjs.org/babel-register/-/babel-register-6.26.0.tgz",
+      "resolved": "https://artifacts.netflix.com/api/npm/npm-netflix/babel-register/-/babel-register-6.26.0.tgz",
```
apanicker-nflx, line 1729: Is this artifact available publicly outside of Netlix?

ui/package-lock.json
```diff
@@ -8010,7 +7990,7 @@
     },
     "home-or-tmp": {
       "version": "2.0.0",
-      "resolved": "https://registry.npmjs.org/home-or-tmp/-/home-or-tmp-2.0.0.tgz",
+      "resolved": "https://artifacts.netflix.com/api/npm/npm-netflix/home-or-tmp/-/home-or-tmp-2.0.0.tgz",
```
apanicker-nflx, line 7993: same as above

ui/package-lock.json
kishorebanala: Good catch. It should point to `registry.npmjs.org`. I'll revert locks update.

PR #1734, opened 2020-06-09
fix BETWEEN operator; ensure task cancellation when terminating workflow

apanicker-nflx:
- Fix `BETWEEN` operator
Fix the parsing of the operator

- Ensure task is cancelled when terminating a workflow
Remove tasks from the queue first so that polling for tasks will not retrieve these
Update task status to CANCELED 
Update workflow status to TERMINATED

### DIFF
--- a/es5-persistence/src/main/java/com/netflix/conductor/elasticsearch/query/parser/ComparisonOp.java
+++ b/es5-persistence/src/main/java/com/netflix/conductor/elasticsearch/query/parser/ComparisonOp.java
@@ -1,20 +1,14 @@
-/**
- * Copyright 2016 Netflix, Inc.
+/*
+ * Copyright 2020 Netflix, Inc.
  *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ * http://www.apache.org/licenses/LICENSE-2.0
  *
- * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


es5-persistence/src/main/java/com/netflix/conductor/elasticsearch/query/parser/ComparisonOp.java
```diff
@@ -70,7 +62,7 @@ protected void _parse() throws Exception {
 			this.value = "IS";
 		}else if(peeked[0] == '!' && peeked[1] == '='){
 			this.value = "!=";
-		}else if(peeked.length == betwnLen && new String(peeked).equals(Operators.BETWEEN.value())) {
+		}else if(peeked[0] == 'B' && peeked[1] == 'E' && peeked[2] == 'T' && peeked[3] == 'W' && peeked[4] == 'E' && peeked[5] == 'E' && peeked[6] == 'N'){
```
kishorebanala, line 65: Couple of things:
1. Now that we're not checking length of peeked, could it possibly throw ArrayIndexOutOfBounds exception?
2. Given that `STARTS_WITH` below still checks with String equals, what was the case for checking each character by element?

apanicker-nflx: Ahh good catch on the `ArrayIndexOutOfBounds` exception. Pushed out a fix for this.

apanicker-nflx: The parse method always peeks the `maxOperatorLength` number of characters. After the `STARTS_WITH` operator was added, the length of this operator is now the max length (11). Now consider the search query excerpt `startTime BETWEEN 100 AND 200`. In this case, the parse method would now parse the following string `BETWEEN 100` which is of length 11. This would fail the String equals comparison and the BETWEEN operator is now non-functional.

PR #1355, opened 2019-10-15
MySQL queues: reduce the impact of racing by popping the queue 1 by 1

ggrekhov:
As I understand, the current implementation is done in this way (everything or nothing) just because there is no way to get the updated row IDs from a MySQL response. And therefore the only consistent behavior is to completely cancel the partial update as it conflicts with another worker.

But it hurts a lot if there is a huge queue of tasks and a lot of workers trying to poll - they get stuck in a deadlock trying to do `set popped = true` for a bunch of messages.

In this change the messages are marker as popped one by one for each row and the popped ones are returned.

**EDIT**
Closes #577

PR #1741, opened 2020-06-17
Postgres queues: reduce the impact of racing by popping the queue 1 by 1

rickfish:
I like the new logic from PR #1355 and the postgres_persistence module can benefit from the same changes.


apanicker-nflx:
@mactaggart Can you please help review this PR? Thanks

PR #1748, opened 2020-06-24
Use RxExecutionScheduler instead of default RxComputation Scheduler for polling events

mdepak:
As per official [doc](http://reactivex.io/documentation/scheduler.html), RxComputationScheduler is recommended to be used for short CPU bound tasks and the number of threads in this scheduler is bound to the CPU cores and hence it is not ideal for polling messages from different queues. 

RxJava offers an Executor Scheduler where one can control the number of threads to be used for the polling of the events. RxJava's I/O scheduler is also applicable for this use case, but in case of any unavailability of queues it might end up creating more threads causing OOM. 

The intended change is to use the singleton Executor Scheduler instance that is will be used across different queues. The executor thread count is controlled by  `worklfow.event.queue.scheduler.poll.thread.count` property.  We tried this in our setup and helped scale event processing to an extent.


core/src/main/java/com/netflix/conductor/metrics/Monitors.java
```diff
@@ -309,4 +309,8 @@ public static void recordAcquireLockFailure(String exceptionClassName) {
 	public static void recordSystemTaskWorkerPollingLimited(String queueName) {
 		counter(classQualifier, "system_task_worker_polling_limited", "queueName", queueName);
 	}
+
+	public static void recordQueuePollSize(String queueType, int val) {
```
apanicker-nflx, line 313: Please rename method to `recordEventQueuePollSize` to help with maintainability.

apanicker-nflx:
Please resolve conflict and rebase against latest dev.

mdepak:
@apanicker-nflx  Conflicts are resolved. Please merge when you get a chance.

PR #1755, opened 2020-06-25
HealthCheck Module implementation for Redis/ElasticSearch persistences

MikeTangoEcho:
*This is currently a work in progress, i suggest it to be a feature or an example in the documentation depending of the broad usage of people since "healthy" may not mean the same thing for everyone*

**Context:** We use conductor in k8s env and we wanted a single endpoint to check the readiness/liveness of conductor and the stack built around it. The aim is to prevent access to conductor if part of the stack are degraded and raise an event with a readable error.

**Process:** I saw that the route `/api/health` was exposed but didn't find any implementation or mention in the documentation. Then i found health-api doc on https://github.com/Netflix/runtime-health and started to implement two modules for the main persistence.

So feel free to give me your feedback on package naming or HealthIndicator's implementations !

### DIFF
--- a/es-health/README.md
+++ b/es-health/README.md
@@ -1,0 +1,112 @@
+# Elastic Search Health Check 
+
+This module is optional and can compile with `es5-persistence` or `es6-persistence`.
+
+**!! Only tested with es5**
+
+## Build
+
+### Gradle Configuration
+
+Modify the following files
+
+[settings.gradle](https://github.com/Netflix/conductor/blob/master/settings.gradle)
+
+```diff
+@@ -3,6 +3,7 @@ rootProject.name='conductor'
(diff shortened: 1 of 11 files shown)


MikeTangoEcho:
@apanicker-nflx Added tests. Do you think it is better to move out `com.netflix.conductor.server.resources.HealthCheckResource` from `jersey` module into a dedicated module ?

MikeTangoEcho:
HealthCheck is moving to a Spring component so no need for the `com.netflix.runtime.health.api` implementation.

https://github.com/Netflix/conductor/blob/5d874b878d4c1a10bad6de7278e663096771e3ec/rest/src/main/java/com/netflix/conductor/rest/controllers/HealthCheckResource.java#L23-L32

Issue #1752, opened 2020-06-25
Workflow started one task twice


kushalhalder:
I have a workflow that started the same sub-workflow twice in the same moment. How do we fix this? and Now my workflow is stuck, how do we move on from here?

apanicker-nflx:
@kushalhalder Are you using  a locking module in your deployment. If not, in certain high-throughput scenarios, it is possible that the state evaluation of a workflow may be run from two separate threads leading to an inconsistency.

kushalhalder:
Hey, thanks for the reply. I am not explicitly using a locking module. Is something there in the documentation?

apanicker-nflx:
Seems like we need to improve the documentation about this. But you can refer the [code](https://github.com/Netflix/conductor/blob/master/server/src/main/java/com/netflix/conductor/bootstrap/ModulesProvider.java#L144-L174) for implementation/configuration. There is also documentation regarding a newly added locking implementation available [here](https://netflix.github.io/conductor/technicaldetails/#maintaining-workflow-consistency-with-distributed-locking-and-fencing-tokens)

HenryLauu:
would you pls give me more details about fixing this problem? @apanicker-nflx

HenryLauu:
hi @apanicker-nflx I am using zookeeper as lock service and this issue happened again and again , would you pls give me some advice

HenryLauu:
> Hey, thanks for the reply. I am not explicitly using a locking module. Is something there in the documentation?

hi @kushalhalder , would you pls kindly tell me that how did you fix this problem?

kushalhalder:
Hi @HenryLauu, I have not worked on this yet. But anyway, how are you sure that the locks are made in zookeeper?

HenryLauu:
> Hi @HenryLauu, I have not worked on this yet. But anyway, how are you sure that the locks are made in zookeeper?

I config conductor sever something like :
`workflow.decider.locking.enabled=true

    workflow.decider.locking.server=zookeeper

    workflow.zookeeper.lock.connection=zookeeper:2181

    workflow.decider.locking.namespace=conductor`

kushalhalder:
When you query zookeeper, can you see any data?

HenryLauu:
yes something like _c_1e0c3655-1df3-4766-9b8e-df8b4b9656ed-lock-0000000005

PR #1764, opened 2020-07-04
Remove event execution history when removing workflow

isDoing:
When deleting workflow history, there was an issue where EventExecution could not be deleted in persistence.

ex > conductor.test.EVENT_EXECUTION.{event_handler}.{event_name}.{event_message_id}

Check if event task is within workflow and modify it to delete eventExecution.


apanicker-nflx:
@isDoing The event execution need not be deleted when a workflow is removed. Even if the workflow is removed, the event was successfully published which may have had an effect causing it to trigger some downstream processing. Removing event execution need not be tied together with removing the workflow. Please refer #1752 to see if this fits your needs for cleaning up event execution persistence.

isDoing:
@apanicker-nflx 
Thank you for review
There is an issue where the execution history is not automatically cleared after executing the current eventExecution. Is there any way to improve it?

I'm trying to develop eventExecution History Delete api, so please check if there's any other way.

apanicker-nflx:
@isDoing Please refer the implementation in #1572 which adds a configuration for automatically cleaning up event executions from the persistence layer.

Issue #1681, opened 2020-05-12
/workflow/<wfname>/correlated/<corrid>VERY SLOW if many workflows with that corr id


rickfish:
The ExecutionDAO.getWorkflowsByCorrelationId() method has only correlation id as a parameter. There is no additional method that also passes in the workflow name. So on the /workflow/wfname/correlated/corrid endpoint there is no way to do a query that joins in the workflow_def_to_workflow table to get back just the workflows for that workflow name that have that correlation id. This is a problem, especially if there are several records in the workflow table that have the same correlation id, even though they are for different workflow names. 

The existing logic has to query the workflow table for all workflows with that correlation id, then get each workflow individually. If includeTasks is true all the tasks will be retrieved too. Only then do the results get filtered by workflow name.

In our case, the correlation id was '3' and there were 11,945 workflows with that correlation id but only 19 were for the workflow name specified. It took more than a minute to do that in our database that has millions of rows.

I would propose that an additional getWorkflowsByCorrelationId() method is added to all the ExecutionDAO-derived classes that takes the workflow name as a parameter. This would mean that all implementations would have to be changed. Is this something I should attempt to do? Is there a reason why this was not done?

apanicker-nflx:
@rickfish I believe that this can be accomplished using the `/workflow/search` endpoint. Both the correlationId and workflowType are indexed fields and can be used for this purpose. Would this suffice your use-case?

rickfish:
Yes, this can be accomplished with the /workflow/search endpoint. But that is like saying don't use the /workflow//correlated/ endpoint because it could be really slow. At that point, that endpoint would not be used so why even have it? That endpoint should be optimized I think. Since the workflow name is part of the endpoint path, why not pass it into the method that is supposed to retrieve those workflows?

apanicker-nflx:
@rickfish Ahh I misunderstood the earlier comment. I believe the method implementation needs to be changed to retrieve only the correlated workflows of a specific type. I have labeled this as a bug and we will look into fixing this API soon. Thanks for reporting.

apanicker-nflx:
Fixed in #1766

PR #1766, opened 2020-07-06
Bugfix/correlation id performance fix for issue #1681

rickfish:
This is a performance improvement in searching for workflows by correlation id. It adds a method to the ExecutionDAO interface that takes the workflow name so that implementations of ExecutionDAO can use it to join in other tables in order to get a smaller subset of workflows that only contain the ones with that workflow name.

This is a fix for issue #1681.


apanicker-nflx:
@rickfish Thank you very much for the contribution.
Given that the method `getWorkflowsByCorrelationId` in ExecutionDAOFacade is only being invoked from the ExecutionService, and in turn the `getWorkflowsByCorrelationId` method in ExecutionDAO is only being invoked from the ExecutionDAOFacade, I would vote to modify these method implementations with your changes rather than introduce a new method. Please let me know if you think otherwise.

rickfish:
Agreed. I just removed the getWorkflowsByCorrelationId methods that don't take workflowName as a parameter.

core/src/main/java/com/netflix/conductor/service/ExecutionService.java
```diff
@@ -342,14 +343,21 @@ private boolean requeue(Task pending) {
...
 		List<Workflow> result = new LinkedList<>();
-		for (Workflow wf : workflows) {
-			if (wf.getWorkflowName().equals(workflowName) && (includeClosed || wf.getStatus().equals(Workflow.WorkflowStatus.RUNNING))) {
-				result.add(wf);
+		result = workflows.stream().parallel().filter(wf -> {
+			if (includeClosed || wf.getStatus().equals(Workflow.WorkflowStatus.RUNNING)) {
+				// including tasks for subset of workflows to increase performance
+				if (includeTasks) {
+					List<Task> tasks = executionDAOFacade.getTasksForWorkflow(wf.getWorkflowId());
+					tasks.sort(Comparator.comparingLong(Task::getScheduledTime).thenComparingInt(Task::getSeq));
```
apanicker-nflx, line 353: `Comparator.comparingLong(Task::getScheduledTime)` is not needed here since sorting by `seq` here would arrange the tasks in the correct order.

rickfish:
Good point. I removed it.

PR #1770, opened 2020-07-09
fix subworkflow task def use within fork join

apanicker-nflx:
- Honor task definition for subworkflow task when used within a fork/join
- SUB_WORKFLOW task should not response timeout


test-harness/src/test/groovy/com/netflix/counductor/integration/test/ForkJoinSpec.groovy
```diff
@@ -854,4 +858,181 @@ class ForkJoinSpec extends Specification {
         }
     }
 
+    def "Test fork join with sub workflow task using task definition"() {
```
kishorebanala, line 861: +1

PR #1789, opened 2020-07-16
UI LDAP Auth - Add support for nested security groups

elisherer:
Currently the groups returned from the LDAP server are only the direct memberships of the user.
Big enterprises usually use nested groups to allow easier creation of new employees.

I modified the ldap initialization script based on [this](https://stackoverflow.com/questions/51270427/passport-ldapauth-get-nested-groups)


elisherer:
@apanicker-nflx , Hi, can this be merged into next version?

PR #1793, opened 2020-07-17
fix potential problem of database sequences maxing out

rickfish:
The poll_data table is one example of many mysql and postgres tables that have an id column that is generated from a sequence. Most times these tables will be updated, inserts will only happen the first time the resource is inserted, after that many updates will be done. The current sql does an INSERT/ON CONFLICT for postgres and an INSERT/ON DUPLICATE KEY for mysql. This means that even though the error condition happens, the sequence is incremented but that new value is not used because the update happens (not the insert). In a system with large volumes, this causes large gaps in the id column as well as potentially maxing out the sequence (it happened to us).

The fix is to take out the INSERT/ON CONFLICT and replace it by an UPDATE. Only if the update fails does it do an insert.

Note that we have tested this using our Postgres database but we don't have access to a MySQL database to test. It should work but we have no way to validate. The same issues could happen in MySQL so we made changes to both modules.


apanicker-nflx:
@mactaggart @mashurex Can you please with reviewing this PR? Thanks

PR #1796, opened 2020-07-20
Configure listener queues based on the URI settings or config properties

### LINKED ISSUE #1771: Configurable queue settings for bound_to exchange queues
The settings of bound_to_exchange queues are not configurable. It is currently hard coded as explicit, non durable and auto deleted. Can this be made configurable by getting the values from the settings?

If the queue is exclusive, queue mirroring is not happening in rabbitmq cluster environment. So during failover of rabbitmq nodes, the queues are getting deleted and never created again. 

The fix can be to get the settings of bound_exchange queues also from config properties so that these can be set based on the environment.

msdhivya:
This PR contains changes for configuring queues based on the settings provided in the URI. If no settings are provided, then default settings will be used to create the queue.

Closes #1771

### DIFF
--- a/contribs/src/main/java/com/netflix/conductor/contribs/queue/amqp/AMQPObservableQueue.java
+++ b/contribs/src/main/java/com/netflix/conductor/contribs/queue/amqp/AMQPObservableQueue.java
@@ -536,8 +536,13 @@ public class AMQPObservableQueue implements ObservableQueue {
 			if (useExchange) {
 				// Consume messages from an exchange
 				getOrCreateExchange();
-				// Create queue if not present based on the settings provided in the queue URI or configuration properties
-				// Default settings if not provided in the queue URI or properties: isDurable: true, autoDelete: false, isExclusive: false
+				/**
+				 * Create queue if not present based on the settings provided in the queue URI or configuration properties.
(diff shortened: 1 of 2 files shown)


apanicker-nflx:
@ritu-p Could you please help with reviewing this PR? Thanks

contribs/src/main/java/com/netflix/conductor/contribs/queue/amqp/AMQPObservableQueue.java
```diff
@@ -536,7 +536,8 @@ protected void receiveMessages() {
 			if (useExchange) {
 				// Consume messages from an exchange
 				getOrCreateExchange();
-				// Create queue if not present based on the settings
+				// Create queue if not present based on the settings provided in the queue URI or configuration properties
+				// Default settings if not provided in the queue URI or properties: isDurable: true, autoDelete: false, isExclusive: false
```
ritu-p, line 540: Can you put the example url here as to how it is set,
Also mention that this config gets applied to both exchange and queue and we will need a separate key if they need to be separated out

msdhivya: Thank you @ritu-p . I have updated the comment.
Can you please check?

msdhivya:
Thank you @ritu-p 
@apanicker-nflx one of the coverage check is failing in the build. But it is not related to my changes. I have added Unit tests for the changes I have done. Please advise on what should be done.

Issue #1816, opened 2020-08-04
getNextTask will get wrong task if DO_WHILE task is embed in FORK or DECISION


alex-fu:
I create a workflowDef like below. A DO_WHILE task is inner task of a FORK task. When test1 task finished, getNextTask will return test2, and test2 is also considered as inner task in DO_WHILE.

I found a PR #1479 for similar issue. But that's not apply to my situation. In that PR, it only checks the DO_WHILE task in the top level of tasks, if the DO_WHILE is a inner task, that will has no effect.

```
{
  "name": "dowhile_t2",
  "description": "dowhile test 2",
  "version": 1,
  "tasks": [
    {
      "name": "FORK",
      "taskReferenceName": "fork",
      "type": "FORK_JOIN",
      "forkTasks": [
        [
          {
            "name": "WAIT",
            "taskReferenceName": "wait",
            "type": "WAIT"
          }
        ],
        [
          {
            "name": "Loop Task",
            "taskReferenceName": "loopTask",
            "type": "DO_WHILE",
            "inputParameters": {
            },
            "loopCondition": "$.loopTask.iteration < 10",
            "loopOver": [
              {
                "name": "LAMBDA",
                "taskReferenceName": "test1",
                "type": "LAMBDA",
                "inputParameters": {
                  "scriptExpression": "return 1;"
                }
              }
            ]
          },
          {
            "name": "LAMBDA",
            "taskReferenceName": "test2",
            "type": "LAMBDA",
            "inputParameters": {
              "scriptExpression": "return 2;"
            }
          }
        ]
      ]
    },
    {
      "name": "JOIN",
      "taskReferenceName": "join",
      "type": "JOIN",
      "joinOn": [
        "wait",
        "loopTask"
      ]
    }
  ],
  "schemaVersion": 2,
  "ownerEmail": "xxx@gmail.com"
}
```

manan164:
Sure @apanicker-nflx

PR #1820, opened 2020-08-05
1816: getNextTask will get wrong task if DO_WHILE task is embed in FORK or DECISION

alex-fu:
Patch for #1816.

### What I changed
If the last inner task of DO_WHILE was completed and call getNextTask() in decide(), I return the DO_WHILE task (In current code, it will return the next task of DO_WHILE). 
And then, in DeciderService.getNextTask(), I check if the DO_WHILE task already exists in the workflow's tasks. If it exists, just ignore it.


alex-fu:
Choosed wrong target branch

Issue #1782, opened 2020-07-14
Failure Reason field is not getting cleared while hitting RERUN.


sanyamsidana:
Let's start a **workflow**.
<img width="495" alt="Screenshot 2020-07-14 at 1 45 46 PM" src="https://user-images.githubusercontent.com/33194901/87401858-59e4dc00-c5d8-11ea-9841-c4475c078d6f.png">

An **error** occurred at **fetch_data**.
<img width="449" alt="Screenshot 2020-07-14 at 1 48 00 PM" src="https://user-images.githubusercontent.com/33194901/87402069-a5978580-c5d8-11ea-9f2e-701cb8d16c8d.png">

**Failure Reason** field gets populated with the reason of failure.
![failure](https://user-images.githubusercontent.com/33194901/87402765-a5e45080-c5d9-11ea-8f00-66651c6e2721.png)

After calling [RERUN](https://netflix.github.io/conductor/apispec/#rerun) API,
the workflow starts RUNNING as expected, but the Failure Reason field is not getting cleared.
<img width="1789" alt="Again Workflow in running state" src="https://user-images.githubusercontent.com/33194901/87403091-0f645f00-c5da-11ea-9d68-930778a2546b.png">

apanicker-nflx:
@sanyamsidana Thanks for reporting this issue. I have applied the `bug` label to it. We will debug this issue and fix this in a future release.

beatfreaker:
I think in [WorkflowExecutor.java#L1453](https://github.com/Netflix/conductor/blob/be9e74ebd8ee784abb2b5d0eeb9d4c3786fe108f/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java#L1453)

```java
// Along with changing status to RUNNING
workflow.setStatus(WorkflowStatus.RUNNING);

// We have to update below two properties as well
workflow.setReasonForIncompletion(null);
workflow.setFailedReferenceTaskNames(new HashSet<>());
```

@apanicker-nflx  Can I create a PR with this solution ?

apanicker-nflx:
@beatfreaker Please feel free to contribute a PR with the fix. Please also add tests as part of your PR.

PR #1822, opened 2020-08-06
Fix for #1782

beatfreaker:
- Reset failure reason from previous build to default
  when workflow is rerun

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
@@ -1548,6 +1548,9 @@ public class WorkflowExecutor {
         if (rerunFromTask != null) {
             // set workflow as RUNNING
             workflow.setStatus(WorkflowStatus.RUNNING);
+            // Reset failure reason from previous run to default
+            workflow.setReasonForIncompletion(null);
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -1510,6 +1511,9 @@ private boolean rerunWF(String workflowId, String taskId, Map<String, Object> ta
             workflow.getTasks().forEach(task -> executionDAOFacade.removeTask(task.getTaskId()));
             // Set workflow as RUNNING
             workflow.setStatus(WorkflowStatus.RUNNING);
+            // Reset failure reason from previous run to default
+            workflow.setReasonForIncompletion(null);
+            workflow.setFailedReferenceTaskNames(new HashSet<>());
```
apanicker-nflx, line 1516: I think you would need to make similar changes in line 1585 below.

beatfreaker: Did the changes in this [commit](https://github.com/Netflix/conductor/pull/1822/commits/9b9b11d2b3d8e845be4891bf28ce256ea7f800b0)
is it correct?

PR #1823, opened 2020-08-06
1816: getNextTask will get wrong task if DO_WHILE task is embed in FORK or DECISION

alex-fu:
Patch for #1816 .

### What I changed
If the last inner task of DO_WHILE was completed and call getNextTask() in decide(), I return the DO_WHILE task (In current code, it will return the next task of DO_WHILE).
And then, in DeciderService.getNextTask(), I check if the DO_WHILE task already exists in the workflow's tasks. If it exists, just ignore it.

### DIFF
--- a/common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowTask.java
+++ b/common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowTask.java
@@ -586,19 +586,19 @@ public class WorkflowTask {
 							return nextTask;
 						}
 						if (task.has(taskReferenceName)) {
-							if (TaskType.DO_WHILE.name().equals(task.getType())) {
-								// come here means task is DO_WHILE task and `taskReferenceName` is the last task in
-								// this DO_WHILE task, because DO_WHILE task need to be executed to decide whether to
-								// schedule next iteration, so we just return the DO_WHILE task, and then ignore
-								// generating this task again in deciderService.getNextTask()
-								return task;
-							}
 							break;
 						}
 					}
 					if (iterator.hasNext()) {
 						return iterator.next();
 					}
+				}
+				if (taskType == TaskType.DO_WHILE && this.has(taskReferenceName)) {
+					// come here means this is DO_WHILE task and `taskReferenceName` is the last task in
+					// this DO_WHILE task, because DO_WHILE task need to be executed to decide whether to
+					// schedule next iteration, so we just return the DO_WHILE task, and then ignore
+					// generating this task again in deciderService.getNextTask()
+					return this;
 				}
 				break;
 			case FORK_JOIN:
@@ -616,10 +616,6 @@ public class WorkflowTask {
 							return nextTask;
 						}
 						if (task.has(taskReferenceName)) {
-							if (TaskType.DO_WHILE.name().equals(task.getType())) {
-								// same reason as above
-								return task;
-							}
 							break;
 						}
 					}
--- a/common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowTask.java
+++ b/common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowTask.java
@@ -586,6 +586,13 @@ public class WorkflowTask {
 							return nextTask;
 						}
 						if (task.has(taskReferenceName)) {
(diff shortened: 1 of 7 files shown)


apanicker-nflx:
@alex-fu Thanks for the contribution. Can you please also add an integration test for this [here](https://github.com/Netflix/conductor/blob/master/test-harness/src/test/groovy/com/netflix/counductor/integration/test/DoWhileSpec.groovy). Could you also add an integration test which would handle a failure and automated retry in one of the forked branches within the do_while?

@manan164 We would appreciate it if you could help with reviewing this PR.

common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowTask.java
```diff
@@ -569,6 +569,13 @@ public WorkflowTask next(String taskReferenceName, WorkflowTask parent) {
 							return nextTask;
 						}
 						if (task.has(taskReferenceName)) {
+							if (TaskType.DO_WHILE.name().equals(task.getType())) {
```
manan164, line 572: @alex-fu Since we are diverging the DECISION block by so many lines of code. We can define a separate switch case for DO_WHILE task. And move common code to one function.

core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
```diff
@@ -362,6 +363,13 @@ private boolean checkForWorkflowCompletion(final Workflow workflow) throws Termi
         while (isTaskSkipped(taskToSchedule, workflow)) {
             taskToSchedule = workflowDef.getNextTask(taskToSchedule.getTaskReferenceName());
         }
+        if (taskToSchedule != null && TaskType.DO_WHILE.name().equals(taskToSchedule.getType())) {
+            // check if already has this DO_WHILE task, ignore it if it already exists
+            String nextTaskReferenceName = taskToSchedule.getTaskReferenceName();
+            if (workflow.getTasks().stream().anyMatch(runningTask -> runningTask.getReferenceTaskName().equals(nextTaskReferenceName))) {
+                return Collections.emptyList();
```
manan164, line 370: Why we are returning an empty list. We are making get next task to return the parent DO_WHILE task let it propagates it further to decider service.

ritu-p:
Hi ,
 I have embedded the DO_WHILE in a decision task and I am facing a similar issue where the task outside the loop task is getting executed.Any timeline by which this PR will be merged?

apanicker-nflx:
@alex-fu Are you available to continue working on this PR? Thanks

alex-fu:
@apanicker-nflx  Sorry for the late reply since I was stuck in some things last month. I'll add the integration test ASAP.

common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowTask.java
alex-fu: OK, I'll move this common code to a function. But this function should be called from DECISION block and FORK block, not DO_WHILE block(since we don't allow nested DO_WHILE task).

core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
alex-fu: Because in this case the parent DO_WHILE task is already exist, so we don't need to generate again. By the way, even though we return the parent DO_WHILE task, deciderService will also exclude it in below line. tasksToBeScheduled already contains this task, and putIfAbsent will not add this task again.
```
nextTasks.forEach(nextTask -> tasksToBeScheduled.putIfAbsent(nextTask.getReferenceTaskName(), nextTask));
```

apanicker-nflx:
Please rebase against `dev`, seems like the build failure is due to a conflict. 
https://travis-ci.com/github/Netflix/conductor/builds/186799651 
image

apanicker-nflx:
Also, while unit tests look good, please also add an integration test for this [here](https://github.com/Netflix/conductor/blob/master/test-harness/src/test/groovy/com/netflix/counductor/integration/test/DoWhileSpec.groovy)

alex-fu:
@apanicker-nflx  I've rebased against dev and submitted some code according to @manan164's review and some integration test, please review again.

PR #1824, opened 2020-08-06
DO_WHILE: expand DO_WHILE task in conductor UI

alex-fu:
### Motivation
DO_WHILE task in conductor UI was shown as a simple task. Its inner tasks was not shown in graph.

### What I changed
Add support for DO_WHILE task in wfegraph.js. 

### Show result
image


manan164:
Thanks, @alex-fu for taking this up. This was there since a long time in my backlog but since I have little experience with JavaScript it was always getting delayed.

PR #1825, opened 2020-08-07
Update beginner.md

zhouwentong1993:
Create a task needs an owerEmail, Otherwise HTTP Response 400


zhouwentong1993:
Creating Workflow Definition's HTTP request also need a **ownerEmail** node

PR #1826, opened 2020-08-07
Postgres Workflow Metadata Update/Remove

srjinatl:
Problems discovered in our testing of v28.0:

- adding workflow for a given name but not in increasing version order does not work - workflow out of order will not get added to the database but will incorrectly cause the latest_version column in the db to be set to its version.
- removing workflow does not update latest_version column

The first problem can cause a workflow not found when trying to retrieve the latest version of the workflow via the API - i.e. version number is not supplied.

### DIFF
--- a/postgres-persistence/src/main/java/com/netflix/conductor/dao/postgres/PostgresMetadataDAO.java
+++ b/postgres-persistence/src/main/java/com/netflix/conductor/dao/postgres/PostgresMetadataDAO.java
@@ -141,11 +141,18 @@ public class PostgresMetadataDAO extends PostgresBaseDAO implements MetadataDAO,
     public void removeWorkflowDef(String name, Integer version) {
         final String DELETE_WORKFLOW_QUERY = "DELETE from meta_workflow_def WHERE name = ? AND version = ?";
 
-        executeWithTransaction(DELETE_WORKFLOW_QUERY, q -> {
-            if (!q.addParameter(name).addParameter(version).executeDelete()) {
-                throw new ApplicationException(ApplicationException.Code.NOT_FOUND,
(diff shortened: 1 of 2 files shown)


apanicker-nflx:
@mactaggart @rickfish Can you please help with reviewing this PR? Also, do you think a similar change would be needed in the mysql-persistence module?

rickfish:
@apanicker-nflx, @mactaggart @srjinatl  this looks good to me. I do think that a change to the MySql DAO is also needed. Whenever I make changes to Postgres classes I also make them in MySQL. The logic between the two is very similar. In fact, before the new Postgres-persistence module was added, we had changed mysql-persistence to sql-persistence and when there was a difference between them we extended the appropriate class with MySQL, Postgres variants and added the appropriate change to both. There were just a handful of differences, mainly in the ON CONFLICT type of stuff and some date-related things.

mactaggart:
@apanicker-nflx I don't see anything Postgres specific here, so we'll need to update the MySQL code too. As @rickfish mentioned, the Postgres module is based on the existing MySQL implementation, with minor pg-specific differences between them. I had originally submitted a PR that extracted common code between MySQL/Postgres into a sql-persistence module. Perhaps this approach could be revisited since it would cut down on duplicate logic?

apanicker-nflx:
Thanks @rickfish and @mactaggart for your help. 💯 

@srjinatl It would be really helpful if you could also make the same changes to the mysql-persistence module.

srjinatl:
Ok I will take a look at the mysql-persistence module - if I see the same issue there I will submit a new PR for those changes if that is ok.

apanicker-nflx:
@srjinatl Sounds good, thanks for your contribution. Please make sure to follow through on the mysql module.

PR #1828, opened 2020-08-10
Add the possibility to override retryCount on WorkflowTask

andrea11:


elisherer:
Don't embedded task definition already accomplishes that?

PR #1830, opened 2020-08-17
Feature: Task resiliency from QueueDAO failures

kishorebanala:
### Problem

In the current implementation:

- Failure to schedule a task will terminate the workflow.
-- A Task scheduling can fail when failing to persist the task, or pushing the taskId to Task Queues.
- A task could remain stuck if the persistence layer and queuing layer doesn’t agree:
-- For eg: if the relevant message is lost in the Task queue.

### Solution

The process for scheduling a task would remain the same as now:

1. Task status is set as SCHEDULED, and persisted.
2. Task is pushed to the Task queue.

In this PR, if Step 1 fails, we'll continue to terminate workflows (which is the existing behavior). If Step 2 fails, The next run of sweeper fixes the inconsistency by verifying the state of the message in the Queue. I.e if the Task is in SCHEDULED state, but a relevant message is not found in the Queue (including those messages in invisible state), a new message will be pushed.

### Major additions

- A new WorkflowRepairService has been introduced, that checks non-terminal workflow and task state with message in queue, and performs necessary actions as required. This will be called from Sweeper service.
- `/consistency/verifyAndRepair/{workflowId}` is added to AdminResource, which verifies and pushes if workflow and tasks have relevant messages in the queue.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
@@ -1383,6 +1383,8 @@ public class WorkflowExecutor {
     @VisibleForTesting
     boolean scheduleTask(Workflow workflow, List<Task> tasks) {
         List<Task> createdTasks;
+        List<Task> tasksToBeQueued;
+        boolean startedSystemTasks = false;
 
         try {
             if (tasks == null || tasks.isEmpty()) {
@@ -1408,11 +1410,9 @@ public class WorkflowExecutor {
                     .filter(isSystemTask)
                     .collect(Collectors.toList());
 
-            List<Task> tasksToBeQueued = createdTasks.stream()
+            tasksToBeQueued = createdTasks.stream()
                     .filter(isSystemTask.negate())
                     .collect(Collectors.toList());
-
-            boolean startedSystemTasks = false;
 
             // Traverse through all the system tasks, start the sync tasks, in case of async queue the tasks
             for (Task task : systemTasks) {
@@ -1438,9 +1438,6 @@ public class WorkflowExecutor {
                     tasksToBeQueued.add(task);
                 }
             }
-
-            addTaskToQueue(tasksToBeQueued);
-            return startedSystemTasks;
         } catch (Exception e) {
             List<String> taskIds = tasks.stream()
                     .map(Task::getTaskId)
@@ -1453,6 +1450,19 @@ public class WorkflowExecutor {
             // rollbackTasks(workflow.getWorkflowId(), createdTasks);
             throw new TerminateWorkflowException(errorMsg);
         }
+
+        // On addTaskToQueue failures, ignore the exceptions and let WorkflowRepairService take care of republishing the messages to the queue.
+        try {
+            addTaskToQueue(tasksToBeQueued);
+        } catch (Exception e) {
+            List<String> taskIds = tasksToBeQueued.stream()
+                    .map(Task::getTaskId)
+                    .collect(Collectors.toList());
+            String errorMsg = String.format("Error pushing tasks to the queue: %s, for workflow: %s", taskIds, workflow.getWorkflowId());
+            LOGGER.warn(errorMsg, e);
+            Monitors.error(className, "scheduleTask");
+        }
+        return startedSystemTasks;
     }
 
     /**
--- a/core/src/main/java/com/netflix/conductor/core/execution/WorkflowRepairService.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/WorkflowRepairService.java
@@ -1,0 +1,126 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package com.netflix.conductor.core.execution;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.netflix.conductor.common.metadata.tasks.Task;
+import com.netflix.conductor.common.run.Workflow;
+import com.netflix.conductor.core.config.Configuration;
+import com.netflix.conductor.core.execution.tasks.WorkflowSystemTask;
+import com.netflix.conductor.core.utils.QueueUtils;
+import com.netflix.conductor.dao.ExecutionDAO;
+import com.netflix.conductor.dao.QueueDAO;
+import com.netflix.conductor.metrics.Monitors;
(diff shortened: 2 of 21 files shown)


core/src/main/java/com/netflix/conductor/core/execution/WorkflowRepairService.java
```diff
@@ -0,0 +1,112 @@
...
+
+    /**
+     * Verify if ExecutionDAO and QueueDAO agree for the provided task.
+     * @param task
+     * @return
+     */
+    @VisibleForTesting
+    protected boolean verifyAndRepairTask(Task task) {
+        WorkflowSystemTask workflowSystemTask = WorkflowSystemTask.get(task.getTaskType());
+        if (task.getStatus().equals(Task.Status.SCHEDULED)) {
```
aravindanr, line 99:
```suggestion
        if (task.getStatus() == Task.Status.SCHEDULED)) {
```
Enums can be compared using the operator and it also helps avoiding a `NullPointerException`

common/src/main/java/com/netflix/conductor/common/constraints/FaultInjectionInterceptor.java
```diff
@@ -0,0 +1,14 @@
+package com.netflix.conductor.common.constraints;
```
apanicker-nflx, line 1: Please add license header.

common/src/main/java/com/netflix/conductor/common/constraints/FaultInjectionInterceptor.java
```diff
@@ -0,0 +1,14 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Interceptor intended for failure injection during unit / integration testing.
+ */
+@Retention(RetentionPolicy.RUNTIME) @Target(ElementType.METHOD)
+public @interface FaultInjectionInterceptor {
```
apanicker-nflx, line 13: This should probably be added to the core module.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -785,6 +782,11 @@ public void terminateWorkflow(Workflow workflow, String reason, String failureWo
             if (workflow.getWorkflowDefinition().isWorkflowStatusListenerEnabled()) {
                 workflowStatusListener.onWorkflowTerminated(workflow);
             }
+
+            //remove from the sweep queue
```
apanicker-nflx, line 786: tasks should be removed from queue before status is updated, else tasks to be CANCELED will be polled and processed by the workers.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -1447,6 +1446,19 @@ boolean scheduleTask(Workflow workflow, List<Task> tasks) {
             // rollbackTasks(workflow.getWorkflowId(), createdTasks);
             throw new TerminateWorkflowException(errorMsg);
         }
+
+        // On addTaskToQueue failures, ignore the exceptions and let WorkflowRepairService take care of republishing the messages to the queue.
```
apanicker-nflx, line 1450: Does this also account for partial failures?

core/src/main/java/com/netflix/conductor/core/execution/WorkflowRepairService.java
```diff
@@ -0,0 +1,112 @@
+package com.netflix.conductor.core.execution;
```
apanicker-nflx, line 1: Please add license header

core/src/main/java/com/netflix/conductor/core/execution/WorkflowRepairService.java
```diff
@@ -0,0 +1,112 @@
...
+import javax.inject.Inject;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Predicate;
+
+/**
+ * A helper service that tries to keep ExecutionDAO and QueueDAO in sync, based on the
+ * task or workflow state.
+ *
+ * This service expects that the underlying Queueing layer implements QueueDAO.containsMessage method. This can be controlled
+ * with Configuration.isWorkflowRepairServiceEnabled() property.
```
apanicker-nflx, line 23: It would be better to link out to the method from the javadoc.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowRepairService.java
```diff
@@ -0,0 +1,112 @@
...
+     */
+    @VisibleForTesting
+    protected boolean verifyAndRepairTask(Task task) {
+        WorkflowSystemTask workflowSystemTask = WorkflowSystemTask.get(task.getTaskType());
+        if (task.getStatus().equals(Task.Status.SCHEDULED)) {
+            if (isSystemTask.test(task) && !workflowSystemTask.isAsync()) {
+                return false;
+            }
+            // Ensure QueueDAO contains this taskId
+            if (!queueDAO.containsMessage(task.getTaskDefName(), task.getTaskId())) {
```
apanicker-nflx, line 104: For system tasks, the tasks are enqueued in the task type queues and not by the task def name, does this line need to change?

redis-persistence/src/main/java/com/netflix/conductor/dao/dynomite/RedisExecutionDAO.java
```diff
@@ -115,6 +116,7 @@ public RedisExecutionDAO(DynoProxy dynoClient, ObjectMapper objectMapper, Config
 		return tasks;
 	}
 
+	@FaultInjectionInterceptor
```
apanicker-nflx, line 119: Since this annotation and the interceptor are used only for testing, can this abstracted out from the execution path. Maybe through some combination of Spy and Mock?

test-harness/src/test/groovy/com/netflix/counductor/integration/test/TaskResiliencySpec.groovy
```diff
@@ -0,0 +1,131 @@
+package com.netflix.counductor.integration.test
```
apanicker-nflx, line 1: Please add license header.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
kishorebanala: Yes. Partial failures when adding to queue are fine. For eg., if only 2 of 5 messages are added to the queue and the rest failed, the sweeper service would still check if the queue contains each message as long as the task is in scheduled state, and this can be recovered.

test-harness/src/test/groovy/com/netflix/counductor/integration/test/TaskResiliencySpec.groovy
```diff
@@ -0,0 +1,125 @@
...
+        workflowTestUtil.registerWorkflows(
+                'simple_workflow_1_integration_test.json'
+        )
+    }
+
+    def cleanup() {
+        workflowTestUtil.clearWorkflows()
+    }
+
+    def "Verify that a workflow recovers and completes on schedule task failure from queue push failure"() {
```
apanicker-nflx, line 63: Nice

apanicker-nflx (approved):

Issue #1834, opened 2020-08-19
Add global workflow variables support


heinrichcohn:
@apanicker-nflx @elisherer 
I would like to add a new type of task that allows create/update variables at the workflow level. This feature is based on:
[Issue:1207](https://github.com/Netflix/conductor/issues/1207)
[PR:1323](https://github.com/Netflix/conductor/pull/1323) which was not merged eventually.

The idea is to have a dedicated system-task which can be used to set variables as part of a workflow. These variables will be stored in a Map field of the workflow and will be accessible from other tasks

PR #1835, opened 2020-08-19
Feature/add set workflow variables task

heinrichcohn:
@apanicker-nflx @elisherer 
Pull request related to [Issue 1834](https://github.com/Netflix/conductor/issues/1834)


elisherer:
Running the following workflow:
```json
{
  "schemaVersion": 2,
  "name": "test_set_variable_wf",
  "version": 1,
  "ownerEmail": "test@example.com",
  "variables": {
    "var_from_def": "def_value"
  },
  "inputParameters": [
    "var"
  ],
  "outputParameters": {
    "variables": "${workflow.variables}"
  },
  "tasks": [
    {
      "type": "SET_VARIABLE",
      "name": "set_variable",
      "taskReferenceName": "set_variable_1",
      "inputParameters": {
        "var_from_input": "${workflow.input.var}",
        "num": 2,
        "bool": true,
        "obj": {
          "nested": "halo"
        },
        "to_be_nulled": "some value"
      }
    },
    {
      "type": "SET_VARIABLE",
      "name": "set_variable_2",
      "taskReferenceName": "set_variable_2",
      "inputParameters": {
        "to_be_nulled": null
      }
    }
  ]
}
```

Will yield the output:
```json
{
  "variables": {
    "var_from_def": "def_value",
    "bool": true,
    "var_from_input": null,
    "to_be_nulled": null,
    "obj": {
      "nested": "halo"
    },
    "num": 2
  }
}
```

apanicker-nflx:
@heinrichcohn @elisherer Thank you for taking over this feature. 

I would love to know your thoughts about adding constraints to prevent misuse of this task.

elisherer:
> @heinrichcohn @elisherer Thank you for taking over this feature.
> 
> I would love to know your thoughts about adding constraints to prevent misuse of this task.

So we actually talked about it, seeing your comment on the last PR, and I was thinking that every user of this product will surely begin by going to the documentation. So that's why I added a wraning to the docs:
```md
!!!warning
	Use variable values with caution for payload size. Do not use large payload since these values are not externalized to the [external payload storage](../externalpayloadstorage).
```
I think no matter what the prevention measure will be, it might cause more bad behavior than good. Since if we put limits on the variables size (we need to check it on every step, because it gets changed), the limits can go over size of a single value. limit on the amount of variables. And considering edge cases; what happens when a limit is changed (for the worst) and there are workflows that go over the limit?
If at all, we could plan to move the variables values to the external storage as well but that defeats the purpose. It suppose to be simple small values to help with the orchestration. for us it is useful in failure workflow to understand what have been done already to setup the cancellation.

kishorebanala:
Hey @elisherer & @heinrichcohn, thanks for picking up this long awaited feature.

While documentation definitely helps, we could still run into a situation where setVariables is called enough number of times, knowingly or accidentally to become a large payload blob. This could in turn cause unforeseen issues in Persistence layer, for eg., wide rows in Cassandra, which we'll have limited control over.

How about checking the variables hashmap size in SetVariable's `execute` method, everytime this task is executed? I.e We shall limit max size to a configurable value defaulting to 5MB, and reject any updates to this hashmap over this limit by failing the task.

elisherer:
OK, should we add that check to the LAMBDA task as well? (Since it will be possible to set variables from there too)

elisherer:
@kishorebanala , I made a hard barrier (using the same terminology for the external storage barriers).
new system property `conductor.max.workflow.variables.payload.threshold.kb` with default value of 256 (as suggested)
the SET_VARIABLE will fail with FAILED_WITH_TERMINAL_ERROR if passing the limit (tested in an additional integration test)

I'm leaving LAMBDA alone at this time because it's a wildcard and a lot of "unwanted" changes can occur there.

(I found a typo in the error message of passing the limit in `ExternalPayloadStorageUtils.verifyAndUpload`, while copying the logic from there, stating the payload size in KB while in fact it's in Bytes so I changed it in this PR)

PR #1845, opened 2020-08-26
Acknowledge message received after polling from the RedisDynoQueue to remove the message payload

mdepak:
**DynoQueue context:**
When a message is pushed to queue, it is first added to the queue, and message payload is maintained in a separate hash. When the message is popped from the queue, it is removed from the queue and the message payload still remains in the hash. Once the client acknowledges using (ACK), the message payload is removed from the hash and the message-id is removed from the UNACK set.

**Problem:**
When the ack is not performed by the client polling the message, it leads to a pile-up of messages payloads in the hash. This pile of messages consumes a lot of memory of Redis/Dynomite. ACK is not performed for the polled messages in the following places,

1. When system tasks are polled from the SystemTaskExecutor, the task messages are not acknowledged.
2. In the DynoObservableQueue, when the message is polled, in the ack method, queue "remove" is performed instead of acknowledging the polled message. This removes the UNACK key and tries to remove the message from the queue and fails before deleting the message payload in the hash.

**Solution:**
1. Once the message (system task) is polled and executed in the SystemTaskExecutor, acknowledge that the message is received.
2. In the DynoObservableQueue, acknowledge the message instead of removing the message again after polling the message.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/events/queue/dyno/DynoObservableQueue.java
+++ b/core/src/main/java/com/netflix/conductor/core/events/queue/dyno/DynoObservableQueue.java
@@ -77,7 +77,7 @@ public class DynoObservableQueue implements ObservableQueue {
     @Override
     public List<String> ack(List<Message> messages) {
         for (Message msg : messages) {
-            queueDAO.remove(queueName, msg.getId());
+            queueDAO.ack(queueName, msg.getId());
         }
         return messages.stream().map(Message::getId).collect(Collectors.toList());
     }
--- a/core/src/main/java/com/netflix/conductor/core/execution/tasks/SystemTaskExecutor.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/tasks/SystemTaskExecutor.java
@@ -26,6 +26,8 @@ import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
+
+import com.netflix.conductor.service.ExecutionService;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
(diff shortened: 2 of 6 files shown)


manan164:
Hi @mdepak , 
Thanks for pointing out. 
Are you using clustered redis?
But for the happy case also we see that the message is not removed from .MESSAGE.HTTP even after successful execution. Messages get piled up in high throughput scenario. This is the same behaviour for all other queues also.

mdepak:
@manan164 We use the Redis cluster for our deployment. `HTTP` is an async system task and they are polled also polled and executed using SystemTaskExecutor and hence it is applicable for HTTP system task as well leading to a pile-up of messages in `*.MESSAGE.HTTP` hash. 

When the task is polled by the task workers' clients, it is acknowledged by the client and thus corresponding message payloads are removed. In the deprecated `WorkflowTaskCoordinator` the tasks are acknowledged once the task is received. However, in the `TaskRunnerConfigurer` once the task is polled it is not acknowledged and that could lead to a similar problem related to non-system tasks with RedisDynoQueues. @apanicker-nflx  Can you please shed some light on this problem with respect to the new `TaskRunnerConfigurer` client?

apanicker-nflx:
@mdepak In the TaskRunnerConfigurer, the ack method was intentionally removed and instead the poll for single task API is used which implicitly means that once a task is polled, it will be executed by the client instead of being stored in client side queues. As part of this change, the poll API was retrofitted to actively ack the tasks on the server side ([see](https://github.com/Netflix/conductor/blob/f5cbfa24769802712807988222243f2a1ac6afb6/core/src/main/java/com/netflix/conductor/service/ExecutionService.java#L189)). Hence, this would not be a problem for non-system worker tasks.

However, as you have correctly pointed out, this could lead to a problem with system tasks and dyno event queues. Thanks for your contribution.

PR #1846, opened 2020-08-27
ability to retry workflows timed out due to workflow level timeout

apanicker-nflx:


common/src/main/java/com/netflix/conductor/common/run/Workflow.java
```diff
@@ -492,6 +502,12 @@ public Workflow copy() {
 		copy.setTasks(tasks.stream()
 				.map(Task::deepCopy)
 				.collect(Collectors.toList()));
+		copy.setEndTime(endTime);
```
kishorebanala, line 505: Wondering if adding these fields would cause any side effects, as the copy is being used in decider evaluations at different stages. Do we have enough test coverage around usage of copy() ?

apanicker-nflx: Verified that the integration/unit tests do cover these.

PR #1847, opened 2020-08-28
Add ability to configure the connection request timeout for Elasticsearch rest client

mdepak:
By default, the connection request timeout is interpreted from the system default. Under high throughput scenarios, elasticsearch client timeout when getting the connection from the HTTP rest-client library.


kishorebanala:
@mdepak +1. Can you also add this to es6-persistence module to keep both the modules consistent please?

mdepak:
@kishorebanala Added the config to `es6-persistence` module as well.

apanicker-nflx (approved):

PR #1867, opened 2020-09-14
update jq dependency to latest version

andrea11:


apanicker-nflx:
From the [GitHub page](https://github.com/eiiches/jackson-jq) for this library: 
> develop/1.x: This branch (you are viewing), which is currently under development for the future 1.0 release. The API is not stable yet. You can find preview releases (**not stable, not recommended for production**) at Releases page (tags: 1.0.0-preview.yyyyMMdd)

I think it would be prudent to wait for a stable version to be published before updating the dependency.

PR #1873, opened 2020-09-18
when rerunning workflow, to find the rerunFromTaskId task, check tasks directly under the workflow before checking sub_workflow tasks

rickfish:
In workflows that have a lot of sub_workflow tasks, the rerun API is very slow because of the way it tries to find the task associated with rerunFromTaskId. If the task is directly under the workflow but there are sub_workflow tasks in the tasks list before the specified task, it will walk the chain of subworkflows looking for the task. This PR makes it first look at the tasks directly under the workflow before walking the sub_workflow chains.


rickfish:
@apanicker-nflx my build failed because LoggingMetricsModuleTest verify() did not pass its test. This has nothing to do with my change as far as I can tell. What should I do?

apanicker-nflx (approved):

PR #1885, opened 2020-09-23
simplify how we get PollData for all task types at once

rickfish:
The /tasks/queue/polldata/all endpoint has a very convoluted way of getting PollData for all task types at once. In our system with tons of polling, it causes that endpoint to be very slow. This PR greatly simplifies how we do the queries for MySQL and Postgres. Cassandra still does it the old way because I have no way of testing it with that database.


rickfish:
@apanicker-nflx my build failed because LoggingMetricsModuleTest verify() did not pass its test. This has nothing to do with my change as far as I can tell. What should I do?

apanicker-nflx (approved):

PR #1893, opened 2020-09-30
Concurrent task execution limit

TwoUnderscorez:
While using the Netflix Conductor we found out that we might face some edge cases. Some of our Workflows contain tasks that connect and perform work on some other servers (multiple `WorkflowTasks` connect to the same server), and these servers are legacy and easily overwhelmed. The only way we can fix the problem is place a concurrent execution limit on a specific `WorkflowTask`, that's why we are experiencing a lot of performance issues. I think this could be a really useful feature for some rare edge cases for other people as well and would be happy if you could consider merging this.

- Add semaphore DAO, implemented for redis/dynomite according to the [official docs](https://redislabs.com/ebook/part-2-core-concepts/chapter-6-application-components-in-redis/6-3-counting-semaphores/6-3-1-building-a-basic-counting-semaphore/)
    - Added `zrank` to `DynoProxy`
- Added settings:
   - `workflow.execution.executionLimitingSemaphoreName` - prefix for semaphores used to limit the concurrent executions of tasks. Defaults to `conductor.task_semaphore_`.
    - `workflow.execution.enableLocalTaskConcurrentExecLimit` - System-wide switch for enabling or disabling the check for concurrent execution limits in a workflow. Defaults to `false`.
    - `workflow.execution.enableGlobalTaskConcurrentExecLimit` - System-wide switch for enabling or disabling the check for concurrent execution globally. Defaults to `false`.
- Added fields to `com.netflix.conductor.common.metadata.workflow.WorkflowTask`
    - `localConcurrentExecutionLimit` - Specifies the maximum desired concurrent executions there should be of the specified `WorkflowTask` in the current `WorkflowDef`. Let TD be a `TaskDef`. For example, if workflow A has a task with `taskReferenceName` = t1, `name` = TD and `localConcurrentExecutionLimit` = 2. Then you post trice a start workflow request to `/api/workflow`, Conductor will make sure that t1 will not be scheduled to more than 2 workers at a time. Regardless of how many workers of type TD there are.
    - `globalConcurrentExecutionLimit` - Specifies the maximum desired concurrent executions there should be of the specified `WorkflowTask` globally. Let TD be a `TaskDef`, WF1 and WF2 be `WorkflowDef`s with the following task:  `taskReferenceName` = t1, `name` = TD and `globalConcurrentExecutionLimit` = 2. Now if we post trice a start workflow request to `/api/workflow`, once for WF1 and twice for WF2, Conductor will make sure that t1 will not be scheduled to more than 2 workers at a time. Regardless of how many workers of type TD there are. The `WorkflowTask`s are linked by their `taskReferenceName`.
- Added validations:
    - to `WorkflowTask`:
        - Either `localConcurrentExecutionLimit` or `globalConcurrentExecutionLimit` can be set to more than 0, not both.
        - When `localConcurrentExecutionLimit` or `globalConcurrentExecutionLimit` are set, makes sure the the corresponding `TaskDef` has a timeout defined (the timeout is used for the semaphore).
- Undefnied behavior if (not validated): 
    - `globalConcurrentExecutionLimit` is enabled for 2 or more `WorkflowTask`s with a different limit count.
    - `globalConcurrentExecutionLimit` is enabled for 2 or more `WorkflowTask`s with different `TaskDef`s (and therefore possibly different timeouts).

### DIFF
--- a/core/src/main/java/com/netflix/conductor/dao/SemaphoreDAO.java
+++ b/core/src/main/java/com/netflix/conductor/dao/SemaphoreDAO.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.conductor.dao;
+
+import com.netflix.conductor.common.metadata.tasks.Task;
+
+/**
+ * An abstraction to enable different Rate Limiting implementations
+ */
+public interface SemaphoreDAO {
+
+    /**
+     * Checks if the Task is rate limited or not based on the {@link Task#getRateLimitPerFrequency()} and {@link Task#getRateLimitFrequencyInSeconds()}
+     * @param task: which needs to be evaluated whether it is rateLimited or not
+     * @return true: If the {@link Task} is rateLimited
+     * 		false: If the {@link Task} is not rateLimited
+     */
+    boolean tryAcquire(String semaphoreName, String identifier, int limit, double timeoutMillis);
+
+    boolean release(String semaphoreName, String identifier);
+}
--- a/core/src/main/java/com/netflix/conductor/dao/SemaphoreDAO.java
+++ b/core/src/main/java/com/netflix/conductor/dao/SemaphoreDAO.java
@@ -18,17 +18,27 @@ package com.netflix.conductor.dao;
 import com.netflix.conductor.common.metadata.tasks.Task;
 
 /**
- * An abstraction to enable different Rate Limiting implementations
+ * An abstraction for a non-blocking semaphore object
  */
 public interface SemaphoreDAO {
 
     /**
-     * Checks if the Task is rate limited or not based on the {@link Task#getRateLimitPerFrequency()} and {@link Task#getRateLimitFrequencyInSeconds()}
-     * @param task: which needs to be evaluated whether it is rateLimited or not
-     * @return true: If the {@link Task} is rateLimited
-     * 		false: If the {@link Task} is not rateLimited
+     * Try to acquire the semaphore
+     * @param semaphoreName: The name of the semaphore
+     * @param identifier: A unique identifier that identifies the holder
+     * @param limit: This semaphore's maximum number of holders limit. If changes between calls for the same semaphore name, behavior is undefined.
+     * @param timeoutMillis: A timeout for all the holders of the semaphore If changes between calls for the same semaphore name, behavior is undefined.
+     * @return true: If the semaphore was successfully acquired
+     * 		false: If the semaphore was successfully not acquired (e.g. Reached limit)
      */
     boolean tryAcquire(String semaphoreName, String identifier, int limit, double timeoutMillis);
 
+    /**
+     * Release the semaphore
+     * @param semaphoreName: The name of the semaphore
+     * @param identifier: A unique identifier that identifies the holder
+     * @return true: If the semaphore was successfully released
+     * 		false: If the semaphore was successfully not released (e.g. Never acquired in the first place)
+     */
     boolean release(String semaphoreName, String identifier);
 }
--- a/core/src/main/java/com/netflix/conductor/validations/WorkflowTaskTypeConstraint.java
+++ b/core/src/main/java/com/netflix/conductor/validations/WorkflowTaskTypeConstraint.java
@@ -121,7 +121,7 @@ public @interface WorkflowTaskTypeConstraint {
                 && task.getTimeoutSeconds() < 1)
             {
                 valid = false;
-                String message = String.format("workflowTask: %s task definition %s must have timeoutSeconds defined if *ConcurrentExecutionLimit is defined in the workflowTask", workflowTask.getName(), workflowTask.getName());
+                String message = String.format("workflowTask: %s task definition %s must have timeoutSeconds defined if *ConcurrentExecutionLimit is defined in the workflowTask.", workflowTask.getName(), workflowTask.getName());
                 context.buildConstraintViolationWithTemplate(message).addConstraintViolation();
             }
             if (workflowTask.getGlobalConcurrentExecutionLimit() * workflowTask.getLocalConcurrentExecutionLimit() != 0)
--- a/core/src/main/java/com/netflix/conductor/validations/WorkflowTaskTypeConstraint.java
+++ b/core/src/main/java/com/netflix/conductor/validations/WorkflowTaskTypeConstraint.java
@@ -98,8 +98,38 @@ public @interface WorkflowTaskTypeConstraint {
                 case TaskType.TASK_TYPE_JSON_JQ_TRANSFORM:
                     valid = isJSONJQTransformTaskValid(workflowTask, context);
                     break;
+                case TaskType.TASK_TYPE_SIMPLE:
+                    valid = isSimpleTaskValid(workflowTask, context);
+                    break;
             }
 
+            return valid;
+        }
+
+        private boolean isSimpleTaskValid(WorkflowTask workflowTask, ConstraintValidatorContext context) {
+            TaskDef task = workflowTask.getTaskDefinition();
+            if(task == null) {
+                task = ValidationContext.getMetadataDAO().getTaskDef(workflowTask.getName());
+            }
+            if (task == null) {   
+                String message = String.format("workflowTask: %s must have a task definition defined ", workflowTask.getName());
+                context.buildConstraintViolationWithTemplate(message).addConstraintViolation();
+                return false;
+            }
+            boolean valid = true;
+            if (workflowTask.getGlobalConcurrentExecutionLimit() + workflowTask.getLocalConcurrentExecutionLimit() != 0
+                && task.getTimeoutSeconds() < 1)
+            {
+                valid = false;
+                String message = String.format("workflowTask: %s task definition %s must have timeoutSeconds defined if *ConcurrentExecutionLimit is defined in the workflowTask", workflowTask.getName(), workflowTask.getName());
+                context.buildConstraintViolationWithTemplate(message).addConstraintViolation();
+            }
+            if (workflowTask.getGlobalConcurrentExecutionLimit() * workflowTask.getLocalConcurrentExecutionLimit() != 0)
+            {
+                valid = false;
+                String message = String.format("workflowTask: %s you can't set both globalConcurrentExecutionLimit and localConcurrentExecutionLimit, choose one.", workflowTask.getName());
+                context.buildConstraintViolationWithTemplate(message).addConstraintViolation();
+            }
             return valid;
         }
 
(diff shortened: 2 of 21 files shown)


TwoUnderscorez:
@apanicker-nflx Could you please review my additions?
I would also like to add another implementation for the `SemaphoreDAO` that implements "fair semaphore" ([read more](https://redislabs.com/ebook/part-2-core-concepts/chapter-6-application-components-in-redis/6-3-counting-semaphores/6-3-2-fair-semaphores/)) for when setting up Conductor in HA and the clocks are not synchronized and unsure of how to get an instance of the configuration in the `Configure` method of `RedisWorkflowModule` to make this configurable by the user .

TwoUnderscorez:
@mashurex @jvemugunta @demichej Could somebody please review this?
Thanks!

kishorebanala:
Hey @TwoUnderscorez, Can you provide more details about how this ties together with [Task concurrent execution limits ](https://netflix.github.io/conductor/configuration/taskdef/#task-concurrent-execution-limits) please?

TwoUnderscorez:
@kishorebanala I'm not entirely sure what you mean by how it ties together, but the order of checks in `exceedsInProgressLiint` are as follows:
1. `TaskDef` concurrent execution limit
2. `WorkflowTask` execution limit (what I have added)
3. `TaskDef`rate limit check
Checks 1 and 2 are completely separate checks.

redis-persistence/src/main/java/com/netflix/conductor/dao/dynomite/RedisExecutionDAO.java
```diff
@@ -202,18 +227,38 @@ public boolean exceedsInProgressLimit(Task task) {
...
 			Monitors.recordTaskConcurrentExecutionLimited(task.getTaskDefName(), limit);
 			return true;
 		}
 
+		if(enableGlobalTaskConcurrentExecLimit || enableLocalTaskConcurrentExecLimit) {
+			logger.debug("ExecLimit acquire processing {} with id {}", task.getReferenceTaskName(), task.getTaskId());
+			WorkflowTask wfTask = task.getWorkflowTask();
+			if(enableGlobalTaskConcurrentExecLimit && wfTask.getGlobalConcurrentExecutionLimit() > 0) {
+				logger.debug("acquiring semaphore " + executionLimitingSemaphoreName + task.getReferenceTaskName());
+				if(!semaphoreDAO.tryAcquire(executionLimitingSemaphoreName + task.getReferenceTaskName(), task.getTaskId(), wfTask.getGlobalConcurrentExecutionLimit(), taskDefinition.get().getTimeoutSeconds() * 1000.0)){
```
apanicker-nflx, line 244: Since the semaphore uses task reference names to constrain concurrency limits, it breaks the isolation construct among independent tasks. Since the same task reference name can be used by different tasks in different workflows, setting the global limit for one would impact the execution of an unrelated task in an unrelated workflow, which is an unwanted side-effect.

kishorebanala:
> @kishorebanala I'm not entirely sure what you mean by how it ties together, but the order of checks in `exceedsInProgressLiint` are as follows:
> 
> 1. `TaskDef` concurrent execution limit
> 2. `WorkflowTask` execution limit (what I have added)
> 3. `TaskDef`rate limit check
>    Checks 1 and 2 are completely separate checks.

@TwoUnderscorez Given that Conductor already provides Task Concurrent execution limits and Task rate limits which is currently in use, how does this new addition fit into to the existing terminology? How can a new user discern between all the concurrency limit constructs available?

Can you provide a concrete example when the existing concurrency limits wouldn't be helpful, and what additional use cases this new feature would help with please?

core/src/main/java/com/netflix/conductor/validations/WorkflowTaskTypeConstraint.java
```diff
@@ -98,11 +98,43 @@ public boolean isValid(WorkflowTask workflowTask, ConstraintValidatorContext con
...
+            if (task == null) {
+                if( workflowTask.getGlobalConcurrentExecutionLimit() + workflowTask.getLocalConcurrentExecutionLimit() != 0) {
+                    String message = String.format("workflowTask: %s must use an existing taskDef if *ConcurrentExecutionLimit is defined in the workflowTask.", workflowTask.getTaskReferenceName());
+                    context.buildConstraintViolationWithTemplate(message).addConstraintViolation();
+                }
+                return false;
+            }
+            boolean valid = true;
+            if (workflowTask.getGlobalConcurrentExecutionLimit() + workflowTask.getLocalConcurrentExecutionLimit() != 0
+                && task.getTimeoutSeconds() < 1)
```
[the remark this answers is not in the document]
TwoUnderscorez: This verifies that when enabling the local or global concurrent execution limit, the linked TaskDef has a timeout defined for in the semaphore. The semaphore requires a timeout parameter to ensure that no dead Taskld s are holding the semaphore.

core/src/main/java/com/netflix/conductor/validations/WorkflowTaskTypeConstraint.java
```diff
@@ -98,11 +98,43 @@ public boolean isValid(WorkflowTask workflowTask, ConstraintValidatorContext con
...
                     break;
+                case TaskType.TASK_TYPE_SIMPLE:
+                    valid = isSimpleTaskValid(workflowTask, context);
+                    break;
             }
 
             return valid;
         }
 
+        private boolean isSimpleTaskValid(WorkflowTask workflowTask, ConstraintValidatorContext context) {
```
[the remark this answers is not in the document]
TwoUnderscorez: My mistake. I was thinking of adding this capability to only `SIMPLE` tasks but I guess it could be added to other task types, for example the Http task. What would you prefer? Should I add the validations to other tasks? If so, which tasks do you think should be added and which tasks should not be supported?

core/src/main/java/com/netflix/conductor/dao/SemaphoreDAO.java
```diff
@@ -0,0 +1,44 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.conductor.dao;
+
+import com.netflix.conductor.common.metadata.tasks.Task;
+
+/**
+ * An abstraction for a non-blocking semaphore object
```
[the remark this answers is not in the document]
TwoUnderscorez: What do you mean? Should I add better documentation in the code? It doesn't regulate task execution by itself, it can be used to do that.

redis-persistence/src/main/java/com/netflix/conductor/dao/dynomite/RedisExecutionDAO.java
TwoUnderscorez: I understand your concern here but l'm aware and that's exactly what I want to happened. In our system all the `taskReferenceName` s are controlled by another microservice and that's how we can guarantee zero conflicts. This change won't impact any currently defined workflows, not in our system, nor in anyone else's system because the default value for the limits is 0 and when it's 0 my code is basically skipped. The idea here is that if you are familiar with all the taskReferenceName in your system, you could enable this feature if you need to.

TwoUnderscorez:
@kishorebanala 
This is fundamentally different from TaskDef exec limits and rate limits. This limits the amount of WorkflowTaks that can run concurrently, that is, the ones defined in the WorkflowDef JSON. I admit that the name I chose is probably confusing, I'm more that open to suggestions. I believe that a good documentation on this subject can help a user differentiate all the limits from one another, which I will add after all code review is done, mainly to prevent updating the same thing twice. In our system we use k8s to host the workers, so changing the pod replica count is basically the same as setting the TaskDef execution limit. Whereas what I've added enables a more fine gained control over WorkflowTasks concurrent execution. You might also want to have a look at the responses I have given to the CR above.

redis-persistence/src/main/java/com/netflix/conductor/dao/dynomite/RedisExecutionDAO.java
apanicker-nflx: > In our system all the taskReferenceName s are controlled by another microservice and that's how we can guarantee zero conflicts.
The idea here is that if you are familiar with all the taskReferenceName in your system, you could enable this feature if you need to.

This is more of an exception than the norm. 

The contract that Conductor requires from tasks is to have unique names across the system and for workflow tasks to have unique reference names within a given workflow. However, there is no restriction for workflow tasks to be unique across workflows. 

> because the default value for the limits is 0 and when it's 0 my code is basically skipped.

While this is true, this feature now becomes unusable to any other user because it is not a constraint that is enforced by Conductor. Also, if a user were to start using this feature, it would unknowingly cause side-effects in their other workflows because unlike the task name validation constraint or the unique task reference name validation constraint, there is no validation preventing users from re-using the same task reference name across workflows.

core/src/main/java/com/netflix/conductor/validations/WorkflowTaskTypeConstraint.java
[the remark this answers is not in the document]
TwoUnderscorez: @apanicker-nflx
I understand. 
What do you think about adding another string field to use as the semaphore name instead of the `taskReferenceName`? How about `concurrentLockName` or `concurrentSemaphoreName`?

redis-persistence/src/main/java/com/netflix/conductor/dao/dynomite/RedisExecutionDAO.java
TwoUnderscorez: @apanicker-nflx
I understand. 
What do you think about adding another string field to use as the semaphore name instead of the `taskReferenceName`? Say `concurrentLockName` or `concurrentSemaphoreName`?

apanicker-nflx: @TwoUnderscorez Adding a new field for this purpose does not completely solve the isolation/side-effect argument mentioned above. Workflow owners for independent workflows could still use the same name within these params leading to issues.

Based on our internal discussion, we think it would be best to extract the ConcurrentExecutionLimits into a separate interface with the currently provided implementation as the default. The solution in this PR could be added as a pluggable implementation of this interface for users looking to solve concurrency limiting at a workflow task level. The caveat here being the external system that helps to keep task reference names unique across workflows.

Please let us know if you would be willing to undertake this interface extraction.

TwoUnderscorez: @apanicker-nflx I see.
So have `exceedsInProgressLiint` in all the execution DAOs call on that interface? The only problem with that is I'll run into the same problem as here #1951 because I want to add the `taskReferenceName` based looking - not replace it with the existing one. Basically call all the concurrent limiting implementations in an order specified by the user and if one returns false, stop and return false.

apanicker-nflx: > So have exceedsInProgressLiint in all the execution DAOs call on that interface? 

Yes, that interface would then provide the implementation for the `exceedsInProgressLimit`. 

> I'll run into the same problem as here #1951 because I want to add the taskReferenceName based looking - not replace it with the existing one.

Here again, you could add this as an extension of what is currently provided rather than as a replacement, similar to the status listener use case.

TwoUnderscorez: @apanicker-nflx Alright. So the `exceedsInProgressLiint` will do the task def limit check, then call the new interface, then rate limit check? 
Do you think we should open a issue to discuss this, because I have a few more questions like:
With this PR, I'll add my implementation of the interface, correct?

If so, I'll need a semaphore, is the DAO fine? Is the implementation fine?

I don't think that the task def concurrent execution limit implementation is not as reliable as I would like it to be if I were using it. Should I switch it to use a semaphore? (Probably a different PR though)

Could you please provide me the definition you were thinking about for that interface? Just to make sure we don't have a misunderstanding of some minor detail that ends up not being minor at all.

apanicker-nflx: That is a great idea, let's continue this discussion in a new issue rather than as comments in this PR. It would provide visibility to the problem you are trying to solve to the community and we can gather more inputs. Thanks.

TwoUnderscorez: @apanicker-nflx I've created issue #2066 .

Issue #1892, opened 2020-09-30
Unimplemented QueueDAO.containsMessage() for mysql & postgres


willqiao:
Hi Conductor Team,

We are using conductor with postgres database(previously, we were using mysql). After switching to both of the DBs, we found there is a WorkflowRepairService trying to call QueueDAO.containsMessage(). However, both Mysql and Postgres impl doesnt have this containsMessage() implemented. Logs will constantly throws following exceptions. I tried to put a 
workflow.repairservice.enabled=false in config.properties, hoping it can disable this function, but couldnt. 

I'm asking 2 questions here:
1. This exception doesnt seems a blocker for using any workflow API. Do we know what this WorkflowRepairService does?
2. Is it on purpose to not implement this containsMessage() for Mysql/postgres implementation? Any reason behind it? If not, do we have an estimation when we can address this un-finished method?

Thanks.
Will

`421529 [pool-16-thread-2] ERROR com.netflix.conductor.core.execution.WorkflowSweeper  - Error running sweep for 87614879-4935-4a55-9a05-b287f532288f
java.lang.UnsupportedOperationException: Please ensure your provided Queue implementation overrides and implements this method.
	at com.netflix.conductor.dao.QueueDAO.containsMessage(QueueDAO.java:199)
	at com.netflix.conductor.core.execution.WorkflowRepairService.verifyAndRepairTask(WorkflowRepairService.java:118)
	at com.netflix.conductor.core.execution.WorkflowRepairService.lambda$verifyAndRepairWorkflowTasks$2(WorkflowRepairService.java:84)
	at java.util.ArrayList.forEach(ArrayList.java:1257)
	at com.netflix.conductor.core.execution.WorkflowRepairService.verifyAndRepairWorkflowTasks(WorkflowRepairService.java:84)
	at com.netflix.conductor.core.execution.WorkflowSweeper.lambda$sweep$1(WorkflowSweeper.java:109)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)`

kishorebanala:
[...]

PR #1894, opened 2020-09-30
implemente QueueDAO.containsMessage() for mysql and postgres. issue 1892

Yingsheng-eroad:


Yingsheng-eroad:
https://github.com/Netflix/conductor/issues/1892

PR #1895, opened 2020-09-30
Fix the binding for ServerModule to the root SystemProps Configuration

demichej:
I think this is an issue that has always existed, but hasn't mattered until now since the `DynomiteConfiguration` is now directly overriding the default behavior of `SystemPropertiesConfiguration`. 

The new change to repair workflows is supposed to default to false, but currently it's always true because all configurations were being set to `SystemPropertiesDynomiteConfiguration`. 

I think that this change is what the intention was, and now makes the default actually be false. 

**NOTE:** This tries to make the property set to true by default in the various pre-existing config properties. But, this setting needs to be only controlled by one of these system properties to work correctly.

I think this was made recently by @kishorebanala

### DIFF
--- a/server/src/main/java/com/netflix/conductor/bootstrap/BootstrapModule.java
+++ b/server/src/main/java/com/netflix/conductor/bootstrap/BootstrapModule.java
@@ -8,6 +8,5 @@ public class BootstrapModule extends AbstractModule {
     @Override
     protected void configure() {
         bind(Configuration.class).to(SystemPropertiesConfiguration.class);
-        bind(ModulesProvider.class);
     }
 }
--- a/docker/server/config/config-local.properties
+++ b/docker/server/config/config-local.properties
@@ -24,6 +24,10 @@ workflow.namespace.queue.prefix=conductor_queues
 # No. of threads allocated to dyno-queues (optional)
 queues.dynomite.threads=10
 
+# By default with dynomite, we want the repairservice enabled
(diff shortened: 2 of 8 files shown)


server/src/main/java/com/netflix/conductor/bootstrap/BootstrapModule.java
```diff
@@ -8,6 +8,5 @@
     @Override
     protected void configure() {
         bind(Configuration.class).to(SystemPropertiesConfiguration.class);
-        bind(ModulesProvider.class);
```
demichej, line 11: This bind wasn't doing anything, and is not needed to create an instance of ModulesProvider

`ModulesProvider modulesProvider = bootstrapInjector.getInstance(ModulesProvider.class);`

demichej:
I see a lot of PRs failing with 

```
com.netflix.conductor.contribs.metrics.LoggingMetricsModuleTest > testCollector FAILED
    org.mockito.exceptions.verification.TooManyActualInvocations at LoggingMetricsModuleTest.java:45
```

but it works locally for me @kishorebanala . Not sure what's going on with that test or if y'all are aware.

demichej:
@kishorebanala Any update here? Thanks.

apanicker-nflx (approved):

kishorebanala:
Thanks for contributing this @demichej !

PR #672, opened 2018-07-22
Adding Kafka support for eventing and event handling

preeth-gopalakrishnan:
Two changes:
1. Minor correction in the event handling logic when using MySQL persistence.
2. Adding Kafka support in the contribs module.

kafka configuration for provider and consumer can be added with the below prefixes for general, producer and consumer configuration respectively.
'kafka.' , 'kafka.provider.' and 'kafka.consumer.'

request for review

related enhancement discussed in #414

PR #1911, opened 2020-10-05
Add Kafka as an event source

rickfish:
Allow messages to be published to a Kafka topic. That Kafka topic will be used as a source for generating Conductor events to start workflows, fail tasks and complete tasks. The original implementation was taken from PR #672 which was never completed and code was added to handle secure topics, error processing and multiple consumers per topic. 

Documentation on how to use it is in docs/docs/kafkaeventsource.md.


rickfish:
Note that one test failed that has nothing to do with my changes: com.netflix.conductor.test.integration.ExternalPayloadStorageSpec. Not sure what to do about that.

contribs/src/main/java/com/netflix/conductor/contribs/queue/kafka/KafkaObservableQueue.java
```diff
@@ -0,0 +1,519 @@
...
+							 * Teams do not have to set up an error topic if they don't want to.  
+							 */
+						} else {
+							logger.error("Publish message to kafka topic {} failed with an error: {}", errorQueueName, e.getMessage(), e);
+							throw new ApplicationException(ApplicationException.Code.INTERNAL_ERROR, "Failed to publish the event");
+						}
+					}
+					this.errorEmailSender.newError(this.queueName, validErrorTopic ? errorQueueName : null, notificationEmailAddress, theFailure);
+				} catch(Exception e) {
+					e.printStackTrace();
```
aravindanr, line 461: If the intention is to swallow the exception, please log it at the `WARN` level.

contribs/src/main/java/com/netflix/conductor/contribs/queue/kafka/KafkaObservableQueue.java
```diff
@@ -0,0 +1,519 @@
...
+			}
+		}
+		logger.info("Messages published to kafka topic {}. count {}", queueName, messages.size());
+
+	}
+
+
+	@Override
+	public void processFailures(List<Message> messages, EventProcessingFailures failures) {
+		processFailures(messages, failures.getTransientFailures());
```
aravindanr, line 423: The pattern here seems to be similar to DLQ (Dead Letter Queue). Transient failures are not acked, which means they are intended to be retried. Should those be pushed to the "error queue" as well?

contribs/src/main/java/com/netflix/conductor/contribs/kafka/KafkaErrorEmailSender.java
```diff
@@ -0,0 +1,301 @@
+package com.netflix.conductor.contribs.kafka;
```
aravindanr, line 1: Understand the reason for this. But this doesn't belong in Conductor. If an email needs to be sent when event processing fails, that should be handled externally. For instance, a listener on the "errorQueue" that KafkaObservableQueue.processFailures posts to could send that email.

core/src/main/java/com/netflix/conductor/core/utils/EmailSender.java
```diff
@@ -0,0 +1,155 @@
+package com.netflix.conductor.core.utils;
```
aravindanr, line 1: Please see comment in KafkaErrorEmailSender. Sending emails is not Conductor's responsibility and should be achieved by other means.

aravindanr (changes requested):
Please add integration tests. Kafka could be run in embedded mode for the tests.

contribs/src/main/java/com/netflix/conductor/contribs/queue/kafka/KafkaObservableQueue.java
rickfish: @aravindanr The intent is that by not acking the offsets are not bumped up so the consumer will automatically 'retry' because it will get the document at the old current offset (the same document that had the transient failure).

contribs/src/main/java/com/netflix/conductor/contribs/kafka/KafkaLoginModule.java
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * Java security module used to provide JAAS username and password for access to a Kafka topic
+ * @author rickfish
+ *
+ */
+public class KafkaLoginModule implements LoginModule {
+
+    String KAFKA_EVENTS_JAAS_USERNAME_PROPERTY_NAME = "kafka.events.jaas.username";
+    String KAFKA_DEFAULT_JAAS_USERNAME_PROPERTY_NAME = "kafka.default.jaas.username";
```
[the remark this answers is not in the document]
rickfish: Forgot to use the constants. Fixed.

contribs/src/main/java/com/netflix/conductor/contribs/kafka/KafkaLoginModule.java
```diff
@@ -0,0 +1,131 @@
...
+	@SuppressWarnings("unchecked")
+	@Override
+	public void initialize(Subject subject, CallbackHandler callbackHandler, Map<String, ?> sharedState, Map<String, ?> options) {
+		this.sharedState = (Map<String, Object>) sharedState;
+		this.options = (Map<String, Object>) options;
+	}
+
+	@Override
+	public boolean login() throws LoginException {
+		String userName = getProperty("workflow.elasticsearch.kafka.jaas.username", "svc-sioetdu");
```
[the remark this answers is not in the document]
rickfish: Forgot to use the constants. Fixed.

contribs/src/main/java/com/netflix/conductor/contribs/kafka/KafkaErrorEmailSender.java
rickfish: Got rid of the email functionality and replaced it with a protected method that can be overridden to provide this or other functionality.

core/src/main/java/com/netflix/conductor/core/utils/EmailSender.java
rickfish: Got rid of this class

contribs/src/main/java/com/netflix/conductor/contribs/kafka/KafkaLoginModule.java
```diff
@@ -0,0 +1,131 @@
...
+import javax.security.auth.spi.LoginModule;
+
+import org.apache.commons.lang3.StringUtils;
+
+/**
+ * Java security module used to provide JAAS username and password for access to a Kafka topic
+ * @author rickfish
+ *
+ */
+public class KafkaLoginModule implements LoginModule {
```
aravindanr, line 20: @rickfish Can you elaborate how this class will be leveraged to login into Kafka? Conductor runs on Jetty by default . Is this class expected to be called by the Java application server?

rickfish: This is used by JAAS. In the init() method of  KafkaObservableQueue it will load  a JAAS config file if it is  specified as a property. If specified, you reference the login module to use to get the userid/pw. See contribs/src/main/resources for examples of  these JAAS config files. I think security was left out of the other Kafka stuff because it is so convoluted.

rickfish:
@aravindanr somehow I clicked on something that got rid of your question about the KafkaLoginModule class. Sorry, not sure what happened.

The KafkaLoginModule class can  be referenced in a JAAS configuration file to allow adding JAAS security for access to the Kafka topic. Please see example files under src/main/resources in contribs. The 'kafka.default.jaas.config.file' or 'kafka.events.jaas.config.file' property can be specified that will point to the location of the JAAS config file and that JAAS config file can include a reference to the KafkaLoginModule class which  will allow for secure access to the topic. I don't know of any other  way to provide secure access although i have only used Kafka at my current employer and that is the way they require that security be done for Kafka.

PR #1921, opened 2020-10-12
Adding resiliency tests and relevant changes for QueueDAO.

kishorebanala:


core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -514,7 +514,23 @@ public void rewind(String workflowId, boolean useLatestDefinitions) {
...
+        try {
+            executionDAOFacade.createWorkflow(workflow);
+        } catch (Exception e) {
+            Monitors.recordWorkflowStartError(workflowDef.getName(), WorkflowContext.get().getClientApp());
+            LOGGER.error("Unable to restart workflow: {}", workflowDef.getName(), e);
+
+            // It's possible the terminate workflow call hits an exception as well, in that case we want to log both
+            // errors to help diagnosis.
+            try {
+                terminateWorkflow(workflowId, "Error when restarting the workflow");
```
apanicker-nflx, line 527: I am not convinced that a ```try...catch``` block here is the right way to address this. The exception when terminating a workflow should be caught at a granular level within the terminate logic and this should be logged/handled accordingly.

test-harness/src/test/groovy/com/netflix/conductor/test/resiliency/QueueResiliencySpec.groovy
```diff
@@ -0,0 +1,557 @@
+package com.netflix.conductor.test.resiliency
```
apanicker-nflx, line 1: Please add license header.

test-harness/src/test/groovy/com/netflix/conductor/test/resiliency/QueueResiliencySpec.groovy
```diff
@@ -0,0 +1,557 @@
...
+        def pollResult = taskResource.poll("integration_task_1", "test", "")
+
+        then:
+        1 * queueDAO.pop(*_) >> { throw new IllegalStateException("Queue pop failed from Spy") }
+        0 * queueDAO._
+        notThrown(Exception)
+        pollResult == null
+    }
+
+    def "Verify updateTask succeeds when QueueDAO is unavailable"() {
```
apanicker-nflx, line 420: Do you want to add another test case for updateTask with IN_PROGRESS state? 
This case only covers the updateTask with COMPLETED state, which does not directly use the queue, but the former uses queue to postpone.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
kishorebanala: This exception is primarily for `executionDAOFacade.createWorkflow(workflow)`, and `terminateWorkflow` is just a clean up as part of that exception. I'd still like to throw the exception for `createWorkflow`, irrespective of what happened with `terminateWorkflow`, while logging `terminateWorkflow` exception. Thoughts?

apanicker-nflx: Yes, the exception handling for createWorkflow needs to be handled. 
My comment is regarding the extended handling for the terminate part which seems out of place here. Something like this - 


```java
try {
    executionDAOFacade.createWorkflow(workflow);
} catch (Exception e) {
    Monitors.recordWorkflowStartError(workflowDef.getName(), WorkflowContext.get().getClientApp());
    LOGGER.error("Unable to restart workflow: {}", workflowDef.getName(), e);
    terminateWorkflow(workflowId, "Error when restarting the workflow");
    throw e;
}
```

test-harness/src/test/groovy/com/netflix/conductor/test/resiliency/QueueResiliencySpec.groovy
```diff
@@ -0,0 +1,605 @@
...
+        task.status == Task.Status.IN_PROGRESS
+        task.taskType == 'integration_task_1'
+
+        when: "the above task is updated, while QueueDAO is unavailable"
+        def taskResult = new TaskResult(task)
+        taskResult.setStatus(TaskResult.Status.IN_PROGRESS)
+        taskResult.setCallbackAfterSeconds(120)
+        def result = taskResource.updateTask(taskResult)
+
+        then: "updateTask returns successfully without any exceptions"
```
apanicker-nflx, line 496: update the text here?

test-harness/src/test/groovy/com/netflix/conductor/test/resiliency/QueueResiliencySpec.groovy
```diff
@@ -0,0 +1,605 @@
...
+        2 * queueDAO.postpone(*_) >> { throw new IllegalStateException("Queue postpone failed from Spy") }
```
apanicker-nflx, line 497: Why is postpone invoked twice? Given that we retry twice, shouldn't this be 3?

kishorebanala: The RetryUtil we're using here retries up to the provided number, which should be greater than >=1.

apanicker-nflx (approved):

PR #1931, opened 2020-10-21
Make ApplicationException extensible

andrea11:
Small change to give access to __code__ inside sub-classes


core/src/main/java/com/netflix/conductor/core/execution/ApplicationException.java
```diff
@@ -46,11 +46,11 @@ public int getStatusCode(){
 		}
 	}
 	
-	private Code code;
+	protected Code code;
```
kishorebanala, line 49: Why not use `getCode()` method from L#78 instead of changing this to `protected`?

andrea11: Indeed, I think that it is not needed

PR #1939, opened 2020-10-28
Preserve task domains for tasks inside do_while loop

marosmars:
On every iteration, apply task domains as set in the workflow execution.
This feature was missing and made do wile behave incorrectly when using
task domains (since the domains were reset for tasks inside the loop).

Signed-off-by: Maros Marsalek <mmarsalek@frinx.io>


marosmars:
Hey @apanicker-nflx , this is a small improvement (single line changed) to the do while task, including unit tests.

apanicker-nflx (approved):

Issue #1928, opened 2020-10-20
TERMINATE workflowOutput not overriding the outputParameters in workflow.


vamsibandarupalli:
I have a basic workflow which have one decision task, for decision when first condition is met I have TERMINATE task, on all other conditions I want to execute some other tasks.

For TERMINATE tasks "workflowOutput", I'm giving some static string output. And for workflows "outputParameters" key, I've given other tasks output.

When TERMINATE task is executed, I'm expecting just the static output that assigned to "workflowOutput" key, but I'm getting output from what I defined in "outputParameters".

Example Workflow definition:
{
    "name": "netflix_identation_terminate",
    "description": "Adds Netflix Identation to video files.",
    "version": 2,
    "schemaVersion": 2,
    "tasks": [{
            "name": "verify_if_idents_are_added",
            "taskReferenceName": "ident_verification",
            "inputParameters": {
                "contentId": "${workflow.input.contentId}"
            },
            "type": "SIMPLE"
        }, {
            "name": "decide_task",
            "taskReferenceName": "is_idents_added",
            "inputParameters": {
                "case_value_param": "${ident_verification.output.is_idents_added}"
            },
            "type": "DECISION",
            "caseValueParam": "case_value_param",
            "decisionCases": {
                "false": [{
                        "name": "add_idents",
                        "taskReferenceName": "add_idents_by_type",
                        "inputParameters": {
                            "identType": "${workflow.input.identType}",
                            "contentId": "${workflow.input.contentId}"
                        },
                        "type": "SIMPLE"
                    }
                ]
            },
            "defaultCase": [{
                    "name": "terminate",
                    "taskReferenceName": "terminate0",
                    "inputParameters": {
                        "terminationStatus": "COMPLETED",
                        "workflowOutput": "${ident_verification.output}"
                    },
                    "type": "TERMINATE",
                    "startDelay": 0,
                    "optional": false
                }
            ]
        }
    ],
    "outputParameters": {
        "result": "${ident_verification.output}",
        "add_idents_by_type": "${add_idents_by_type.output}"
    }
}

I need just the "workflowOutput" when TERMINATE task executed. Instead getting whatever I mentioned in "outputParameters"

apanicker-nflx:
@vamsibandarupalli Thanks for reporting this. Upon initial investigation, it seems that the `outputParameters` in the workflow definition will overwrite the workflow output that was set from within the TERMINATE task.
[...]

PR #1945, opened 2020-10-31
Fix for bug 1928, TERMINATE workflowOutput is being override by workflowOutput parameters set from workflowDefinition

vamsibandarupalli:
Reference [1928 issue](https://github.com/Netflix/conductor/issues/1928).

Added fix to not to override the workflow output that was set in TERMINATE task.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/tasks/Terminate.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/tasks/Terminate.java
@@ -81,9 +81,6 @@ public class Terminate extends WorkflowSystemTask {
     private void setWorkflowOutput(Map<String, Object> taskOutput, Workflow workflow) {
         if(!taskOutput.isEmpty()) {
             workflow.setOutput(taskOutput);
-			if (workflow.getWorkflowDefinition() != null && (workflow.getWorkflowDefinition().getOutputParameters() != null && !workflow.getWorkflowDefinition().getOutputParameters().isEmpty())) {
-                workflow.getWorkflowDefinition().setOutputParameters(taskOutput);
-            }
         }
     }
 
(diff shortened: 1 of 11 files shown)


core/src/main/java/com/netflix/conductor/core/execution/tasks/Terminate.java
```diff
@@ -81,6 +81,9 @@ public static Boolean validateInputStatus(String status) {
     private void setWorkflowOutput(Map<String, Object> taskOutput, Workflow workflow) {
         if(!taskOutput.isEmpty()) {
             workflow.setOutput(taskOutput);
+			if (workflow.getWorkflowDefinition() != null && (workflow.getWorkflowDefinition().getOutputParameters() != null && !workflow.getWorkflowDefinition().getOutputParameters().isEmpty())) {
```
apanicker-nflx, line 84: Changing the workflow definition within the workflow does not seem to be the best way to achieve this. This makes the behavior seem like a side-effect rather than intentional.

apanicker-nflx:
@vamsibandarupalli Thank you for your contribution.
Please also add an integration test [here](https://github.com/Netflix/conductor/blob/master/test-harness/src/test/groovy/com/netflix/conductor/test/integration/LambdaAndTerminateSpec.groovy) as part of this change.

core/src/main/java/com/netflix/conductor/core/execution/tasks/Terminate.java
vamsibandarupalli: Agreed, but when called updateWorkflowOutput() in deciderService, Checking the workflowDefintion's outputParameters rather than workflow output which has changed in Terminate.java.

vamsibandarupalli: Removed the code change which modifies workflow definition and added check in DeciderService which seems the better way to fix the bug.

apanicker-nflx (approved):

Issue #1164, opened 2019-05-28
Question: Wishlist #8 Validating Task inputKeys and outputKeys


cclulu:
I was hoping I would be able to contribute to this open source project, and I wanted to ask if what I had planned for the validation would be enough or was there more to the validation that was intended for this. What I had planned is simply adding a `@NotEmpty(message = "")` On the [inputkeys](https://github.com/Netflix/conductor/blob/master/common/src/main/java/com/netflix/conductor/common/metadata/tasks/TaskDef.java#L72) and [outputKeys](https://github.com/Netflix/conductor/blob/master/common/src/main/java/com/netflix/conductor/common/metadata/tasks/TaskDef.java#L75) fields on [TaskDef.java](https://github.com/Netflix/conductor/blob/master/common/src/main/java/com/netflix/conductor/common/metadata/tasks/TaskDef.java) file.My reasoning for doing something so simple is because in your [documentation](https://netflix.github.io/conductor/configuration/taskdef/#using-inputkeys-and-outputkeys) on the fourth bullet point it states that 

> In the future, this can be extended to be a strict template that all task implementations must adhere to ...

My understanding was that users creating tasks would need to have these fields populated to be able to document what is needed for a task to be run and what you can expect it to output once finished, making it easier for re-use. Looking forward to your response, Thank you.

naveenchlsn:
We are looking for this feature as well. @cclulu  I think we need to think through some details. 

* How do we ensure backward compatibility when we introduce validation?
* How to have strict validation? For example, the data structure of an input/output key. 


@cyzhao @apanicker-nflx @mashurex @kishorebanala @mudit3774 Thoughts?

cclulu:
[...]

PR #1959, opened 2020-11-05
Feature: [WIP] add typing and validation for input/output

andrea11:
Hello, based on this discussion #1164, I would like to contribute by providing a way to define a better structure for task and workflow input/output, adding the possibility to validate them. I decided to opt for json schema to define the structure, as it is meant exactly for validation. For the implementation in Java, I decided to use [everit-org](https://github.com/everit-org/json-schema).
I decided to create this PR in order to collect feedback.

Here the list of steps that I want to follow:
1. Adding new field for input/output for task and workflow, and marking old one as deprecated - DONE
2. Adding configuration to enable/disable validation - DONE
3. Performing a validation on the Json Schema provided by the user inside the WorkflowDef/MetadataTask - DONE
4. Validate workflow input at creation time, preventing its creation in case of failure - DONE
5. Validate task input at scheduling time, failing the entire workflow in case of failure - ONGOING
6. Validate task output at updating time, rejecting the task if it is not compliant - ONGOING
7. Validate workflow output once the last task is completed, failing the workflow if the output is not valid - ONGOING
8. Add support for custom validators (cf. https://github.com/everit-org/json-schema#example) 

Please do not hesitate to provide feedback

### DIFF
--- a/core/src/main/java/com/netflix/conductor/service/WorkflowValidatorImpl.java
+++ b/core/src/main/java/com/netflix/conductor/service/WorkflowValidatorImpl.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.conductor.service;
+
+import com.netflix.conductor.common.metadata.workflow.WorkflowDef;
+import com.netflix.conductor.common.run.Workflow;
+import com.netflix.conductor.core.execution.ValidationException;
+import org.everit.json.schema.Schema;
+import org.everit.json.schema.loader.SchemaLoader;
+import org.json.JSONObject;
+
+import java.util.List;
+
+
+public class WorkflowValidatorImpl implements WorkflowValidator {
+	@Override
+	public void validate(WorkflowDef workflowDef, Workflow workflow) {
+		if (workflowDef.getInputDefinition().isEmpty()) {
+			return;
+		}
+		final JSONObject jsonSchema = new JSONObject(workflow.getInput());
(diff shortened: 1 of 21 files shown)


core/src/main/java/com/netflix/conductor/service/WorkflowValidatorImpl.java
```diff
@@ -0,0 +1,47 @@
...
+import org.json.JSONObject;
+
+import java.util.List;
+
+
+public class WorkflowValidatorImpl implements WorkflowValidator {
+	@Override
+	public void validate(WorkflowDef workflowDef, Workflow workflow) {
+		if (workflowDef.getInputDefinition().isEmpty()) {
+			return;
```
Jiehong, line 32: what if the `workflow` has some input? shouldn't this be invalid? (as no input is expected)

core/src/main/java/com/netflix/conductor/service/WorkflowValidatorImpl.java
```diff
@@ -0,0 +1,47 @@
...
+		}
+		final JSONObject jsonSchema = new JSONObject(workflow.getInput());
+		final JSONObject jsonSchemaDefinition = new JSONObject(workflowDef.getInputDefinition());
+		final Schema schema = SchemaLoader.builder().draftV7Support()
```
Jiehong, line 36: Could the latest draft be used instead? (See in https://json-schema.org/, `16 September 2019: Draft 2019-09 (formerly known as draft-08) has been published!`)

Jiehong:
@andrea11 : very interesting PR!

core/src/main/java/com/netflix/conductor/service/WorkflowValidatorImpl.java
elisherer: I think most toolset still only support up to draft-07, having said that, the code could be changed in the future to support more than one version.

core/src/main/java/com/netflix/conductor/service/WorkflowValidatorImpl.java
andrea11: This line was used to avoid validation in case of old workflows that are not yet using the inputDefinition. However, after some tests, I realized that you need to explicitly define your object in the JSON schema with 
`"additionalProperties": false`, if you want to mark the input as invalid if any non-specified parameter is present inside your object. So, we could think to add automatically this property to the root object (___input___), but I think the best would be to give the freedom to the end-user to define its own workflow

core/src/main/java/com/netflix/conductor/service/WorkflowValidatorImpl.java
andrea11: Indeed, as @elisherer already explained, draft-08 is not supported by this library (yet)

Issue #1961, opened 2020-11-06
Metadata APIs return 500 instead of 409/404 when using Postgres


michaelpaliy:
We started to use Postgres instead of Dynomite and found some incorrect response codes when using the Metadata API.

Trying to create workflow that already exists:
```
{
    "code": "BACKEND_ERROR",
    "message": "BACKEND_ERROR - Workflow with my_wf_1.1 already exists!",
    "retryable": true,
    "instance": "d1-conductor-server-5b86c7687b-g2rnc"
}
```
Trying to delete workflow that not exists:
```
{
    "code": "BACKEND_ERROR",
    "message": "BACKEND_ERROR - No such workflow definition: sdfsdf version: 5",
    "retryable": true,
    "instance": "d1-conductor-server-5b86c7687b-g2rnc"
}
```

logs:
```
ERROR com.netflix.conductor.server.resources.ApplicationExceptionMapper  - Error ApplicationException url: 'metadata/workflow/'
com.netflix.conductor.core.execution.ApplicationException: BACKEND_ERROR - Workflow with my_wf_1.1 already exists!
        at com.netflix.conductor.dao.postgres.PostgresBaseDAO.getWithTransaction(PostgresBaseDAO.java:128)
        at com.netflix.conductor.dao.postgres.PostgresBaseDAO.lambda$getWithRetriedTransactions$3(PostgresBaseDAO.java:142)

Caused by: com.netflix.conductor.core.execution.ApplicationException: Workflow with my_wf_1.1 already exists!
        at com.netflix.conductor.dao.postgres.PostgresMetadataDAO.lambda$createWorkflowDef$1(PostgresMetadataDAO.java:105)
        at com.netflix.conductor.dao.postgres.PostgresBaseDAO.lambda$withTransaction$4(PostgresBaseDAO.java:193)
        at com.netflix.conductor.dao.postgres.PostgresBaseDAO.getWithTransaction(PostgresBaseDAO.java:122)
```

https://github.com/Netflix/conductor/blob/master/postgres-persistence/src/main/java/com/netflix/conductor/dao/postgres/PostgresBaseDAO.java#L127
It looks like it catching any exception and rethrowing it with 500 code instead with the original code.

If it something that you want to be fixed, I can fix and open PR.

aravindanr:
hi @michaelpaliy, thanks for logging the issue. Sending the right status codes is important, please create a PR with the fix.

PR #1970, opened 2020-11-15
Fix for bug #1961 - fix response codes in posgres and mysql

michaelpaliy:
fix for bug #1961

### DIFF
--- a/grpc-server/build.gradle
+++ b/grpc-server/build.gradle
@@ -13,6 +13,6 @@ dependencies {
     compile "log4j:log4j:1.2.17"
 
     testCompile "io.grpc:grpc-testing:${revGrpc}"
-    testCompile 'org.testinfected.hamcrest-matchers:all-matchers:1.8'
+    testCompile "org.testinfected.hamcrest-matchers:all-matchers:${revHamcrestAllMatchers}"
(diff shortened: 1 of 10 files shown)


grpc-server/build.gradle
```diff
@@ -13,6 +13,6 @@ dependencies {
     compile "log4j:log4j:1.2.17"
 
     testCompile "io.grpc:grpc-testing:${revGrpc}"
-    testCompile 'org.testinfected.hamcrest-matchers:all-matchers:1.8'
+    testCompile "org.testinfected.hamcrest-matchers:all-matchers:${revHamcrestAllMatchers}"
```
aravindanr, line 16: Can this version be updated in the `versionsOfDependencies.gradle`?

aravindanr (commented):
@picaron please review.

grpc-server/build.gradle
michaelpaliy: I moved this version to `versionsOfDependencies.gradle` file, because I'm using it also in the postgres and mysql modules.
Unfortnuly 1.8 is the latest: [all-matchers](https://mvnrepository.com/artifact/org.testinfected.hamcrest-matchers/all-matchers)

Issue #1849, opened 2020-08-28
Subworkflow doesn't return output


btafarelo:
Hi!

I made a subworkflow to organize some variables and there isn't an output but the workflow show the output correctly.

On my second screenshoot there is my last attempt in make it works, where I have set a hardcode variable.

Other subflows work fine but this one doesn't. This sub/workflow only has lambda tasks.

`{
      "name": "FORK_VARIABLES",
      "taskReferenceName": "FORK_VARIABLES",
      "type": "FORK_JOIN",
      "forkTasks": [
        [
          {
           "name": "sumSF",
           "taskReferenceName": "sumSF",
           "type": "LAMBDA",
           "inputParameters": {
               "sf": "${workflow.input.sf}",
               "scriptExpression": "var r=0; for(var i=0; i<$.sf.length; i++) r += $.sf[i].amount; return r;"
           }
          }
        ],
        [
          {
           "name": "sumIO",
           "taskReferenceName": "sumIO",
           "type": "LAMBDA",
           "inputParameters": {
               "io": "${workflow.input.io}",
               "scriptExpression": "var r=0; for(var i=0; i<$.io.length; i++) r += $.io[i].amount; return r;"
           }
          }
        ],
        [
          {
           "name": "sumTX",
           "taskReferenceName": "sumTX",
           "type": "LAMBDA",
           "inputParameters": {
               "tx": "${workflow.input.tx}",
               "rfe": "${workflow.input.rfe}",
               "scriptExpression": "var r=0; for(var i=0; i<$.tx.length; i++) if ($.rfe == true && $.tx[i].taxType == 'TC') continue; else if ($.tx[i].status == 'On') r += $.tx[i].amount; return r;"
           }
          }
        ],
        [
          {
            "name": "digitDocument",
            "taskReferenceName": "digitDocument",
            "type": "LAMBDA",
            "inputParameters": {
                "cpf": "${workflow.input.cpf}",
                "scriptExpression": "return $.cpf.charAt(5) + '' + $.cpf.charAt(6);"
            }
          }
        ],
        [
          {
            "name": "financingProdutoApplied",
            "taskReferenceName": "financingProdutoApplied",
            "type": "LAMBDA",
            "inputParameters": {
                "input": "${workflow.input.financingProductId}",
                "dp": "${workflow.input.dp}",
[...]

PR #1984, opened 2020-11-24
Subworkflow doesn't return output

sukruthiarjula:
Subworkflow doesn't return output #1849

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/tasks/SubWorkflow.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/tasks/SubWorkflow.java
@@ -114,7 +114,10 @@ public class SubWorkflow extends WorkflowSystemTask {
 			task.getOutputData().put(SUB_WORKFLOW_ID, subWorkflowId);
 
 			// Set task status based on current sub-workflow status, as the status can change in recursion by the time we update here.
-			Workflow subWorkflow = provider.getWorkflow(subWorkflowId, false);
+			Workflow subWorkflow = provider.getWorkflow(subWorkflowId, false);			
+			if (subWorkflow.getStatus() == WorkflowStatus.COMPLETED) {
+				task.getOutputData().putAll(subWorkflow.getOutput());
+			}
 			updateTaskStatus(subWorkflow, task);
 		} catch (Exception e) {
 			task.setStatus(Status.FAILED);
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/netflix/conductor/core/execution/tasks/SubWorkflow.java
```diff
@@ -114,7 +114,10 @@ public void start(Workflow workflow, Task task, WorkflowExecutor provider) {
 			task.getOutputData().put(SUB_WORKFLOW_ID, subWorkflowId);
 
 			// Set task status based on current sub-workflow status, as the status can change in recursion by the time we update here.
-			Workflow subWorkflow = provider.getWorkflow(subWorkflowId, false);
+			Workflow subWorkflow = provider.getWorkflow(subWorkflowId, false);			
+			if (subWorkflow.getStatus() == WorkflowStatus.COMPLETED) {
```
pweatherby-ccctc, line 118: @sukruthiarjula While this definitely gets the job done for the best-case scenario, there are other cases it misses. Consider how the code does this check/copy over of the payload as seen on lines 135 to 142.

Using the isTerminal function will cover more scenarios than just the COMPLETED scenario and then making sure to account for external payload storage is crucial as well.

Perhaps those lines (135-142) can be refactored out into a function of its own to be called by both the execute function and the start function. just a thought.

pweatherby-ccctc (changes requested):

core/src/main/java/com/netflix/conductor/core/execution/tasks/SubWorkflow.java
sukruthiarjula: Hi @pweatherby-ccctc,

As discussed, I have made the mentioned changes.

pweatherby-ccctc (approved):
Thank you so much for putting this together. Unfortunately, the build is now failing on a Logging Metric test. so once that gets looked at and adjusted, I think this will be good to go.

sukruthiarjula:
Build is failing on a Logging Metric test due to timeout issue. In local I could not replicate as all the 77 tests passed.
image

The code coverage is failing in CassandraMetadataDAO.java unrelated to the changes which I have made. Any inputs provided would be helpful.

apanicker-nflx (approved):

Issue #1806, opened 2020-07-24
PR #1748 causes a limit to the number of 'conductor' events that get polled


rickfish:
PR #1748 seems to limit the number of event types that are polled by the DynoObservableQueue implementation because of the new Scheduler that it uses. The new 'workflow.event.queue.scheduler.poll.thread.count' property puts a hard limit on the number of 'conductor' event types that get polled. I don't think that was the intent because adding a new event handler should not require that thread count to have to be changed, right? 

The symptoms are that, if this is set to 4 and there are 5 'conductor' event handlers, only 4 of those event types will be polled for and events of the the last event type will never get processed.

How should I be setting this property so that no matter how many event handlers I add it flexes to poll for all of them?

If it cant be set to be dynamic and grow, why not just use Schedulers.io() for the Observable like it was doing before? Seems like that fits the bill unless I don't know about a pitfall with that one.

apanicker-nflx:
@mdepak Can you please comment on this change?

mdepak:
Using the Execution scheduler does not limit the number of events that can be polled. In the implementation, there will be one instance of ObservableQueue implementation per event type. Singleton scheduler object will be used across all the observable queues and threads from this scheduler thread pool will be used for polling.

In earlier implementation Computation scheduler is used and that has upper bound on the number of threads in pool equal to the CPU cores. With the computation scheduler, there will more contention for the acquiring threads. As per docs, the computation scheduler is recommended for CPU intensive tasks, and polling from external queues is i/o bound task.

Using Schedulers.io() is not ideal here since, if one of the external queues is misbehaving, the scheduler might end up creating more threads while others are still waiting for the i/o. This can consume more heap space and might lead to OOM. With the Execution scheduler, one can control the number of threads that are used for polling from queues.

Ideally, if the number of threads is equal to the event types there will not be contention. But since the poll operation is not long-running, it should not affect having lesser thread count.

@apanicker-nflx  Do you think adding README documentation for the property will help? If so, I will add documentation for this property and similar ones. 

@rickfish  Please let me if there are any further questions.

rickfish:
[...]

PR #1987, opened 2020-11-29
Fix bug in the event polling with Rx Executor scheduler

mdepak:
Fixes the issue Netflix#1806 where the executor scheduler does not poll messages from certain queue types when the thread pool size is less than the number of queues to poll for.

As per conversion in https://github.com/Netflix/conductor/issues/1806#issuecomment-722015404, restoring default behavior and keeping the option to configure the thread pool size for the scheduler.


aravindanr:
@rickfish Does this change fix the issue #1806?

core/src/main/java/com/netflix/conductor/core/events/queue/EventPollSchedulerProvider.java
```diff
@@ -37,11 +37,14 @@ public EventPollSchedulerProvider(Configuration configuration) {
...
-
-        return Schedulers.from(executorService);
+        if (configuration.getEventSchedulerPollThreadCount() != null) {
+            ThreadFactory threadFactory = new ThreadFactoryBuilder()
+                    .setNameFormat("event-queue-poll-scheduler-thread-%d")
+                    .build();
+            Executor executorService = Executors.newFixedThreadPool(configuration.getEventSchedulerPollThreadCount(), threadFactory);
+            return Schedulers.from(executorService);
+        } else {
+            return Schedulers.computation();
```
aravindanr, line 47: @mdepak Based on the documentation here, http://reactivex.io/RxJava/javadoc/rx/schedulers/Schedulers.html#computation--

> Do not perform IO-bound work on this scheduler. Use io() instead.

and the fact that IO needs to be performed as seen [here](https://github.com/Netflix/conductor/blob/master/core/src/main/java/com/netflix/conductor/core/events/queue/dyno/DynoObservableQueue.java#L129-L137)

Does it make sense to use `io()` instead here?

mdepak: @aravindanr  `io()` is a good candidate but the potential problem with this is it can grow the thread pool exponentially in case the IO wait is taking too long because of a problem with the Queue system and could cause OOM. RxJava 2.x doc has a more detailed explanation regarding the same http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/schedulers/Schedulers.html#io--

The objective of this fix is to ensure that the event processing functionality works as expected by default without tuning up the thread pool size. Meanwhile, I plan to debug why [Executor scheduler](http://reactivex.io/RxJava/javadoc/rx/schedulers/Schedulers.html#from-java.util.concurrent.Executor-) is not working as expected, and its behavior in the later versions of RxJava. Please let me know your feedback.

rickfish:
@mdepak according to the documentation for Schedulers.computation():
 the number of threads, by default, is equal to the number of processors

again, when using OpenShift or any Kubernetes implementation I believe, each pod (jvm) is allocated logical CPUs so it could definitely have just one logical CPU so that would again mean that only one event could be processed, unless I am missing something...

apanicker-nflx:
@mdepak Can you please pick this back up and drive to completion as it seems to be affecting other users as well? Thanks.

mdepak:
@apanicker-nflx I will work on identifying the root cause and ensure it is fixed sooner. Apologies for the delay.

apanicker-nflx:
@mdepak Pinging again as a reminder.

mdepak:
The intent of this pull request is to ensure that the default behavior is maintained before my previous changes (https://github.com/Netflix/conductor/pull/1748) of introducing Execuution scheudler. The following version of Obervable.interval() method was invoked earlier from the ObservableQueue implementations and it is using compuation scheudler. 

```
public static Observable<Long> interval(long initialDelay, long period, TimeUnit unit) {
        return interval(initialDelay, period, unit, Schedulers.computation());
}
```

With this PR, when the config (`Configuration#getEventSchedulerPollThreadCount()`) is not provided (NULL), `Schedulers.computation()` is returned thus maintining the default behavior. I have verified with 100 types of event queues in my local with Schedulers.computation() scheduler and it works as expected. Whereas dedicated threadpool helps in high througput scenairos where each thread is allocated for one event type.  Could we merge this PR so that existing behavior is maintained before I dig deep to identify root cause to fix in Executor scheduler area as well?  Please let me know your thoughts on this.

Sidenote: Ideally, the obervables from interval() should share the threads from the passed scheduler object to poll from the queues. This happends as expected for Schedulers.computation(). In theory,Executor scheduler is expected to behave same way. But a thread from the Executor scheudler pool always polls a particular queue type instead of relinquishing it after a single call. This causes only few event types to be processed. RxJava version (1.2.2) is deprecated and I guess no support is available. I just started trying to migrate to latest RxJava 3.0.0 version to see if any bug is fixed. Apologies for the dealyed respones.

apanicker-nflx:
@mdepak Please resolve conflicts.

Issue #81, opened 2017-02-14
Null input for  task_30 in  kitchensink demo 


lydialmr120:
While running the kitchensink demo,the get_es_1 task completed successful,but the task_30's input value is null.

get_es_1 task Output:
[...]

PR #1993, opened 2020-12-02
Python Client: Error print statement now also prints exec_function name.

DPain:
Detecting errors happening when a workflow is running is somewhat difficult. When the workers log the problematic exec_function's name, it is much easier to detect the issue without having to view the Conductor UI.

### DIFF
--- a/client/python/conductor/ConductorWorker.py
+++ b/client/python/conductor/ConductorWorker.py
@@ -78,7 +78,7 @@ class ConductorWorker:
                 task['reasonForIncompletion'] = resp['reasonForIncompletion']
             self.taskClient.updateTask(task)
         except Exception as err:
-            print('Error executing task: ' + str(err))
+            print(f'Error executing task: {exec_function.__name__} with error: {str(err)}')
             task['status'] = 'FAILED'
             self.taskClient.updateTask(task)
 


client/python/conductor/ConductorWorker.py
```diff
@@ -78,7 +78,7 @@ def execute(self, task, exec_function):
                 task['reasonForIncompletion'] = resp['reasonForIncompletion']
             self.taskClient.updateTask(task)
         except Exception as err:
-            print('Error executing task: ' + str(err))
+            print(f'Error executing task: {exec_function.__name__} with error: {str(err)}')
```
kishorebanala, line 81: What does `exec_function.__name__` usually have? Can it be task name from the task object instead of a function name?

DPain: `exec_function.__name__` would contain the name of the exec_function that will be executed in the `start` function of `ConductorWorker.py`

Consider the snippet below:
`
self.start('task_name_in_conductor_server', self._function_to_execute, True)
`
This is the code for the worker to start poll for `task_name_in_conductor_server` in the conductor server. It will execute the function `self._function_to_execute` whenever `task_name_in_conductor_server` needs to be completed within conductor server.

My suggestion was to print the actual name of the function that will be executed in the python client (in this example above: `_function_to_execute`) instead of printing `task_name_in_conductor_server`, because whenever I view logs from the worker, it would be used for debugging the worker and the actual function name helps more than what the name of the task is inside conductor server. Alternatively if one wants to print the task name instead of function name, line #81 can be changed to `print(f'Error executing task: {task['taskType]} with error: {str(err)}')` instead. Though I would rather go to the Conductor Server UI to view the name of the task, so the name of the task within Conductor Server feels less helpful to be in the worker logs.

kishorebanala (approved):

Issue #2007, opened 2020-12-12
Do-While may case OOM


xumyselfcn:
some information:
1. exec a workflow within a do-while task.
2. do-while task need loop 200+ times.
3. vmopthion is -Xmx2500m -Xms1500m

then oom comes...

I think workflowExecutor.decide() deep recursive call is the reason of OOM. below is a simplified call logic.

![image](https://user-images.githubusercontent.com/1269523/101980887-b04b6900-3ca3-11eb-9d0c-ebfb2aeceb8f.png)

**1. each state change will cause a decide call
2. in decide method, will create a new workflow instance from database (includes task that contains)**

----

I fixed this bug with guava asyncEventBus...

PR #2009, opened 2020-12-13
avoid deep recursive WorkflowExecutor.decide calls, which may cause oom.

xumyselfcn:
fix [issues#2007 ](https://github.com/Netflix/conductor/issues/2007)

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/eventbus/DecideAsyncEventBus.java
+++ b/core/src/main/java/com/netflix/conductor/core/eventbus/DecideAsyncEventBus.java
@@ -1,0 +1,33 @@
+package com.netflix.conductor.core.eventbus;
+
+import com.google.common.eventbus.AsyncEventBus;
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/netflix/conductor/core/eventbus/DecideAsyncEventBus.java
```diff
@@ -0,0 +1,33 @@
...
+import com.google.common.eventbus.EventBus;
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+public class DecideAsyncEventBus {
+
+    private static ExecutorService executorService =
```
xumyselfcn, line 13: maybe its better to share executorService with ExecutionConfig.executorService

Issue #1997, opened 2020-12-08
PostgreSQL persistence: Unable to shard 'queue_message' table using Citus


raghavendra-bhat:
I am using PostgreSQL and trying to shard the tables using Citus extension on PostgeSQL. I have hit a road block as I am unable to shard 'queue_message' table (which I believe is the 'queue_name' column is the table that needs to be distributed in the Postgres cluster or have I erred here?). I get the below error when I try to distribute the table

postgres=# SELECT create_distributed_table('queue_message', 'queue_name');
ERROR: cannot create constraint on "queue_message"
DETAIL: Distributed relations cannot have UNIQUE, EXCLUDE, or PRIMARY KEY constraints that do not include the partition column (with an equality operator if EXCLUDE).

It appears that Citus doesn't allow sharding on column that's not part of UNQIUE or PRIMARY key constraints. This table has "id" as primary key. It would be of great help if someone can let me know if I am on the right track and how do I go about distributing data using PostgreSQL.

kishorebanala:
Hey @mactaggart, @marosmars Is this something you can help with? Thanks!

mactaggart:
@raghavendra-bhat I've not used Conductor with Citus, however, I suggest trying the 'id' column instead of 'queue_name' since it's the primary key. Other tables also use 'id' as the primary key, so they could be sharded by id as well. Let me know if that works.

raghavendra-bhat:
@mactaggart Nope, using 'id' also has the same issue as it needs to be part of unique key constraints. 

_postgres=# SELECT create_distributed_table('queue_message', 'id');
ERROR:  cannot create constraint on "queue_message"
DETAIL:  Distributed relations cannot have UNIQUE, EXCLUDE, or PRIMARY KEY constraints that do not include the partition column (with an equality operator if EXCLUDE)._

PR #2013, opened 2020-12-19
modify primary key on queue_message table

mactaggart:
This PR modifies the primary key on the queue_message table to be queue_name, message_id columns for postgres and mysql. 

Currently, on the queue_message table, there is a primary key of 'id' (SERIAL) and a unique index of 'queue_name' and 'message_id' columns. (both columns non-null) Since, AFAICT, the id column is unused, the index should replace id as the primary key. 

The motivation behind this PR is to enable support for sharding the queue_message table using Citus. (see #1997) In order to shard, the queue_message table cannot have a unique constraint or primary key that does not include the partition (shard) column. By removing the id pk/column and replacing the pk with the existing unique index, the queue_message table can be sharded in Citus. (by queue_name or message_id)

While Citus is postgres only, these changes seem appropriate for mysql as well.

@kishorebanala @raghavendra-bhat

### DIFF
--- a/mysql-persistence/src/main/resources/db/migration/V7__new_queue_message_pk.sql
+++ b/mysql-persistence/src/main/resources/db/migration/V7__new_queue_message_pk.sql
@@ -1,0 +1,24 @@
+# no longer need separate index if pk is queue_name, message_id
+SET @idx_exists := (SELECT COUNT(INDEX_NAME)
+                    FROM information_schema.STATISTICS
+                    WHERE `TABLE_NAME` = 'queue_message'
+                      AND `INDEX_NAME` = 'unique_queue_name_message_id'
+                      AND TABLE_SCHEMA = database());
+SET @idxstmt := IF(@idx_exists > 0, 'ALTER TABLE `queue_message` DROP INDEX `unique_queue_name_message_id`',
+                                    'SELECT ''INFO: Index unique_queue_name_message_id does not exist.''');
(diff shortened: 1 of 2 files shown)


kishorebanala:
This looks good to me. Just to have another pair of eyes, preferably from Postgres module users, can one of you help reviewing this? @raghavendra-bhat, @marosmars , @rickfish.. Thank you!

kishorebanala (approved):

rickfish:
@kishorebanala looks good to me. That id column was never used anyway and always had a potential to max out.

raghavendra-bhat:
@mactaggart I have a query, in general, on the use of 'id' columns in the tables. I think you might have a reason in using but I don't see this column being used in any of the execution and queue dao related tables queries (except for one instance where one of the queries use ordering by Id in task_in_progress table but even that could be replaced by created_on?).
@kishorebanala @rickfish

rickfish:
@mactaggart @raghavendra-bhat I agree, in fact it is a pain because there is always a chance of overflow, especially the way it used to be (insert/on conflict) which would increase the sequence even if there was already a row (on update). I would be in favor of getting rid of the id columns in general. I had thought about it but I thought at the time it would be a big change.

mactaggart:
@rickfish @raghavendra-bhat Agree we could remove the id column on several tables, replacing the PK with existing unique indices. I can review the other tables and open another PR removing the id column and updating the PK.

PR #2024, opened 2020-12-28
Issue #2026: Fixed the retry mechanism for subworkflows so that retry starts from the last failed task.

mehmet-islamoglu:
Fixed the retry mechanism for subworkflows so that retry starts from the last failed task.

Co-authors: @grkn @azb96


kishorebanala:
@mehmet-islamoglu We only need one PR against dev branch, which will eventually move to master. Closing this in favor of https://github.com/Netflix/conductor/pull/2025

Issue #2012, opened 2020-12-15
Retry Mechanism in SubWorkflow


mehmet-islamoglu:
We are using a **sub workflow** in our main workflow. When this sub workflow fails **retry mechanism** restarts it from the first task of the sub workflow. 

We want to utilize **retry mechanism** in a way so that we can **start immediately from the failed task of the subworkflow** like how it would happen in a retry mechanism in the **parent** workflow.

How can we achieve this? 

Thank you very much for your help.

@v1r3n 
@apanicker-nflx 
@kishorebanala 
@jfernandez

kishorebanala:
Copying over my response from Gitter here:

```
@mehmet-islamoglu You're right. However, you can retry the sub workflow instead, which would do what you're expecting.

We've considered doing the same for retry from parent workflow without breaking the current usage. I.e when retried from parent, with a flag to keep the existing behavior of restarting the subworkflow vs retrying it, which would recursively evaluate the subworkflow(s) until finding the last discrete task and retrying it (For eg., in a multi-level sub workflow).

It might be a while until we can prioritize this, but if you've bandwidth, please feel free to make a contribution starting with your proposal in a Github issue for the community to follow through.
```

mehmet-islamoglu:
Thank you for the feedback @kishorebanala 

I will go ahead and create a Pull Request with my colleagues @grkn @azb96

mehmet-islamoglu:
I created two PRs to dev and master for this issue. 

I can't re-open the issue and move the issue to backlog->TO-DO, so checks on my PRs are failing. Can you please help me with this problem?

@v1r3n
@apanicker-nflx
@kishorebanala
@jfernandez

PR #2025, opened 2020-12-28
Issue #2026: Fixed the retry mechanism for subworkflows so that retry starts from the last failed task.

mehmet-islamoglu:
Issue #2012: Fixed the retry mechanism for subworkflows so that retry starts from the last failed task.

Co-authors: @grkn @azb96

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
@@ -13,6 +13,7 @@
 package com.netflix.conductor.core.execution;
 
 import static com.netflix.conductor.common.metadata.tasks.Task.Status.CANCELED;
+import static com.netflix.conductor.common.metadata.tasks.Task.Status.TIMED_OUT;
 import static com.netflix.conductor.common.metadata.tasks.Task.Status.FAILED;
 import static com.netflix.conductor.common.metadata.tasks.Task.Status.FAILED_WITH_TERMINAL_ERROR;
 import static com.netflix.conductor.common.metadata.tasks.Task.Status.IN_PROGRESS;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -547,7 +548,8 @@ public void retry(String workflowId) {
         if (workflow.getTasks().isEmpty()) {
             throw new ApplicationException(CONFLICT, "Workflow has not started yet");
         }
-
+        Optional<Task> lTask = workflow.getTasks().stream().filter(this::findLastFailedOrTimeOutTask).findFirst();
+        workflow = findLastFailedSubWorkflow(lTask.get(),workflow);
```
kishorebanala, line 552: @mehmet-islamoglu Thanks for the contribution. As mentioned in the previous comment, this change should be backwards compatible, ```I.e when retried from 
parent, with a flag to keep the existing behavior of restarting the subworkflow vs retrying it, which would recursively evaluate the subworkflow(s) until finding the last discrete task and retrying it (For eg., in a multi-level sub workflow).```.

Subworkflows will no longer be restarted with this change, which might break few users. Please update the PR to accommodate this.

kishorebanala (changes requested):
Also, please add or update the integration tests in test-harness module.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
azb96: Hello @kishorebanala what is the reason behind keeping existing behavour if it is causing bugs?

mehmet-islamoglu:
@kishorebanala @apanicker-nflx @aravindanr  The new PR is #2038

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
kishorebanala: @azb96 Can you elaborate on the existing behavior causing bugs please? 

While it naturally makes sense to retry Subworkflow task on retrying the parent, it is possible that folks might've built their existing pipelines in such a way that the expectation is to recreate a new Subworkflow on retrying the parent. Hence the reason to keep this backwards compatible. We generally avoid removing the existing features in minor releases, and Conductor 3.0 effort might be a good place to re-consider this.

kishorebanala:
@mehmet-islamoglu I see the build failure is due to one of the failing tests in `test-harness` module. Can you please fix it and let us know when ready to go.

PR #2027, opened 2020-12-28
Provide Prometheus module for metrics

andrea11:
Hello,
The purpose of this PR is to create a module to collect the metrics produced via Spectator inside a Prometheus collector.
I tested on my Applications and it seems to work. Unfortunately, it seems that Spectator does not support Prometheus client natively, so I needed to use [dropwizard](https://www.robustperception.io/exposing-dropwizard-metrics-to-prometheus) (similarly to the logging metrics module)


andrea11:
Looking at this comment Netflix/spectator#782, I think that new Metrics should be used instead (in fact, exporting the metrics  as in this PR, make a lot of label/tags unavailable in Prometheus)

andrea11:
Ok, I finally succeed to expose metrics with labels/tags for Prometheus thanks to micrometer and spectator-reg-micrometer. I updated the PR accordingly.
Let me know if there is something to change.

apanicker-nflx (commented):
> Prometheus client natively, so I needed to use dropwizard (similarly to the logging metrics module)

I do not see the dropwizard metrics library being used however, was this missed?

Also, could you also make this contribution to the [3.0 branch](https://github.com/Netflix/conductor/tree/3.0) which uses SpringBoot. [Here's the LoggingMetricsConfiguration for reference](https://github.com/Netflix/conductor/blob/3.0/contribs/src/main/java/com/netflix/conductor/contribs/metrics/LoggingMetricsConfiguration.java).

andrea11:
Hi @apanicker-nflx,
Indeed what I said in my first comment is not anymore valid. I succeed to expose metrics for Prometheus via micrometer (supported by Spectator) that has a sort of adapter for formatting data in a friendly format supported by Prometheus. 

I tried before to use Dropwizard but the metrics exported were not compliant with Prometheus data.

Furthermore, I followed your suggestion, using Spring boot

mehmet-islamoglu:
How do you restart build in TravisCI? We are also trying to contribute to the project with my colleague @azb96 by merging PR #2038 @andrea11 ?

andrea11:
> How do you restart build in TravisCI? We are also trying to contribute to the project with my colleague @azb96 by merging PR #2038 @andrea11 ?

If you are not admin on the project, the only way is to close and re-open the PR. It is mostly a workaround, but it works

azb96:
@andrea11 Is there any way we can communicate with you privately about the contribution process for guidance?

andrea11:
@azb96 maybe we can discuss this publicly here. Otherwise, I invited you to a private chat on gitter

andrea11:
@apanicker-nflx could you please review this PR and eventually merge it? Thank you

apanicker-nflx (approved):
@andrea11 Thank you for making these changes. 
Please remove the dependency locks that are committed as part of this PR. We want to refrain from locking into dependency versions till the 3.0 branch is ready for release.

andrea11:
@apanicker-nflx done 👍

PR #2029, opened 2020-12-31
use springboot property conversion

apanicker-nflx:
- Use SpringBoot property conversion for `Duration` and `DataSize`
- Un-commented the `Audit` and `Trace` annotations


cassandra-persistence/src/main/java/com/netflix/conductor/cassandra/config/CassandraProperties.java
```diff
@@ -72,17 +75,20 @@
     /**
      * The time in seconds after which the in-memory task definitions cache will be refreshed
```
aravindanr, line 76: Comment includes the unit. "time in seconds"

cassandra-persistence/src/main/java/com/netflix/conductor/cassandra/config/CassandraProperties.java
```diff
@@ -72,17 +75,20 @@
...
      */
-    private int taskDefCacheRefreshTimeSecs = 60;
+    @DurationUnit(ChronoUnit.SECONDS)
```
aravindanr, line 78: Is the DurationUnit still needed? We could update the documentation to say that Duration format is supported.

contribs/src/main/java/com/netflix/conductor/contribs/queue/sqs/config/SQSEventQueueProperties.java
```diff
@@ -25,12 +28,13 @@
     /**
      * The polling interval (in milliseconds)
```
aravindanr, line 29: Same as above.

contribs/src/main/java/com/netflix/conductor/contribs/storage/config/S3Properties.java
```diff
@@ -25,7 +28,8 @@
     /**
      * The time (in seconds) for which the signed url will be valid
      */
-    private int signedUrlExpirationSeconds = 5;
+    @DurationUnit(ChronoUnit.SECONDS)
```
aravindanr, line 31: Same as above, `DurationUnit` may not be necessary and the comment needs update.

PR #2036, opened 2021-01-06
Client Starter for Spring

jeremy-l-ford:


client-spring/src/main/java/com/netflix/conductor/client/spring/ConductorClientAutoConfiguration.java
```diff
@@ -0,0 +1,62 @@
...
+
+@Configuration(proxyBeanMethods = false)
+@EnableConfigurationProperties(ClientProperties.class)
+public class ConductorClientAutoConfiguration {
+
+    @Autowired(required = false)
+    private EurekaClient eurekaClient;
+
+    @Autowired
+    private List<Worker> workers;
```
aravindanr, line 35: If there are no `Worker` instances, would the autowiring fail? Does this also need a `required=false` property?

jeremy-l-ford: Yes, if no Worker instances bound, this would need to be optional.  I'm not sure what the client can accomplish without a worker though.

aravindanr: Fair point.

aravindanr (approved):

client-spring/src/main/java/com/netflix/conductor/client/spring/ClientProperties.java
```diff
@@ -0,0 +1,83 @@
...
+ */
+package com.netflix.conductor.client.spring;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+import java.time.Duration;
+import java.util.HashMap;
+import java.util.Map;
+
+@ConfigurationProperties("com.netflix.conductor.client")
```
apanicker-nflx, line 21: Consider using a different prefix for the properties, maybe just ```conductor.client``` 
Using the package name as the property prefix leaks out implementation details.

client-spring/build.gradle
```diff
@@ -0,0 +1,10 @@
+
+dependencies {
+
+    implementation project(':conductor-common')
+    api project(':conductor-client')
+
+    implementation "com.netflix.eureka:eureka-client:${revEurekaClient}"
+    implementation 'org.springframework.boot:spring-boot-starter'
+    annotationProcessor 'org.springframework.boot:spring-boot-configuration-processor'
```
apanicker-nflx, line 9: Since this is already included at the root project level, this dependency is not needed in this module.

client-spring/src/test/java/com/netflix/conductor/client/spring/ExampleClient.java
```diff
@@ -0,0 +1,78 @@
...
+import java.io.IOException;
+import java.util.Properties;
+
+@SpringBootApplication
+public class ExampleClient {
+
+    private static final Logger log = LoggerFactory.getLogger(ExampleClient.class);
+
+    public static void main(String[] args) throws IOException {
+        loadExternalConfig();
```
aravindanr, line 35: This method is not needed. It's needed for the conductor server but not for the clients.

apanicker-nflx (changes requested):

apanicker-nflx (approved):

PR #2041, opened 2021-01-08
doc: Update apispec sort format from <direction>:<fieldname> to <fieldname>:<direction>

n-h:
### What
doc: Update apispec sort format from `<direction>:<fieldname>` to `<fieldname>:<direction>`.

### Why
The implementation uses the reversed format, than what is in the docs. Implementation is here:
https://github.com/Netflix/conductor/blob/be9e74ebd8ee784abb2b5d0eeb9d4c3786fe108f/jersey/src/main/java/com/netflix/conductor/server/resources/WorkflowResource.java#L214

Note: Swagger UI already matches the documentation, so no change needed there.
SwaggerUI

### DIFF
--- a/docs/docs/apispec.md
+++ b/docs/docs/apispec.md
@@ -66,7 +66,7 @@ Conductor uses Elasticsearch for indexing workflow execution and is used by sear
 |---|---|
 |start|Page number.  Defaults to 0|
 |size|Number of results to return|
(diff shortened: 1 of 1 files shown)


n-h:
Didn't open to `dev` branch. Closing. Will open new one to that branch.

PR #2060, opened 2021-01-22
fix npe

andyyumiao:
企业咚咚20210122212003
fix the npe when subworkflow task is null.


kishorebanala:
Marking this as closed; Please feel free to reopen if this is still relevant.

PR #2062, opened 2021-01-25
add task failed hook method for WorkflowStatusListener

andyyumiao:
We extended WorkflowStatusListener in our production environment and added onWorkflowFailed method for task fails. This feature was added because of it is not enough for terminate and complete status trigger and sometimes we need hook mechanism when one of the tasks fails when the workflow is not terminated. Pls evaluate this feature.


apanicker-nflx:
Hey @andyyumiao, while the purpose of this PR is to implement task-level notifications (IIUC), the implementation is incomplete and when bundled in together with the WorkflowStatusListener conflates the responsibilities of workflow-level notifications and violates the Interface Segregation principle. 
Task level notifications should be a different interface and should provide controls to tweak notification levels based on attempts, retries, timeouts etc

PR #2028, opened 2020-12-30
remove id column from tables, set primary key to existing unique index

mactaggart:
This PR removes id column/primary key from conductor MySQL/Postgres tables and replaces pk with unique index of table. MySQL/Postgres DAOs were updated to sort by "created_on" column instead of auto-incremented "id" column. Supersedes #2013 

@kishorebanala @raghavendra-bhat @rickfish

PR #2075, opened 2021-02-03
Microsoft SQL server persistence

TwoUnderscorez:
- sqlserver lock implementation for decider service
- sqlserver persistence implementation
  - conductor queues: shared: requires the decider service lock to be enabled. local_only: each conductor instance gets it's own queue (identified by the `LOCAL_RACK` property).

This is not complete yet, but I think it's ready for code review.
Thanks.


TwoUnderscorez:
@apanicker-nflx When do you think someone could have a look? I personally think it's almost done. The only things that are missing are integ test and potentially QueueDAO test on in memory tables and maybe some of the ExecutionDAO tables.

I don't think the in memory tables should be configurable, as it introduces a lot of complexities and is probably a niche feature anyway. So just testing general compatibility is fine and then letting the users define the in memory table on the sqlserver is probably a better approach.

apanicker-nflx:
@TwoUnderscorez Thank you very much for this contribution. We will begin reviewing this PR and post updates.

apanicker-nflx:
@mactaggart @rickfish @mashurex @s50600822  As users/contributors to other relational persistence layers (MySQL/Postgres) in Conductor, we would really appreciate your help with reviewing this pull request. Thank you for taking the time.

mashurex:
@apanicker-nflx I've done some cursory review at this point but my SQL Server is a bit rusty compared to MySQL and PostgreSQL.

@mactaggart @rickfish @s50600822 now that there are going to be 3 different RDBMS persistence layers, do you think there would be an advantage to unifying them to some degree? I have lots of thoughts about the schema and some issues I've noticed in the MySQL layer over time. @apanicker-nflx should I create an issue for the thought? Pull request?

mashurex (approved):
The DDL for the schema appears to be the SQL server equivalent of MySQL and all of the queries I reviewed seemed so as well.

rickfish:
@mashurex @apanicker-nflx I don't have any recent SQL Server experience so I wouldn't be a good resource to review this. As far as combining all sql-based database persistence modules, that is what we did before the Postgres module was introduced. We created a sql-persistence module. There were actually very few differences between MySQL and Postgres which we accommodated by having an interface for all query strings with default methods for the ones that were the same and the individual implementations would override what they needed to. This worked pretty well.

The problem that we faced with doing it that way is that our knowledge of MySQL started to dwindle and whenever we would make a change for Postgres we would also have to be on the hook for making the MySQL changes. Since we no longer had a MySQL DBMS inhouse we couldn't test the changes for that db. When the Postgres persistence module was introduced we decided to bag our implementation in favor of that one because of the logistics of contributing back.

sqlserver-persistence/src/main/java/com/netflix/conductor/dao/sqlserver/Query.java
```diff
@@ -0,0 +1,635 @@
+package com.netflix.conductor.dao.sqlserver;
+
+import com.microsoft.sqlserver.jdbc.ISQLServerPreparedStatement;
```
mactaggart, line 3: unused import

sqlserver-persistence/src/main/resources/db/migration_sqlserver/V1__initial_schema.sql
```diff
@@ -0,0 +1,275 @@
...
+-- --------------------------------------------------------------------------------------------------------------
+GO
+
+-- --------------------------------------------------------------------------------------------------------------
+-- TAGLES FOR METADATA DAO
+-- --------------------------------------------------------------------------------------------------------------
+
+-- meta_event_handler
+CREATE TABLE [dbo].[meta_event_handler] (
+    id INT PRIMARY KEY IDENTITY(1, 1),
```
mactaggart, line 13: Recently (#2028) we removed the "id" column in all conductor tables (mysql/postges) since the id column is not used. I suggest we do the same here for consistency.

sqlserver-persistence/src/test/java/com/netflix/conductor/dao/sqlserver/SqlServerDAOTestUtil.java
```diff
@@ -0,0 +1,183 @@
...
+import org.flywaydb.core.Flyway;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.sql.DataSource;
+
+import java.nio.file.Paths;
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
```
mactaggart, line 17: unused import

sqlserver-persistence/src/test/java/com/netflix/conductor/dao/sqlserver/SqlServerLockTest.java
```diff
@@ -0,0 +1,166 @@
...
+        SqlServerLock lck = new SqlServerLock(null, null, cnf);
+        assertEquals("asdf.qwerty", lck.getLockId("qwerty"));
+    }
+
+    @Test(expected = UnsupportedOperationException.class)
+    public void testDangerousApi() {
+        lock.acquireLock("lockId");
+    }
+
+    // Can't get static mocks to work
```
mactaggart, line 131: is this a WIP?

sqlserver-persistence/src/test/java/com/netflix/conductor/dao/sqlserver/SqlServerQueueDAOTest.java
```diff
@@ -0,0 +1,382 @@
+package com.netflix.conductor.dao.sqlserver;
+
+import com.google.common.collect.ImmutableList;
+import com.jayway.jsonpath.Configuration;
```
mactaggart, line 4: some unused imports here

test-harness/docker-compose.yml
```diff
@@ -20,4 +20,24 @@ services:
...
-      POSTGRES_PASSWORD: postgres
\ No newline at end of file
+      POSTGRES_PASSWORD: postgres
+
+  sqlserver:
+    image: conductor:sqlserver
+    build: 
+      context: ../sqlserver-persistence/docker
+      dockerfile: sqlserver/Dockerfile
+    restart: always
```
mactaggart, line 30: restart defined twice

test-harness/src/test/java/com/netflix/conductor/tests/utils/SqlServerTestModule.java
```diff
@@ -0,0 +1,98 @@
+/*
```
mactaggart, line 1: This module doesn't appear to be used. In other modules it's used in integration tests (end to end).

test-harness/src/test/java/com/netflix/conductor/tests/utils/SqlServerTestRunner.java
```diff
@@ -0,0 +1,71 @@
+/**
```
mactaggart, line 1: This file doesn't appear to be used. In other modules it's used in integration tests (end to end). Perhaps we add an end to end integration test?

mactaggart (changes requested):
Overall looks good. Some changes suggested.

TwoUnderscorez:
@mactaggart Thanks for taking the time to review my code. We are currently working on improving the performance of the queries / indices / tables and right after that I'll try to resolve all of your comments.

TwoUnderscorez:
@apanicker-nflx @mactaggart @rickfish @mashurex    what are your opinions on using stored procedures? We are currently running about 50,000 WFs per day and the performance is either like MySql or little bit better at times which is pretty good. We will be running about 10 times as much in the future though. I asked our DBA what he thinks and he basically said that we should use stored procedures. The SQL Server does undesirable things with the queue_message table which sometimes cause the pop to take longer than 1 second. And these things can only be adjusted with stored procedures. 

Also, do you prefer I merge to version 2 or 3? There are a lot of changes in 3 so that might take longer to figure out. 

Thanks

rickfish:
The only problems I have with using stored procedures is (1) debugging issues if they happen and (2) potential problems where different organizations have different versions of the same db which could change the available functions, syntax, etc. They definitely save time on situations where there are multiple round-trips to the db server on one logical api. There are enough of those to make a difference, like adding/updating/removing task/workflow, getting workflows by type, and others so it could improve performance. 

That said, it is a large effort in development time and testing time to convert everything to stored procs, at least for the mysql and postgres modules.

As far as pop() goes, the slow response time could also be related to your indexes and the plan that is being executed by the dbms. I know that changing the postgres combo_queue_message index from using ascending priority to descending priority made a significant performance improvement to the pop() for us. As the queue_message table grows, that makes a huge difference.  We had at one point 12 million rows in that table and pop() was very slow until we made that index change. We also got to the bottom of why there were 12 million rows and that helped when we were able to get it under a million. You might want to look at the indexes for your sql server implementation.

I am not involved in v3 so I have no opinion on v2 vs v3 yet but if I was just starting a new module I would probably start on v3 just so I wouldn't have to do the implementation twice.

aravindanr:
>Up until v3.0.2 it seemed like conductor.db.type was just ignored and the in memory redis/dynomite was used, and on v3.0.3 I'm getting some kind of a Bean exception when trying to use MySQL.

@TwoUnderscorez This is addressed in the latest release [v3.0.4](https://github.com/Netflix/conductor/releases/tag/v3.0.4). Could you please try again? Since 3.0.x is the version with long-term support, we recommend that SQL server persistence is added to the `main` with `2.31` as the next choice.

sqlserver-persistence/src/main/java/com/netflix/conductor/dao/sqlserver/Query.java
TwoUnderscorez: Removed all unused imports
https://github.com/TwoUnderscorez/conductor/commit/913256ee1058ce863147e676eac0eac7a2cee105

sqlserver-persistence/src/main/java/com/netflix/conductor/dao/sqlserver/SqlServerQueueDAO.java
```diff
@@ -0,0 +1,323 @@
+package com.netflix.conductor.dao.sqlserver;
+
+import com.amazonaws.services.kms.model.InvalidArnException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Maps;
+import com.google.common.util.concurrent.Uninterruptibles;
+import com.netflix.conductor.core.events.queue.Message;
+import com.netflix.conductor.core.execution.ApplicationException;
```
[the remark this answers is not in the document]
TwoUnderscorez: Removed all unused imports
https://github.com/TwoUnderscorez/conductor/commit/913256ee1058ce863147e676eac0eac7a2cee105

sqlserver-persistence/src/main/java/com/netflix/conductor/dao/sqlserver/SqlServerBaseDAO.java
```diff
@@ -0,0 +1,264 @@
...
+
+import static com.netflix.conductor.core.execution.ApplicationException.Code.BACKEND_ERROR;
+import static com.netflix.conductor.core.execution.ApplicationException.Code.CONFLICT;
+import static com.netflix.conductor.core.execution.ApplicationException.Code.INTERNAL_ERROR;
+import static java.lang.Integer.parseInt;
+import static java.lang.System.getProperty;
+
+public abstract class SqlServerBaseDAO {
+    // TODO: Move to configuration
+    private static final String MAX_RETRY_ON_DEADLOCK_PROPERTY_NAME = "conductor.mysql.deadlock.retry.max";
```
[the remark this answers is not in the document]
TwoUnderscorez: Fixed in https://github.com/TwoUnderscorez/conductor/commit/44bcafaac894648992c2e16e754a252d35d8e79b

sqlserver-persistence/src/main/java/com/netflix/conductor/sqlserver/SqlServerDataSourceProvider.java
```diff
@@ -0,0 +1,89 @@
+package com.netflix.conductor.sqlserver;
```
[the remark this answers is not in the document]
TwoUnderscorez: Removed all unused imports
https://github.com/TwoUnderscorez/conductor/commit/913256ee1058ce863147e676eac0eac7a2cee105

sqlserver-persistence/build.gradle
```diff
@@ -0,0 +1,57 @@
...
+}
+
+test.doFirst {
+    dockerCompose.exposeAsSystemProperties(test)
+}
+
+test.finalizedBy("intDockerDown")
+
+test {
+    //the MySQL unit tests must run within the same JVM to share the same embedded DB
```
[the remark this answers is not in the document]
TwoUnderscorez: Fixed in https://github.com/TwoUnderscorez/conductor/commit/44bcafaac894648992c2e16e754a252d35d8e79b

sqlserver-persistence/src/main/resources/db/migration_sqlserver/V1__initial_schema.sql
```diff
@@ -0,0 +1,275 @@
+-- --------------------------------------------------------------------------------------------------------------
+-- Migration script V1 for SQLSERVER
+-- Table and index names are taken from MySQL
+-- --------------------------------------------------------------------------------------------------------------
+GO
+
+-- --------------------------------------------------------------------------------------------------------------
+-- TAGLES FOR METADATA DAO
```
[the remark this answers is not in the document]
TwoUnderscorez: Fixed in https://github.com/TwoUnderscorez/conductor/commit/44bcafaac894648992c2e16e754a252d35d8e79b

sqlserver-persistence/src/main/resources/db/migration_sqlserver/V1__initial_schema.sql
TwoUnderscorez: I removed `id` from all the tables except `queue_message` because I need the `id`  column there.
https://github.com/TwoUnderscorez/conductor/commit/44bcafaac894648992c2e16e754a252d35d8e79b

test-harness/docker-compose.yml
TwoUnderscorez: Fixed in https://github.com/TwoUnderscorez/conductor/commit/44bcafaac894648992c2e16e754a252d35d8e79b

test-harness/src/test/java/com/netflix/conductor/tests/utils/SqlServerTestModule.java
TwoUnderscorez: There are now integration tests for sqlserver
https://github.com/TwoUnderscorez/conductor/commit/44bcafaac894648992c2e16e754a252d35d8e79b

test-harness/src/test/java/com/netflix/conductor/tests/utils/SqlServerTestRunner.java
TwoUnderscorez: There are now integration tests for sqlserver
https://github.com/TwoUnderscorez/conductor/commit/44bcafaac894648992c2e16e754a252d35d8e79b

TwoUnderscorez:
@apanicker-nflx @mactaggart @rickfish @mashurex  @aravindanr
I have made a few more changes, can you please have a look?
As for conductor 3, I'll open a seprate pull request soon.

rickfish:
I can only validate for Postgres. Since there don't seem to be any changes to that module it is ok with me.

TwoUnderscorez:
> I can only validate for Postgres. Since there don't seem to be any changes to that module it is ok with me.

@rickfish Acctually there are, this is to make the integration tests run when running from windows. I'm not sure this is the best way to fix it, by the fact is that when `Path.get` was used, flyway was unable to find the migrations folder (for both sqlserver and postgres).
https://github.com/Netflix/conductor/blob/89fbe86005e7a1d456e20b5884e9862fb61b130c/postgres-persistence/src/main/java/com/netflix/conductor/postgres/PostgresDataSourceProvider.java#L89

rickfish:
@TwoUnderscorez that change works fine for me in Postgres.

PR #2103, opened 2021-02-26
send notification on workflow finalization

apanicker-nflx:
Workflow completion notification will be sent immediately after a workflow moves to COMPLETED state.
Workflow termination notification will be sent immediately after a workflow moves to TERMINATED/FAILED/TIMED_OUT state.

Workflow finalization notification will be sent after all the tasks within the workflow have been moved to an expected terminal state, the parent workflow (if any) has been updated, the failure workflow (if any) has been triggered and the workflow is cleaned up from the decider queue to stop any further evaluations.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/listener/WorkflowStatusListener.java
+++ b/core/src/main/java/com/netflix/conductor/core/listener/WorkflowStatusListener.java
@@ -19,7 +19,29 @@ import com.netflix.conductor.common.run.Workflow;
  */
 public interface WorkflowStatusListener {
 
+    default void onWorkflowCompletedIfEnabled(Workflow workflow) {
+        if (workflow.getWorkflowDefinition().isWorkflowStatusListenerEnabled()) {
+            onWorkflowCompleted(workflow);
+        }
+    }
+
+    default void onWorkflowTerminatedIfEnabled(Workflow workflow) {
+        if (workflow.getWorkflowDefinition().isWorkflowStatusListenerEnabled()) {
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/netflix/conductor/core/listener/WorkflowStatusListener.java
```diff
@@ -22,4 +22,8 @@
     void onWorkflowCompleted(Workflow workflow);
 
     void onWorkflowTerminated(Workflow workflow);
+
+    default void onWorkflowFinalized(Workflow workflow) {
```
aravindanr, line 26: General comment for all methods on this interface. The methods here are not called if `isWorkflowStatusListenerEnabled()` returns `false`. In every place where these need to be called, a conditional `if` block precedes. I suggest that we bring that check into the contract of this interface, like this,

```java
default void onWorkflowCompletedIfEnabled(Workflow workflow) {        if(workflow.getWorkflowDefinition().isWorkflowStatusListenerEnabled()) {
            onWorkflowCompleted(workflow);
        }
    }
```
A method like this can be added for all the listeners. It would make the code more readable.

aravindanr (approved):

core/src/main/java/com/netflix/conductor/core/listener/WorkflowStatusListener.java
apanicker-nflx: good idea, updated the interface

PR #2112, opened 2021-03-03
Fix conditional sync system task with large payload

apanicker-nflx:
- Fixed case where task input data is not populated from external storage when sync system task is executed
- Fixed case where terminate task output from external storage is not applied to workflow
- Fixed case where lastRetriedTime of a workflow is not reset when a workflow is restarted

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
@@ -40,22 +40,11 @@ import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Service;
 
 import java.time.Duration;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.LinkedHashMap;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
```diff
@@ -315,33 +304,36 @@ void updateWorkflowOutput(final Workflow workflow, Task task) {
...
-        if (terminateTask.isPresent()) {
-            if (!terminateTask.get().getOutputData().isEmpty()) {
-                workflow.setOutput(terminateTask.get().getOutputData());
+        Optional<Task> optionalTask = allTasks.stream()
+                .filter(t -> TaskType.TERMINATE.name().equals(t.getTaskType()) && t.getStatus().isTerminal()
+                        && t.getStatus().isSuccessful())
+                .findFirst();
+        if (optionalTask.isPresent()) {
+            Task terminateTask = optionalTask.get();
+            if (StringUtils.isNotBlank(terminateTask.getExternalOutputPayloadStoragePath())) {
```
aravindanr, line 313: Should this be applied to all sync system tasks?

aravindanr (approved):

core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
apanicker-nflx: other sync system tasks are handled in the else block below, here we need special handling of the terminate task because terminate task can be used to override workflow output.

PR #2115, opened 2021-03-05
Disable background processes when instance is disabled

apanicker-nflx:


contribs/src/main/java/com/netflix/conductor/contribs/queue/amqp/AMQPObservableQueue.java
```diff
@@ -63,6 +65,7 @@
     private Channel channel;
     private final Address[] addresses;
     protected LinkedBlockingQueue<Message> messages = new LinkedBlockingQueue<>();
+    private final AtomicBoolean running = new AtomicBoolean();
```
aravindanr, line 68: The `AtomicBoolean` seems unnecessary here, since there is no "compare and set" operation. For this scenario, a `volatile boolean` should suffice.

aravindanr (commented):
Instead of repeating the logic in all classes that need to be stopped, it can be contained a base class like this,

```java
package com.netflix.conductor.lifecycle;

import org.springframework.context.SmartLifecycle;

public class LifecycleAwareWorker implements SmartLifecycle {

    private volatile boolean running = false;

    @Override
    public void start() {
        running = true;
    }

    @Override
    public void stop() {
        running = false;
    }

    @Override
    public boolean isRunning() {
        return running;
    }
}

```

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -1695,6 +1695,8 @@ private boolean rerunWF(String workflowId, String taskId, Map<String, Object> ta
             rerunFromTask.setUpdateTime(0);
             rerunFromTask.setEndTime(0);
             rerunFromTask.setOutputData(null);
+            rerunFromTask.setRetried(false);
```
aravindanr, line 1698: Is this change related to disabling background workers?

core/src/main/java/com/netflix/conductor/core/execution/WorkflowSweeper.java
```diff
@@ -48,36 +52,31 @@ public WorkflowSweeper(WorkflowExecutor workflowExecutor, WorkflowRepairService
...
     public void init(WorkflowExecutor workflowExecutor, WorkflowRepairService workflowRepairService) {
         ScheduledExecutorService deciderPool = Executors.newScheduledThreadPool(1);
         deciderPool.scheduleWithFixedDelay(() -> {
-            try {
-                boolean disable = properties.isSweepDisabled();
-                if (disable) {
-                    LOGGER.info("Workflow sweep is disabled.");
-                    return;
+            if (!running.get()) {
+                LOGGER.debug("Instance disabled, skip workflow sweep");
```
aravindanr, line 65: The comment may not be appropriate. There could be other scenarios where Spring calls `stop()` on the `Lifecycle`.  A generic comment would be better here.

core/src/main/java/com/netflix/conductor/core/execution/tasks/SystemTaskWorkerCoordinator.java
```diff
@@ -75,15 +80,12 @@ public SystemTaskWorkerCoordinator(QueueDAO queueDAO, WorkflowExecutor workflowE
...
-        int threadCount = properties.getSystemTaskWorkerThreadCount();
-        if (threadCount > 0) {
-            this.workflowSystemTasks.forEach(this::add);
-            this.systemTaskExecutor = new SystemTaskExecutor(queueDAO, workflowExecutor, properties, executionService);
-            new Thread(this::listen).start();
-            LOGGER.info("System Task Worker Coordinator initialized with poll interval: {}", pollInterval);
-        } else {
-            LOGGER.info("System Task Worker DISABLED");
-        }
+        int threadCount =
```
aravindanr, line 83: Instead of this, the default value of 1 could be set in `ConductorProperties`.

contribs/src/main/java/com/netflix/conductor/contribs/queue/amqp/AMQPObservableQueue.java
apanicker-nflx: good idea, changed

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
apanicker-nflx: no, this is just some required cleanup.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowSweeper.java
apanicker-nflx: changed

core/src/main/java/com/netflix/conductor/core/execution/tasks/SystemTaskWorkerCoordinator.java
apanicker-nflx: The default value is 10, this is to prevent the value being overridden to 0.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowSweeper.java
```diff
@@ -48,36 +50,31 @@ public WorkflowSweeper(WorkflowExecutor workflowExecutor, WorkflowRepairService
...
         this.queueDAO = queueDAO;
-        this.executorThreadPoolSize = properties.getSweeperThreadCount();
-        if (this.executorThreadPoolSize > 0) {
-            this.executorService = Executors.newFixedThreadPool(executorThreadPoolSize);
-            init(workflowExecutor, workflowRepairService);
-            LOGGER.info("Workflow Sweeper Initialized");
-        } else {
-            LOGGER.warn("Workflow sweeper is DISABLED");
-        }
+        this.executorThreadPoolSize = properties.getSweeperThreadCount() > 0 ? properties.getSweeperThreadCount() : 1;
```
aravindanr, line 53: The value `1` can be set as a default in `ConductorProperties`.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowSweeper.java
```diff
@@ -12,31 +12,33 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.stereotype.Component;
 
 @SuppressWarnings("SpringJavaInjectionPointsAutowiringInspection")
-@Service
-public class WorkflowSweeper {
+@Component
+@ConditionalOnProperty(name = "conductor.workflow-sweeper.enabled", havingValue = "true", matchIfMissing = true)
```
aravindanr, line 36: +1 👏  on renaming the property.

core/src/main/java/com/netflix/conductor/core/execution/tasks/SystemTaskWorkerCoordinator.java
aravindanr: If the value is set to `0`, shouldn't we use the default of `10`? The property's value determination gets complicated,

1. If the value is set to `0`, it becomes `1`.
2. If no value is set, then it's `10` which is the default.
3. If the value set is > 0, then the set value is used.

I vote to throw an error when the value is set to `0`. The error message could inform the user that to disable workers, the flag can be set as `false`.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowSweeper.java
aravindanr, line 53: I think an error should be thrown here if the thread count is set to `0`.

core/src/main/java/com/netflix/conductor/core/execution/tasks/SystemTaskWorkerCoordinator.java
aravindanr: On a related note, it might be worth considering the default values a function of the available processors. Something like this,

```java
private int systemTaskWorkerThreadCount = Runtime.getRuntime().availableProcessors() * 2;
```

PR #2137, opened 2021-03-22
add support for escaped expressions

rafaelfbs:
Hi, my first contribution here 👋🏻. We're evaluating Netflix conductor as a workflow executor for our custom automation. In our tests, we have a service that receives a template string like conductor expressions ```${variableName}```. So, we had to escape the expression to send it to our service. I think it's a simple and useful feature to start contributing here 🤩.

To escape an expression, prefix it with an extra _$_ character. So, as an example, the expression will be  ```$${variableName}```.

### DIFF
--- a/core/src/test/java/com/netflix/conductor/core/execution/TestParametersUtils.java
+++ b/core/src/test/java/com/netflix/conductor/core/execution/TestParametersUtils.java
@@ -215,12 +215,14 @@ public class TestParametersUtils {
 
 		Map<String, String> mapValue = new HashMap<>();
 		mapValue.put("a", "${someString}");
-		mapValue.put("b", "$${someNumber}");
+		mapValue.put("b", "${someNumber}");
+		mapValue.put("c", "$${someString} ${someNumber}");
 		input.put("map", mapValue);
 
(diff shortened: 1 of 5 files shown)


core/src/test/java/com/netflix/conductor/core/execution/TestParametersUtils.java
```diff
@@ -200,4 +199,51 @@ public void testReplaceInputWithMapAndList() throws Exception {
...
+		Map<String, Object> replaced = parametersUtils.replace(input, jsonObj);
+		assertNotNull(replaced);
+
+		// Verify that values are replaced correctly.
+		assertEquals("conductor ${$.someNumber}2 2${$.someString}", replaced.get("k1"));
+		assertEquals("${$.someString}afterText", replaced.get("k2"));
+		assertEquals("beforeText${$.someString}", replaced.get("k3"));
+		assertEquals("${$.someString} afterText", replaced.get("k4"));
+		assertEquals("beforeText ${$.someString}", replaced.get("k5"));
+
```
apanicker-nflx, line 237: Please also add the verification the replaced values for "map" and "list".

PR #2148, opened 2021-03-30
Bug/#2124 Make retry and max time to wait when connection exhausts  configurable

guru1306:
Hi All

I have made the retries to acquire the connection and max time to wait when connection exhausted as configurable. I have followed the same approach as max connections per host. The default values are the same as the hardcoded in the dyno client library.

@kishorebanala   Can you please review this?

Thanks
Guru


guru1306:
@apanicker-nflx @aravindanr  Could you take a look at this?

kishorebanala (approved):

apanicker-nflx (approved):

PR #2158, opened 2021-04-02
remove deprecated fields and methods

aravindanr:


common/src/main/java/com/netflix/conductor/common/run/Workflow.java
```diff
@@ -444,9 +383,8 @@ public String getWorkflowName() {
      * @return the workflow definition version.
      */
     public int getWorkflowVersion() {
-        return getWorkflowDefinition() != null ?
-            getWorkflowDefinition().getVersion() :
-            version;
+        Preconditions.checkNotNull(workflowDefinition, "Workflow definition is null");
+        return getWorkflowDefinition().getVersion();
```
apanicker-nflx, line 387: is this indirection needed? Similar to the method above this could be `workflowDefinition.getVersion()`

common/src/main/java/com/netflix/conductor/common/run/Workflow.java
```diff
@@ -85,14 +86,6 @@ public boolean isSuccessful() {
     @ProtoField(id = 9)
     private Map<String, Object> output = new HashMap<>();
 
-    @ProtoField(id = 10)
```
apanicker-nflx, line 88: Should we leave a note here to explicitly dissuade the use of these proto ids?

grpc/src/main/proto/model/workflow.proto
```diff
@@ -26,12 +26,9 @@ message Workflow {
     repeated Task tasks = 6;
     map<string, google.protobuf.Value> input = 8;
     map<string, google.protobuf.Value> output = 9;
-    string workflow_type = 10;
```
apanicker-nflx, line 29: Consider making a note here to prevent the use of these ids.

apanicker-nflx (approved):

grpc/src/main/proto/model/workflow.proto
aravindanr: Done.

PR #2176, opened 2021-04-14
support bulk termination of workflows

noamlevy81:
Pull Request type
----

- [x] Feature

Changes in this PR
----
added functionality supported by the API to the workflow client.


client/build.gradle
```diff
@@ -18,7 +18,7 @@ configurations.all {
 
 dependencies {
     implementation project(':conductor-common')
-
+    implementation project(':conductor-core')
```
apanicker-nflx, line 21: This dependency seems unnecessary and would bring in a large number of libraries into any application that would use the conductor client.

client/src/main/java/com/netflix/conductor/client/http/ClientBase.java
```diff
@@ -24,6 +24,7 @@
 import com.netflix.conductor.common.run.ExternalStorageLocation;
 import com.netflix.conductor.common.utils.ExternalPayloadStorage;
 import com.netflix.conductor.common.validation.ErrorResponse;
+import com.netflix.conductor.core.utils.BulkResponse;
```
apanicker-nflx, line 27: A better approach here would be to move the `BulkResponse` object to the `common` module.

apanicker-nflx (changes requested):

apanicker-nflx (approved):

aravindanr (approved):

PR #2181, opened 2021-04-16
Ensure async complete system tasks are not response timed out

apanicker-nflx:
Pull Request type
----
- [x] Feature

Changes in this PR
----
This change ensures that system tasks that are configured to be async complete are not response timed out. Such tasks are not expected to be monitored for completion by Conductor and hence, do not need to evaluated for response timeouts.

Re-enable ignored tests since the error may have been addressed.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
@@ -41,22 +41,11 @@ import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Service;
 
 import java.time.Duration;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.LinkedHashMap;
-import java.util.LinkedList;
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
```diff
@@ -676,7 +667,8 @@ boolean isResponseTimedOut(TaskDef taskDefinition, Task task) {
             LOGGER.warn("missing task type : {}, workflowId= {}", task.getTaskDefName(), task.getWorkflowInstanceId());
             return false;
         }
-        if (task.getStatus().isTerminal() || task.getTaskType().equals(SUB_WORKFLOW.name())) {
+
+        if (task.getStatus().isTerminal() || isAyncCompleteSystemTask(task)) {
```
kishorebanala, line 671: +1

kishorebanala (approved):

PR #2206, opened 2021-04-25
Docker & Docker Compose fixes - 3.X

maheshyaddanapudi:
Pull Request type
----

- [x] Bugfix
- [x] Feature

Changes in this PR
----

Issue # 1. Docker Builds are failing due to

1. Incorrect JDK & JRE versions - jdk8 & jre8. Whereas jdk11 and jre11 have to be used for conductor 3.x
2. Incorrect startup jar name in startup script is causing "No such file or directory" issue. The conductor-server-*-boot.jar has to be used for conductor 3.x instead of conductor-server-*-all.jar

Issue # 2. Docker Compose configuration missing out indexing property.

1. Docker Compose though specifies an elasticsearch container, misses out configuring the indexing property in config*.properties excluding config-local.properties.

All the above mentioned are fixed and tested.

Feature # Added Healtcheck to Dockerfile for conductor

Alternatives considered
----

No alternative possible as these are required changes for 2.x to 3.x migration.

### DIFF
--- a/server/src/main/java/com/netflix/conductor/Conductor.java
+++ b/server/src/main/java/com/netflix/conductor/Conductor.java
@@ -16,20 +16,23 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.autoconfigure.cassandra.CassandraAutoConfiguration;
+import org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration;
 import org.springframework.core.io.FileSystemResource;
 import org.springframework.util.StringUtils;
 
 import java.io.IOException;
 import java.util.Properties;
 
-@SpringBootApplication
+@SpringBootApplication(exclude = {CassandraAutoConfiguration.class, DataSourceAutoConfiguration.class})
 public class Conductor {
 
     private static final Logger log = LoggerFactory.getLogger(Conductor.class);
 
     public static void main(String[] args) throws IOException {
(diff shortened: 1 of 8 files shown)


arjunbiyani (approved):
Changing to openjdk:13-jdk worked

server/src/main/java/com/netflix/conductor/Conductor.java
```diff
@@ -16,20 +16,23 @@
...
+import org.springframework.boot.autoconfigure.cassandra.CassandraAutoConfiguration;
+import org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration;
 import org.springframework.core.io.FileSystemResource;
 import org.springframework.util.StringUtils;
 
 import java.io.IOException;
 import java.util.Properties;
 
-@SpringBootApplication
+@SpringBootApplication(exclude = {CassandraAutoConfiguration.class, DataSourceAutoConfiguration.class})
```
aravindanr, line 27: Can you explain the reason for excluding these configurations?

server/src/main/java/com/netflix/conductor/Conductor.java
```diff
@@ -56,4 +59,23 @@ private static void loadExternalConfig() throws IOException {
             }
         }
     }
+    
+    private static void loadExternalConfigFromArgs(String[] args) throws IOException {
```
aravindanr, line 63: Is it possible to provide the configs as a file and leverage `CONDUCTOR_CONFIG_FILE`? Two ways to accomplish the same outcome usually adds confusion, we also have to consider the possibility of `CONDUCTOR_CONFIG_FILE` and command line args present at the same time.

Issue #2174, opened 2021-04-14
/api/workflow/{workflowId} endpoint not working in swagger


komatiraju032:
**Describe the bug**
Get  the workflow by id(/api/workflow/{workflowId}) endpoint is not working 

**To Reproduce**
1. Go to 'http://<conductor_ip>:8080/swagger-ui/index.html?configUrl=/api-docs/swagger-config'
2. Click on '/api/workflow/{workflowId}' endpoint. It's not working. When I click on it page is stuck (not responding)

**Expected behavior**
It should give the response with workflow details of given workflow id.

**Screenshots**

![Screenshot (5)](https://user-images.githubusercontent.com/22415463/114668334-3cd36a00-9d1e-11eb-82ec-c7aad1e8bef3.png)

**Additional context**

apanicker-nflx:
@komatiraju032 What type of persistence layer are you using? Do you see any errors in the server logs?

komatiraju032:
@apanicker-nflx I am using default configuration(docker-compose.yml). And For your information endpoint is working from postman.

maheshyaddanapudi:
No errors observed but the issue happens only in Swagger UI. I have opened a discussion thread https://github.com/Netflix/conductor/discussions/2192

komatiraju032:
@apanicker-nflx  I have find it's not working because of this issue https://github.com/OAI/OpenAPI-Specification/issues/822

maheshyaddanapudi:
@komatiraju032 is there any specification for a work around it different way to configure ? I couldn't directly find any in their Read Me. I'll explore too.

komatiraju032:
@maheshyaddanapudi  I found one more related issue on stackoverflow https://stackoverflow.com/questions/35605427/swagger-ui-freezes-after-api-fetch-and-browser-crashes

aravindanr:
@komatiraju032 we believe its because of the `com.google.protobuf.Any` type that the `inputMessage` and `outputMessage` fields in `Task` class. Excluding those fields in Swagger definition should solve the issue. We will get to it  asap.

komatiraju032:
@aravindanr I have fixed the issue with the information you have provided. I also created a PR #2210

aravindanr:
Thanks @komatiraju032 . I left a minor comment in the PR.

PR #2210, opened 2021-04-27
fixed the issue some endpoints not working in swagger-ui

komatiraju032:
Pull Request type
----

- [x] Bugfix

Changes in this PR
----

- Added the `springdoc-openapi-ui` dependency in "conductor/common/build.gradle".
- Updated the following entity classes(added the `@Hidden` annotation to `inputMessage` and `outputMessage` fields )
    1. EventHandler
    2. Task

Issue #2174

### DIFF
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -27,6 +27,8 @@ dependencies {
     implementation "com.fasterxml.jackson.core:jackson-databind"
     implementation "com.fasterxml.jackson.core:jackson-core"
 
+    implementation "org.springdoc:springdoc-openapi-ui:${revOpenapi}"
+
     testImplementation 'org.springframework.boot:spring-boot-starter-validation'
 }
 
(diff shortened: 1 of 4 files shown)


common/build.gradle
```diff
@@ -27,6 +27,8 @@ dependencies {
     implementation "com.fasterxml.jackson.core:jackson-databind"
     implementation "com.fasterxml.jackson.core:jackson-core"
 
+    implementation "org.springdoc:springdoc-openapi-ui:${revOpenapi}"
```
aravindanr, line 30: Can you change this to `compileOnly` dependency and add it as `implementation` in the `server` module? As it stands, projects that use `conductor-common` as a dependency will get the `springdoc-openapi-ui` needlessly.

komatiraju032: @aravindanr I made the changes

apanicker-nflx (approved):

aravindanr (approved):

PR #2222, opened 2021-05-02
add postgres missing dependency

noamlevy81:
Pull Request type
----

- [x] Bugfix

Changes in this PR
----
Added missing dependency in the server for Postgres

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -10,6 +10,7 @@ dependencies {
     implementation project(':conductor-contribs')
     implementation project(':conductor-redis-persistence')
     implementation project(':conductor-cassandra-persistence')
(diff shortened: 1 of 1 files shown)


server/build.gradle
```diff
@@ -10,6 +10,7 @@ dependencies {
     implementation project(':conductor-contribs')
     implementation project(':conductor-redis-persistence')
     implementation project(':conductor-cassandra-persistence')
+    implementation project(':conductor-postgres-persistence')
```
aravindanr, line 13: Can you also add `conductor-mysql-persistence`? That was also left out during the migration.

aravindanr (approved):

PR #2226, opened 2021-05-04
exclude FlywayAutoConfiguration

michaelpaliy:
Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----
Disable spring flyway. 

currently, the Postgres persistence layer throws exceptions on startup. by default when spring is used flyway try to run the migration, but it's not configured. also currently the flyway migration is configured in https://github.com/Netflix/conductor/blob/main/postgres-persistence/src/main/java/com/netflix/conductor/postgres/config/PostgresDataSourceProvider.java#L76.

`23318 [main] ERROR org.springframework.boot.SpringApplication [] - Application run failed
org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'flyway' defined in class path resource [org/springframework/boot/autoconfigure/flyway/FlywayAutoConfiguration$FlywayConfiguration.class]: Unsatisfied dependency expressed through method 'flyway' parameter 1; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'spring.datasource-org.springframework.boot.autoconfigure.jdbc.DataSourceProperties': Instantiation of bean failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.springframework.boot.autoconfigure.jdbc.DataSourceProperties]: Constructor threw exception; nested exception is java.lang.NoClassDefFoundError: org/springframework/jdbc/datasource/embedded/EmbeddedDatabaseType`


aravindanr (commented):
@michaelpaliy Thanks for taking the time to fix this issue.  Netflix does not use Postgres or Mysql internally and we don't enough cycles to completely test them.

Instead of excluding `FlywayAutoConfiguration`, we think it's worth the effort to embrace it. `MySQLDataSourceProvider#flywayMigrate` and `PostgresDataSourceProvider#flywayMigrate` are remnants of 2.31.x Guice based dependency injection. If you have the time, we highly encourage you to [configure Flyway in Spring Boot.](https://docs.spring.io/spring-boot/docs/2.1.1.RELEASE/reference/html/howto-database-initialization.html#howto-execute-flyway-database-migrations-on-startup)

PR #2227, opened 2021-05-05
Use Spring Boot's auto configuration to create an ObjectMapper

aravindanr:
…ectMapperProvider is used by the client and tests.

Pull Request type
----

Changes in this PR
----

_This is part of the effort to embrace more Spring Boot features. Custom code that creates an ObjectMapper is replaced with configuration that Spring Boot uses to configure the default ObjectMapper that it creates._


common/src/test/java/com/netflix/conductor/common/utils/SummaryUtilTest.java
```diff
@@ -1,40 +1,41 @@
...
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+@ContextConfiguration(classes = {TestObjectMapperConfiguration.class, SummaryUtilTest.SummaryUtilTestConfiguration.class})
 @RunWith(SpringRunner.class)
 public class SummaryUtilTest {
 
-    @TestConfiguration
+    @Configuration
```
apanicker-nflx, line 38: Why change from `TestConfiguration` in this test class?

common/src/main/java/com/netflix/conductor/common/config/ObjectMapperConfiguration.java
```diff
@@ -1,26 +1,34 @@
...
 
+import javax.annotation.PostConstruct;
+
 @Configuration
 public class ObjectMapperConfiguration {
 
-    @Bean
-    public ObjectMapper objectMapper() {
-        return new ObjectMapperProvider().getObjectMapper();
+    @Autowired
```
apanicker-nflx, line 25: Is it possible to change this to constructor injection?

apanicker-nflx (approved):

common/src/main/java/com/netflix/conductor/common/config/ObjectMapperConfiguration.java
aravindanr: Yes. Switched to constructor injection

apanicker-nflx (approved):

PR #2230, opened 2021-05-06
Remove workflows from decider queue when paused; add when resumed

apanicker-nflx:
Pull Request type
----

- [x] Feature

Changes in this PR
----
_Workflow id will be removed from the decider queue when a workflow is paused. Consequently, the workflow id will be added to the decider queue when a workflow is resumed. This prevents PAUSED workflows from being marked as TIMED_OUT. Additionally, the `lastRetriedTime` for a resumed workflow is reset so that the workflow is not TIMED_OUT when processing is resumed._

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
@@ -12,6 +12,22 @@
  */
 package com.netflix.conductor.core.execution;
 
+import static com.netflix.conductor.common.metadata.tasks.Task.Status.CANCELED;
+import static com.netflix.conductor.common.metadata.tasks.Task.Status.FAILED;
+import static com.netflix.conductor.common.metadata.tasks.Task.Status.FAILED_WITH_TERMINAL_ERROR;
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -1289,7 +1290,11 @@ public void resumeWorkflow(String workflowId) {
                     "Current status is " + workflow.getStatus().name());
         }
         workflow.setStatus(WorkflowStatus.RUNNING);
+        workflow.setLastRetriedTime(System.currentTimeMillis());
```
aravindanr, line 1293: What is the significance of this change? Does it affect the workflow timeout in any way?

aravindanr (approved):

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
apanicker-nflx: yes, this ensures that the workflow will not be timed out because of the time it was PAUSED. Currently, we do not have an execution time measurement, this acts as a stop-gap and is used similarly in other uses.

PR #2247, opened 2021-05-12
Allow retrying timed_out workflow with no unsuccessful tasks

apanicker-nflx:
Pull Request type
----
- [x] Feature

Changes in this PR
----
_Allow workflows that TIMED_OUT because of the timeoutSeconds configuration at the workflow level to be retried. In such workflows, there may not be any unsuccessful tasks, but the workflows should be able to be retried by the user._

_The workflow will be set to RUNNING state by the action and will be acted upon later by the sweeper._

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
@@ -18,6 +18,7 @@ import static com.netflix.conductor.common.metadata.tasks.Task.Status.FAILED_WIT
 import static com.netflix.conductor.common.metadata.tasks.Task.Status.IN_PROGRESS;
 import static com.netflix.conductor.common.metadata.tasks.Task.Status.SCHEDULED;
 import static com.netflix.conductor.common.metadata.tasks.Task.Status.SKIPPED;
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -632,9 +633,11 @@ private void retry(Workflow workflow) {
             }
         }
 
-        if (retriableMap.values().size() == 0) {
+        // if workflow TIMED_OUT due to timeoutSeconds configured in the workflow definition,
+        // it may not have any unsuccessful tasks that can be retried
+        if (retriableMap.values().size() == 0 && workflow.getStatus() != WorkflowStatus.TIMED_OUT) {
```
aravindanr, line 638: Is it to account for the scenario, when the last task was successful and the workflow is TIMED_OUT before scheduling the next task?

aravindanr (approved):

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
apanicker-nflx: yes, because the lazy timeout evaluation will mark the workflow as TIMED_OUT before scheduling the next task.

PR #2273, opened 2021-05-25
Oracle Persistence

maheshyaddanapudi:
Pull Request type
----

- [x] Feature
- [x] Build related changes
- [x] Other (please describe): Docker / Docker Compose files included.

Changes in this PR
----

- Supports Oracle Database version > 12.2 - Tested well with 19C
- Updated Dockerfile's startup.sh script (java -jar .... | tee server.log) to write logs to both Docker container as well as server.log file.

Feature
--

1. Oracle Persistence / Option to use Oracle Database as persistence unit.
2. Docker Compose example with Oracle Container.

Technical Note
--

1. Utilizes Oracle Database version 11G for Unit Tests 

> Due to GuitHub GI Container disk space limitation, no version > 11G can be used for Unit Tests. 
Can be easily upgraded to 18 XE or 19C Enterprise once GitHub Action updates size capacity which they are apparantly working on).

2. Due to a technical version dependency, spring.flyway.locations cannot be automatically set to db/migration_oracle (similar to postgres-persistence) and **has to be set through .properties file while running the jar along with conductor.db.type**

> Refer to docker/server/config-oracle.properties for example
 
3. Flyway 6.4.4 (5.2+) - Only compatible with Oracle 12.2+
4. Flyway 4.0.3 (<5.2) - Used for Unit Tests as Oracle 11G XE is used for Oracle Container.

Alternatives considered
----

Many other persistence alternatives like Cassandra, MySQL, Postgres already exist. This is just an addition to persistence options.


maheshyaddanapudi:
It took a good chunk of a week to finish this module.

First the Flyway Initialization SQLs had to be changed / converted to match Oracle Database format. It wasn't easy but was straight forward. However, tiny issues like "LIMIT ?" in MySQL being changed to "ROWNUM < ?" causes issues as it has to be a <= and not a < and so on.

After moving on, the module development was fairly simple and very easy. The tough part was the Unit tests.
Currently there are no official images for Oracle like MySQL or Postgres and we need to make our own image --> [Build Your Oracle Image](https://github.com/oracle/docker-images/blob/main/OracleDatabase/SingleInstance/README.md) 

Then the next / biggest hurdle is, the compatibality matrix as shown below.

| Flyway Version | Oracle Version          | GitHub Action          |
| ------- | ------------------ | ------------------ |
| Flyway 5.2+ | Oracle 12.2+ | GitHub Action - CI Pipeline : CI Container / VM Size constraint for Unit tests |
| Flyway 5.1- | Oracle 11G | GitHub Action - CI Pipeline : CI Container / VM Size is sufficient for Unit tests |

The above scenario can only be resolved in the below ways.

1) GitHub Action (team is apparantly working on this) has to increase the CI VM / Container disk size to accommodate bigger Unit Test Docker Images.
    - Once this is completed, a Docker Image like 18C XE [Oracle 18C XE JUnit Tester Image](https://hub.docker.com/layers/conductorboot/oracle/18.4.0-xe-slim-test/images/sha256-afc438582ebf60798036030653a4db3e715dc41c592fd2eec7780dd84feeecd6?context=explore) can be used along with existing Flyway 6.4.4 version.
    - Also this would remove the ambiguity of Flyway version and can be universally 6+ across Conductor modules.
2) Use Flyway < 5.1 version and use Oracle 11G Docker Image [Oracle 11G XE JUnit Tester Image](https://hub.docker.com/layers/conductorboot/oracle/11g-xe/images/sha256-8b740e77d4b90add693fedb22938f340821e89665fb58ecaeeb0dace853b9ee5?context=explore)
    - This would resolve the issue of GitHub Actions VM / Container Disk space issue as it's relatively smaller image.
    - However, there is a catch, as after the build / during runtime, the Flyway version 6.4.4 from parent takes precedence.
        - Due to this, the final Conductor Boot Jar only supports Oracle 12.2+ which is fine as even the official support for versions below 18 are ended or ending soon from Oracle themselves.
        - The issue is, for obvious reasons, the Flyway locations have to be set as db/migration_oracle or something other than db/migration (as MySQL already uses it). To do this, no matter how many @Configuration or @Component are built, with @Order of HIGHEST_PRECEDENCE and @ApplicationContext .getBean("flyway").setLocations("classpath:db/migrations_oracle") etc. , the Flyway bean 
            - is already initialized by the point it reaches the Oracle-Persistence.
            - cannot be ovverriden as the IDE / compilation time code only has visibility to Flyway < 5.1 and has a different constructor than 6.4.4 which the parent module ovverrides with.
            - The only way to tackle this is, to set spring.flyway.locations=classpath:db/migration_oracle along with conductor.db.type=oracle for it to work.

As we have control only on one of the above mentioned solutions, the second approach was chosen.
Bottomline, Flyway 4.0.3 (< 5.1) + Oracle 11G (< 12.2) is used for Unit Tests and Flyway 6.4.4 (> 5.2) + Oracle 19C Enterprise (> 12.2) is used for Runtime.

For this, I had to 

- Create (Near to official) Oracle 19C Enterprise Image - For official use - any purpose not just for Conductor though.
- Create (Near to official) Oracle 19C Enterprise Test Image - For Unit Tests in future of on other CI Pipeline platforms which provide higher capacity CI Containers / VMs. This is also generic and anyone can use it for their own Unit Tests as well.
- Find and Clone Oracle 11G XE Image - For current Unit Tests on GitHub Actions CI. It was very hard to find a working Image to clone first.
- Find and Clone Oracle 18G XE Image - For future Unit Tests on GitHub Actions CI, even if the disk space is just slightly increased like say 5G at max. I was easier to find this than 11G but still, took some effort, as I had to spin up each and every image and test if it works in local and so on.
    
It was only then I was able to 

1) Pass GitHub CI Pipeline with all Unit Tests cleared.
2) Build the perfect Conductor Boot Jar with Oracle Persistence.
3) Perform Runtime Test with all persistence units, to ensure zero impact - MySQL, Im Memory and finally Oracle

And lastly

4) Raise this PR :)

maheshyaddanapudi:
@aravindanr @apanicker-nflx @kishorebanala 

Please help with your thoughts, suggestions and if it seems good, then reviews.

apanicker-nflx:
Hi @maheshyaddanapudi, thanks for your valuable contributions. 

While this is definitely useful to increase the extensibility of Conductor, it also increases the surface area of support/maintenance for the project. This is currently a pain point since Netflix internally does not use the MySQL/Postgres persistence layers and for any issues that are reported for these modules, we have to rely on the community to debug/reproduce/rootcause/fix them. While the community is very active and and prompt in helping out with resolutions, there are cases when the support can be a long drawn out process. Additionally, it also inhibits the evolution of the project (for example, migrating to spring boot was slowed down to accommodate these modules with proper testing). 

Since we do see the value in this contribution, we feel that it would be best served in a repo outside Conductor which depends on the `conductor-core` and `conductor-common`. This would help us direct support requests directly to the repo to be answered by the maintainers and also help with faster evolution of the project.

maheshyaddanapudi:
Hi @apanicker-nflx 

I understand.

I have rather added the link to forked repo in RELATED.md and created a separate PR https://github.com/Netflix/conductor/pull/2295

Please help in approving the new PR.

PR #2285, opened 2021-06-04
Microsoft SQL server persistence (v3)

TwoUnderscorez:
Pull Request type
----

- [x] Feature

Changes in this PR
----

- sqlserver lock implementation for decider service
- sqlserver persistence implementation
  - conductor queues: shared: requires the decider service lock to be enabled. local_only: each conductor instance gets it's own queue (identified by the `LOCAL_RACK` property).
 
Issue #2075


apanicker-nflx:
Hi @TwoUnderscorez , thanks for your valuable contributions.

While this is definitely useful to increase the extensibility of Conductor, it also increases the surface area of support/maintenance for the project. This is currently a pain point since Netflix internally does not use the MySQL/Postgres persistence layers and for any issues that are reported for these modules, we have to rely on the community to debug/reproduce/rootcause/fix them. While the community is very active and and prompt in helping out with resolutions, there are cases when the support can be a long drawn out process. Additionally, it also inhibits the evolution of the project (for example, migrating to spring boot was slowed down to accommodate these modules with proper testing).

Since we do see the value-add in this contribution, we feel that it would be best served in a repo outside Conductor which depends on the `conductor-core` and `conductor-common` modules. This would help us direct support requests directly to your repo to be answered by the maintainers and also help with faster evolution of the project.

TwoUnderscorez:
@apanicker-nflx While I understand where you're coming from, I unfortunately won't be able to do what you've suggested.
I personally think that the SQL Server is one of the most feature-rich modules and therefore should be in the main repo.
* It supports fully all the workflow DAOs
* Can act as a decider lock service
* Highly configurable and documented
* The SQL is written cooperatively with a DBA with 20+ years of experience.
* It can be set up in any of these modes with only an SQL Server (no other technologies required):
    1. Solo conductor with solo SQL Server (1 queue zone / no decider lock required)
    2. Conductor per zone with HA (handled by Conductor) and DR (handled by SQL Server always-on cluster)
        a. 1 Conductor queue shard, 1 conductor locking namespaces -> A cluster of Conductors with them dynamically assigning work depending on how big / busy each environment is. Any workflow started by any Conductor can be run by any other Conductor instance/zone.
        b. N Conductor queue shards, no decider lock required -> A cluster of Conductor but all workflows are local to the Conductor they were started by. A custom scheduling service is probably required for this which will devide the workload.
    3. The ultimate setup - given an SQL Server always-on cluster running on 3 regions, and you have a k8s cluster on each region, deploy 2 Conductor instances on each k8s, both of the Conductor instances in each region share a queue shard and a locking namespace. Why 2 Conductor instances? In case one of them goes down and k8s is having trouble bringing it back up. We also wrote an external scheduling microservice which will do the right thing if a whole region goes down. It's not ready yet, so till then we're using 2a.

While Dynomite is deprecated and is extremly difficult to set up.
MySQL/Postgres - no way to devide queues by zone / zookeeper/redis is required for HA and it's not really documented
Cassandra - verry little documentation and no full implemntation. 
So just redis is fully working? I haven't checked, but I think that sqlserver is either as feature-rich as redis or richer.

If you decide to not accept this, I will just close my PRs and we'll just maintain an internal fork. I am not a Java developer and I don't want to be. I don't know how to do what you've described and unfortunately, I don't have the time and patience to learn how. 

Also please note, this is not yet updated with version 2 of the PR, that's why it's still marked as a draft. You should probably not look at my code here until I mark it as "Ready for review".

I will ask you to reconsider the idea of merging this module to your repo and update me, thanks.

apanicker-nflx:
Hi @TwoUnderscorez Thank you very much for listing out the pros in favor of the sql-server-persistence module. While we agree with the feature completeness of this module, the ongoing support load coupled with the proprietary nature of the database itself would make it difficult for us to incorporate it as part of the project. Additionally, evolution of the project becomes cumbersome with a large number of modules which are unused by the maintainers.

We can work with you to move this contribution into a separate repo outside of Conductor and link to this from our documentation which would also provide users a directed forum to ask questions and engage with the module developers.

TwoUnderscorez:
@apanicker-nflx Thanks for taking the time to respond, but no. The main reason we chose to contribute our code to your repo is so that we can be mostly confident that our code won't be broken by anybody else once it's in the `main` branch. If we can only publish this as an external module, we might as well just keep an internal fork. It basically has the same effect, but with less headaches.

Also, I didn't quite understand what you meant by "the ongoing support load coupled with the proprietary nature of the database itself would make it difficult for us to incorporate it as part of the project." but that doesn't seem to matter, even though SQL Server is the third most popular DB according to Stack Overflow. I don't think it's any more proprietary than MySQL: INSERT IGNORE / Indexes with IGNORE_DUP_KEY, ON DUPLICATE KEY UPDATE / MERGE statements, are all proprietary nonstandard SQL things that Microsoft and Oracle chose to implement differently. And if you're talking about the availability of SQL Server itself, a free version can be installed on Windows, Linux or run as a docker container as I've shown in the `docker-compose.yaml` file in the module.

Anyhow, if other layers will be split into different repos, we might reconsider publishing our module as an external module, but right now, we don't see a reason why our module should be treated differently than other modules (also the whole me not being a Java developer, I would prefer to copy from an existing external module rather than try to be the first to publish one).

Hopefully you'll understand where I'm coming from, but for now I'm closing all SQL Server related PRs, but if you ever change your mind, please do let me know as I'll happily work with you to make this work.

PR #2330, opened 2021-06-22
change combo_queue_message index to use descending priority for huge performance increase

rickfish:
change combo_queue_message index to use descending priority for huge performance increase on databases with millions of workflow and task rows

Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----When polling, the postgres implementation does a query that orders by descending priority. For databases with lots of workflows, an index created using ascending priority causes these queries to be extremely slow. Changing the index to use descending priority dramatically increases performance on those polling queries, especially with databases with millions of workflows/tasks.

Alternatives considered
----

### DIFF
--- a/postgres-persistence/src/main/resources/db/migration_postgres/V7__new_qm_index_desc_priority.sql
+++ b/postgres-persistence/src/main/resources/db/migration_postgres/V7__new_qm_index_desc_priority.sql
@@ -1,0 +1,3 @@
+DROP INDEX IF EXISTS combo_queue_message;
+
+CREATE INDEX combo_queue_message ON queue_message USING btree (queue_name , priority desc, popped, deliver_on, created_on)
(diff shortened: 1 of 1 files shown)


rickfish:
This build failed with this message: Could not resolve io.chaossystems.grpc:grpc-healthcheck:1.0.1. This PR only added one file and it is a resource file. No build.gradle files were changed and versionsOfDependencies.json was not changed so not sure why the build would fail with this error.

kishorebanala:
@rickfish This has been [fixed](https://github.com/Netflix/conductor/commit/9b4573b9ca7576dd4e75b71029e96536abfcc576), Rebasing should fix.

Also, Please consider porting these by creating another similar PR against 3.x branch.

PR #2331, opened 2021-06-22
Fix usage of 'in' clause for postgres query in processUnacks and processAllUnacks

rickfish:
Fix usage of 'in' clause for postgres query in processUnacks and processAllUnacks

Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----The current way that the update portion of the processAllUnacks method works is that a String is created concatenating all of the message ids that should be updated and that String is passed into the query as a parameter to replace the 'in' variable. This does not work in Postgres. We must use the standard Query.generateInBindings method to generate the parameter in the proper syntax.

Issue #

Alternatives considered
----

### DIFF
--- a/postgres-persistence/src/main/java/com/netflix/conductor/dao/postgres/PostgresQueueDAO.java
+++ b/postgres-persistence/src/main/java/com/netflix/conductor/dao/postgres/PostgresQueueDAO.java
@@ -190,6 +190,8 @@ public class PostgresQueueDAO extends PostgresBaseDAO implements QueueDAO {
 
         logger.trace("processAllUnacks started");
 
+		// Get queue_name and message_id because those are in an index so using both increases performance
+		
         getWithRetriedTransactions(tx -> {
             String LOCK_TASKS = "SELECT queue_name, message_id FROM queue_message WHERE popped = true AND (deliver_on + (60 ||' seconds')::interval)  <  current_timestamp limit 1000 FOR UPDATE SKIP LOCKED";
 
--- a/postgres-persistence/src/main/java/com/netflix/conductor/dao/postgres/PostgresQueueDAO.java
+++ b/postgres-persistence/src/main/java/com/netflix/conductor/dao/postgres/PostgresQueueDAO.java
@@ -17,15 +17,16 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
 import com.google.common.util.concurrent.Uninterruptibles;
 import com.netflix.conductor.core.events.queue.Message;
-import com.netflix.conductor.core.execution.ApplicationException;
 import com.netflix.conductor.dao.QueueDAO;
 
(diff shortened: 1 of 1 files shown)


rickfish:
This build failed with this message: Could not resolve io.chaossystems.grpc:grpc-healthcheck:1.0.1. This PR only changed one file and it would not have impacted any build.gradle files nor the versionsOfDependencies.json file so not sure why the build would fail with this error.

mactaggart (approved):
LGTM. Thanks!

postgres-persistence/src/main/java/com/netflix/conductor/dao/postgres/PostgresQueueDAO.java
```diff
@@ -190,54 +191,85 @@ public void processAllUnacks() {
         logger.trace("processAllUnacks started");
 
         getWithRetriedTransactions(tx -> {
-            String LOCK_TASKS = "SELECT message_id FROM queue_message WHERE popped = true AND (deliver_on + (60 ||' seconds')::interval)  <  current_timestamp FOR UPDATE SKIP LOCKED";
+            String LOCK_TASKS = "SELECT queue_name, message_id FROM queue_message WHERE popped = true AND (deliver_on + (60 ||' seconds')::interval)  <  current_timestamp limit 1000 FOR UPDATE SKIP LOCKED";
```
mactaggart, line 194: Do you think there's value in making the batch size configurable? Did you test with different values? 1000 seems like a reasonable default.

rickfish: @mactaggart I had thought about that but I can't think of a situation where someone would actually figure out that for their database there is a better value to use and there are already so  many properties  that I didn't want to add another one.

kishorebanala:
> This build failed with this message: Could not resolve io.chaossystems.grpc:grpc-healthcheck:1.0.1. This PR only changed one file and it would not have impacted any build.gradle files nor the versionsOfDependencies.json file so not sure why the build would fail with this error.

@rickfish This has been [fixed](https://github.com/Netflix/conductor/commit/9b4573b9ca7576dd4e75b71029e96536abfcc576), Rebasing should work.

Also, Please consider porting these by creating another similar PR against 3.x branch.

rickfish:
Added a comment, merged the changes and got the same error on the build. Not used to rebasing so I added the comment. What is the process for merging to 3.x? What branch do I merge to?

apanicker-nflx:
@rickfish The `main` branch is used for 3.0.x and `2.31` branch is used for `2.31.x` versions.

rickfish:
@apanicker-nflx created a pull request for 3.x, how to get around the io.chaossystems.grpc:grpc-healthcheck:1.0.1 build error for 2.31?

PR #2344, opened 2021-07-07
fix for retrying subworkflows within dynamic fork join

apanicker-nflx:
Pull Request type
----
- [x] Bugfix

Changes in this PR
----
- Retrying subworkflows that are within a dynamic fork join will correctly update parent workflow state
- Upgrade to SpringBoot 2.3.12.RELEASE
- Added a metric for tracking workflow starts
- Updated testcontainers version

### DIFF
--- a/contribs/src/test/java/com/netflix/conductor/contribs/tasks/http/HttpTaskTest.java
+++ b/contribs/src/test/java/com/netflix/conductor/contribs/tasks/http/HttpTaskTest.java
@@ -70,7 +70,7 @@ public class HttpTaskTest {
 
     @ClassRule
     public static MockServerContainer mockServer = new MockServerContainer(
-            DockerImageName.parse("mockserver/mockserver"));
+        DockerImageName.parse("mockserver/mockserver"));
 
     @BeforeClass
     public static void init() throws Exception {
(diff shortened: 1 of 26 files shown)


contribs/src/test/java/com/netflix/conductor/contribs/tasks/http/HttpTaskTest.java
```diff
@@ -68,12 +67,13 @@
...
     private static String JSON_RESPONSE;
 
-    @ClassRule
-    public static MockServerContainer mockServer = new MockServerContainer(
-            DockerImageName.parse("mockserver/mockserver"));
+    public static MockServerContainer mockServer;
 
     @BeforeClass
     public static void init() throws Exception {
+        mockServer = new MockServerContainer(DockerImageName.parse("jamesdbloom/mockserver:mockserver-5.11.2"));
```
aravindanr, line 74: If this change does not solve the OOM error, it should be reverted. The `@ClassRule` is less code and more readable.

core/src/main/java/com/netflix/conductor/metrics/Monitors.java
```diff
@@ -222,6 +221,11 @@ public static void recordWorkflowTermination(String workflowType, WorkflowStatus
             "" + ownerApp);
     }
 
+    public static void recordWorkflowStartSuccess(String workflowType, String version, String ownerApp) {
+        counter(classQualifier, "workflow_start_success", "workflowName", workflowType, "version", version, "ownerApp",
+            "" + ownerApp);
```
aravindanr, line 226: This is tangential to the change.  The `"" + ownerApp` produces a String with the characters "null" if the value of `ownerApp` is `null`. Is that intention here? Do we want add a tag to the counter whose value is the String "null"? I see this pattern in some other methods in this class as well.

aravindanr (changes requested):

contribs/src/test/java/com/netflix/conductor/contribs/tasks/http/HttpTaskTest.java
apanicker-nflx: unfortunately, it does not. Reverted

core/src/main/java/com/netflix/conductor/metrics/Monitors.java
apanicker-nflx: changed to use `StringUtils.defaultIfBlank(ownerApp, "unknown")`

PR #2355, opened 2021-07-10
feature: allow custom logic in workflow for task retries

Jiehong:
Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Changes in this PR
----

This is a proposal to allow workflows to decide _if_ a task should be retried according to its output.
A workflow is the context within which the task is run, and this changes when/if a task should be retried.
This logic would be evaluated by the workflow decider.

I propose this by using a jq query that evaluates to "true" or anything else otherwise.

Why jq?
- makes parsing the output of the task easy (because the output is some kind of json)
- also allows for regexps

Alternatives:

- put the task to be retried in a do_while and followed by a decision: possible, but quite cumbersome.
- use a javascript (Nashorn) expression (just like in the DECISION task): not easier, and with Nashorn going away soon, I don't find it compelling.

Drawbacks:

- the workflow decider could be stuck for a while if the jq expression takes too long. Its execution should be bound to a maximum time (ex: 200ms) to avoid this.

Note: this PR contains a documentation change only at this stage. This is normal, and is here to allow discussing the idea first, and going forward with the implementation later.

### DIFF
--- a/docs/docs/configuration/taskdef.md
+++ b/docs/docs/configuration/taskdef.md
@@ -6,7 +6,6 @@ Conductor maintains a registry of worker tasks.  A task MUST be registered befor
 {
   "name": "encode_task",
   "retryCount": 3,
-  
   "timeoutSeconds": 1200,
   "pollTimeoutSeconds": 3600,
   "inputKeys": [
--- a/docs/docs/configuration/workflowdef.md
+++ b/docs/docs/configuration/workflowdef.md
@@ -20,6 +20,7 @@ Workflows are defined using a JSON based DSL.
       "name": "deploy",
       "taskReferenceName": "d1",
       "type": "SIMPLE",
+      "retryIfJqIsTrue": "if .subkey == 4 then true else false end",
       "inputParameters": {
         "fileLocation": "${encode.output.encodeLocation}"
       }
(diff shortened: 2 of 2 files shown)


aravindanr:
Thanks for the creating the pull request. In the future, please use the Discussion feature for brainstorming ideas.

Reading between the lines, i presume that you want to retry a task without marking it as `FAILED`. This is problematic in several ways, some being,

1. There will be no clear indication in the state machine on whether the task is being retried or not. A Task that is in `IN_PROGRESS` could technically be in the process of being retried.
2. It also complicates the state machine evaluation, since it needs to react to changes in a task's output data. Output data should be opaque to Conductor and the evaluation engine should not make decisions based on it.

Hope this makes sense. Let me know if you need further clarifications.

Jiehong:
@aravindanr thanks for pointing me to the discussion tab. I never noticed it I guess. I'll use it for next time.

Regarding your points, perhaps they can be compatible:

- If the task is FAILED already, then "retryIfJqIsTrue" is not evaluated (aka today's behaviour)
- If not (like COMPLETED) and the number of retries is not exhausted and retryIfJqIsTrue is not null, then "retryIfJqIsTrue" is evaluated its output is checked:

- if "true": task is set to FAILED, with a message ("set for retry because of retryIfJqIsTrue"), and the normal retry mecanism is used
- if anything other than "true": status is not changed, and a message is added (something like "retryIfJqIsTrue did not evaluate to false")

What do you think?

aravindanr:
Thanks for the additional details. We still think that `retryIfJqIsTrue` adds more cognitive load to a general Conductor user and should not be part of Conductor. It's also a very niche use-case. Adding that flag also adds more complexity to the state machine evaluation. We feel this adds more complexity to an already complex codebase.

Jiehong:
@aravindanr : alright then. I'm closing this PR.

PR #2360, opened 2021-07-13
SubWorkflow task will be auto-retried on BACKEND_ERROR

aravindanr:
Pull Request type
----

- [x] Feature

Changes in this PR
----

_On backend errors, clients are programmed to retry, but Conductor internally does not retry. For SubWorkflow, BACKEND_ERROR exceptions leave the task in SCHEDULED state, allowing it to retried by SystemTaskCoordinator at a later time._

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
@@ -103,7 +103,6 @@ public class WorkflowExecutor {
     private final SystemTaskRegistry systemTaskRegistry;
 
     private long activeWorkerLastPollMs;
-    private final long queueTaskMessagePostponeSecs;
     public static final String DECIDER_QUEUE = "_deciderQueue";
     private static final String CLASS_NAME = WorkflowExecutor.class.getSimpleName();
     private final ExecutionLockService executionLockService;
@@ -119,7 +118,6 @@ public class WorkflowExecutor {
 
     private static final Predicate<Task> NON_TERMINAL_TASK = task -> !task.getStatus().isTerminal();
 
-    @SuppressWarnings("SpringJavaInjectionPointsAutowiringInspection")
     public WorkflowExecutor(DeciderService deciderService, MetadataDAO metadataDAO, QueueDAO queueDAO,
         MetadataMapperService metadataMapperService, WorkflowStatusListener workflowStatusListener,
         ExecutionDAOFacade executionDAOFacade, ConductorProperties properties,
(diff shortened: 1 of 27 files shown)


core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -1434,38 +1435,51 @@ public void executeSystemTask(WorkflowSystemTask systemTask, String taskId, long
...
+                    // postpone based on callbackTime
+                    queueDAO.postpone(queueName, task.getTaskId(), task.getWorkflowPriority(), callbackTime);
+                    LOGGER.debug(
+                            "Task: {} postponed in taskQueue: {} since the task status is {} with callbackAfterSeconds: {}",
+                            task, queueName, task.getStatus(), callbackTime);
+                    return null;
+                }, null, null, 2, "Postponing Task message in queue for taskId: " + task.getTaskId(), "postponeTaskMessage");
             }
 
-            updateTask(new TaskResult(task));
```
aravindanr, line 1465: @apanicker-nflx I think this call is not appropriate here, as it seems to do more than just updating the task. Please verify.

For this particular case, where we want to do retries for BACKEND_ERRORs, even if the `WorkflowSystemTask.start` leaves or sets the `Task` in `SCHEDULED` state, [updateTask](https://github.com/Netflix/conductor/blob/main/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java#L964-L968) and [TaskResult constructor](https://github.com/Netflix/conductor/blob/main/common/src/main/java/com/netflix/conductor/common/metadata/tasks/TaskResult.java#L86-L88) change it to `IN_PROGRESS`. I removed the call and wrote only the essential and relevant parts of `updateTask`.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -1434,38 +1435,51 @@ public void executeSystemTask(WorkflowSystemTask systemTask, String taskId, long
...
-                default:
-                    break;
+            if (task.getOutputData() != null && !task.getOutputData().isEmpty()) {
+                deciderService.externalizeTaskData(task);
             }
 
-            if (!task.getStatus().isTerminal()) {
+            // Update message in Task queue based on Task status
+            // Stop polling for asyncComplete system tasks that are not in SCHEDULED state
+            if (systemTask.isAsyncComplete(task) && task.getStatus() != SCHEDULED) {
```
aravindanr, line 1457: @apanicker-nflx Instead of removing the task in the next run, i changed the logic to remove it eagerly. Previously, an "async complete" task in `IN_PROGRESS` state is postponed. And in the next run, the "async completeness" is checked and removed from the queue.

apanicker-nflx (approved):

Issue #2382, opened 2021-08-01
[BUG]: AMQP Unknown Tags -  Messages can't be acked because channel is recreated each time there is a network failure


heinrichcohn:
**Describe the bug**
I'm using Rabbitmq to send/receive messages from or to workflows. I found out that in case of a network failure, messages that were taken from a queue but not yet acked, conductor won't be able to ack them at all until the server is restarted. 

This happens because in case of a network error the channel is __recreated each time!__ . From the moment the channel is recreated  all delivery tags from previous messages can't be use for acknowledgement  and  "**Unknown Tag**" errors are thrown each time an ack attempt is made. In addition, trying to acknowledge a message with delivery tag created in a different channel  will cause the channel to __close!__.

**Details**
Conductor version: 3.X
Workflow definition:
Task definition:
Event handler definition:

**To Reproduce**
1. Enable amqp in conductor config and start
2. Created event-handler that listens to an amqp queue
3. Start sending messages to the queue
4. Simulate a network error between the queue and conductor
5. Connection errors will appear in conductor logs
6. Restore connection between  the queue and conductor
7. Acknowledgement of messages taken before the failure will get  unknow_tag errors in the logs from this moment (these message will appear as unacked in Rabbit)
8. Restarting conductor solves the problem

**Expected behavior**
After a connection between a queue and conductor is restored, messages taken before the failure will be acknowledged successfully by conductor. Aqmp client will be responsible of restoring the amqp channel and handle delivery tags that were created before the failure so they can be acknowledged in the new restored channel.

**Logs**
2020-11-15 09:56:18,622  INFO [RxComputationScheduler-1] (com.netflix.conductor.contribs.queue.amqp.AMQPObservableQueue) - Batch from conductor.order.cancelled to conductor is a0de6fe3-fe2e-4579-a657-874c6d29306b={"OrderId":"14043973","CancellationReason":"Automation","workflow_meta":{"sourceQueue":"Comp.OrderOrchestrator.Order.Cancelled","workflowId":"85654cc7-db84-4a6b-8a66-d9501d918f27"}}
2020-11-15 09:56:18,708  INFO [pool-13-thread-1] (com.netflix.conductor.dao.dynomite.RedisExecutionDAO) - updating event execution conductor.qa.EVENT_EXECUTION.PMT_order_cancelled.amqp_queue:conductor.order.cancelled.a0de6fe3-fe2e-4579-a657-874c6d29306b
[...]

PR #2383, opened 2021-08-01
Fix amqp channel and connection creation in order to prevent "unknown delivery tag"

heinrichcohn:
Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
After this PR amqp connection and channel will be created once instead of closing and opening channels each time there is a failure.
Upgrade amqp-client version to 5.13.0

Amqp connection/channel objects will be created once and amqp client will be responsible of restoring channels in case of a failure. 
After this PR Conductor will be able to acknowledge messages with delivery tags from previous channels and process those messages successfully

_Issue_ 
https://github.com/Netflix/conductor/issues/2382

Alternatives considered
----

No

### DIFF
--- a/contribs/src/main/java/com/netflix/conductor/contribs/queue/amqp/AMQPObservableQueue.java
+++ b/contribs/src/main/java/com/netflix/conductor/contribs/queue/amqp/AMQPObservableQueue.java
@@ -60,7 +60,6 @@ public class AMQPObservableQueue implements ObservableQueue {
     private final int batchSize;
     private final boolean useExchange;
     private int pollTimeInMS;
-    private boolean isConnOpened = false, isChanOpened = false;
 
     private final ConnectionFactory factory;
     private Connection connection;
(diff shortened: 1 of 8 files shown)


aravindanr:
@ritu-p , can you please review this PR?

heinrichcohn:
@aravindanr  @apanicker-nflx 

As part of this PR i'm upgrading the version of the amqp-client since the current version (5.9.0) contains an issue related to consumer recovery and consumer tags.
Since I updated the relevant dependencies.lock files, the compilation creates the final jar with the right amqp-client version, but if i delete the dependencies.lock files and run "_./gradlew generateLock updateLock saveLock_" to recreate them, the lock files of the **server** and **test-harness** will have the older version 5.9.0 and not 5.13.0. This situation will cause that the final boot jar will have the amp-client:5.9.0 and not 5.13.0.

Updating the following files fix the problem:
**./server/build.gradle**
_implementation( group: 'com.rabbitmq', name: 'amqp-client'){ version{require "${revAmqpClient}"}}_
**and adding to ./server/build.gradle**
_testImplementation( group: 'com.rabbitmq', name: 'amqp-client'){ version{require "${revAmqpClient}"}}_

Are you ok with this approach or you have a better idea?

Thanks

contribs/src/main/java/com/netflix/conductor/contribs/queue/amqp/AMQPObservableQueue.java
```diff
@@ -366,27 +359,23 @@ public AMQPObservableQueue build(final boolean useExchange, final String queueUR
     }
 
     private Channel getOrCreateChannel() {
-        if (!isConnOpened) {
-            open();
-        }
-        // Return the existing channel if it's still opened
-        if (channel != null && isChanOpened) {
+        // Return the existing channel if it was created
+        if (channel != null) {
```
ritu-p, line 363: Can we also check it the channel is not closed? An existing channel which is already closed will also cause issues

contribs/src/main/java/com/netflix/conductor/contribs/queue/amqp/AMQPObservableQueue.java
```diff
@@ -366,27 +359,23 @@ public AMQPObservableQueue build(final boolean useExchange, final String queueUR
...
             return channel;
         }
         // Channel creation is required
         try {
-            channel = null;
+            connect();
             channel = connection.createChannel();
             channel.addShutdownListener(cause -> {
-                isChanOpened = false;
```
ritu-p, line 381: Would it make sense to retry channel creation in shutdown listener after a configurable time interval?

ritu-p (changes requested):

heinrichcohn:
@ritu-p @aravindanr 
I made the requested changes and update the unit-test.
Can you please review and merge if everything is ok?
Thanks

ritu-p (approved):

ritu-p:
@heinrichcohn I have approved the changes

Issue #2398, opened 2021-08-10
[BUG]: Workflow tasks are not accessable when using workflow archiving with TTL - Redis DAO specific


heinrichcohn:
**Describe the bug**
When archiving workflows using archiving with TTL status listener, their task are not reachable causing the getWorkflow API return Workflows summary with no tasks.  The root cause of this problem is that the WORKFLOW_TO_TASKS mapping is deleted with no TTL during the archiving what leads to a situation in which you have workflows and tasks data ( they have a TTL > 0) but no mapping between them because no TTL was set for the mapping.
Once the TTL expires, Conductor will start retrieving the workflows and their tasks from ES and the problem disappears
Longer the TTL ,  more noticeable the issue will be.

**Details**
Conductor version:3.x
Configuration:
conductor.workflow-status-listener.archival.ttlDuration=3600s
conductor.workflow-status-listener.type=archive

**To Reproduce**
1. Run Conductor with "archive" with TTL enabled. Set a long TTL
2. Run a workflow
3. Right after the workflow finishes, call the api **http://[conductor-endpoint]/api/workflow/[your_workflow_id]?includeTasks=true**
4. The response will contain no tasks data in the 'tasks' field

**Expected behavior**
After running a workflow and calling the getWorkflow API (http://[conductor-endpoint]/api/workflow/[your_workflow_id]?includeTasks=true**), the API should return the tasks data in the 'tasks' field

**Additional context**
This problem is relevant when using Redis Execution DAO + Archive with TTL

apanicker-nflx:
Closing since #2399 is merged

PR #2399, opened 2021-08-10
[redis-persistence] Prevent workflow to tasks mapping from being deleted  immediately during archiving

heinrichcohn:
Prevent workflow_to_tasks mapping from being deleted  immediately during archiving when using archive with ttl  (Redis execution DAO only)

Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
Separate removeTask logic when using archiving with TTL or without TTL
Set TTL to WORKFLOW_TO_TASKS entry instead of deleting it

Issue #2398
This issue is important because if a long TTL is configured in the server,  no tasks execution information  will available until the TTL is reached

Alternatives considered
No


heinrichcohn:
@aravindanr 
Issue created:  #2398

PR #2409, opened 2021-08-17
fix incorrect calculation of task total duration with retries

yannrouillard:
When a task was retried, each task duration but the last was counted
twice in the total computation.

Pull Request type
----

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----

The total task duration is now correctly computed and the metric can now be used accurately to setup alerting

Alternatives considered
----

No alternative implementation considered


apanicker-nflx (commented):
Please add a unit test for this change.

PR #2438, opened 2021-08-29
feat: Add inputTemplate support to workflow metadata

NickTomlin:
Pull Request type
----

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Changes in this PR
----

This adds an `inputTemplate` configuration option to workflow definitions that functions similar to the [`inputTemplate` on task definitions ](https://netflix.github.io/conductor/configuration/taskdef/#using-inputtemplate). 

My current workaround is to define a `jq` task to do things like provide default variables but this does so more elegantly and maps to the current task modelling. 

Alternatives considered
----

- An entirely new field
  - I like the fact that `inputTemplate` is consistent across tasks and workflows, but a new field could have been used here if we wanted to separate the implementations

Questions
---

1. Are there any other places I should add tests? The workflow integration tests seem to be the most logical, but please let me know!
2. I've verified by running the tests locally, but i'm not sure if there's some additional testing/verification I should do manually.

### DIFF
--- a/common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowDef.java
+++ b/common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowDef.java
@@ -302,9 +302,8 @@ public class WorkflowDef extends Auditable {
         return inputTemplate;
     }
 
-    public WorkflowDef setInputTemplate(Map<String, Object> inputTemplate) {
+    public void setInputTemplate(Map<String, Object> inputTemplate) {
         this.inputTemplate = inputTemplate;
-        return this;
     }
 
     public String key() {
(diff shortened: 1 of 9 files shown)


apanicker-nflx (approved):

apanicker-nflx:
>Are there any other places I should add tests? The workflow integration tests seem to be the most logical, but please let me know!

ParametersUtilsTest and WorkflowAndTaskConfigurationSpec provide good coverage for this change. Thanks for adding the tests.

common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowDef.java
```diff
@@ -295,6 +298,15 @@ public void setVariables(Map<String, Object> variables) {
         this.variables = variables;
     }
 
+    public Map<String, Object> getInputTemplate() {
+        return inputTemplate;
+    }
+
+    public WorkflowDef setInputTemplate(Map<String, Object> inputTemplate) {
```
aravindanr, line 305: To keep it consistent with the other setters in the class, can this also return `void`?

aravindanr (approved):

common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowDef.java
NickTomlin: Will do!

PR #2469, opened 2021-09-21
Allow instance to poll for task out of discovery 

manavvasavada:
Pull Request type
----

- [ ] Bugfix
- [X ] Feature
- [ ] Build related changes

Changes in this PR
This is a long pending fix that we need for conductor client in order to be able to poll task from local machine. We have our apps implementing "hacky" logic of creating our own fake eureka clients which have status UP but do not report to discovery. Conductor should be able to support this out of the box.

This change implements a property `conductor.worker.X.poll.out.of.discovery` which will allow the machine to poll for the task even if the discovery status is not UP.  The property name is not perfect and I am open to suggestions :)

This change implements a property `conductor.worker.X.poll.out.of.discovery` which will allow the machine to poll for the task even if the discovery status is not UP.  The property name is not perfect and I am open to suggestions :)

Alternatives considered
Alternatives were to either implement own eureka client or set up break point on the condition and set eureka client as null.

Current implementation is very hacky for all apps to configure a special Eureka client which returns UP status but does not register with eureka service. Which is a very inconsistent state to be in.


aravindanr:
For posterity, the changes in the PR were discussed offline and the notes are shared here.

Conductor client can work without wiring in a `EurekaClient`, which is the right thing to do when running Conductor from a local machine where there is no Eureka registration. Creating a `EurekaClient` with a special status is unnecessary and the right thing to do is to NOT wire a `EurekaClient` where it's not needed.

The property `conductor.worker.X.poll.out.of.discovery` added as part of this PR does not hurt situations where there is no `EurekaClient` but could help in scenarios where it's always available (regardless of local machine or cloud) but should not be used.

client/src/main/java/com/netflix/conductor/client/automator/TaskPollExecutor.java
```diff
@@ -65,15 +66,21 @@
...
+                    .uncaughtExceptionHandler(uncaughtExceptionHandler)
+                    .build());
 
         this.pollingSemaphore = new PollingSemaphore(threadCount);
     }
 
     void pollAndExecute(Worker worker) {
-        if (eurekaClient != null && !eurekaClient.getInstanceRemoteStatus().equals(InstanceStatus.UP)) {
+        Boolean discoveryOverride = Optional.ofNullable(PropertyFactory.getBoolean(worker.getTaskDefName(), OVERRIDE_DISCOVERY, false))
+                .orElseGet(() -> Optional.ofNullable(PropertyFactory.getBoolean(ALL_WORKERS, DOMAIN, null))
```
aravindanr, line 78: Did you mean `PropertyFactory.getBoolean(ALL_WORKERS, OVERRIDE_DISCOVERY, false)` here?

```java
boolean discoveryOverride = Optional.ofNullable(PropertyFactory.getBoolean(worker.getTaskDefName(), OVERRIDE_DISCOVERY, null))
                .orElseGet(() -> Optional.ofNullable(PropertyFactory.getBoolean(ALL_WORKERS, OVERRIDE_DISCOVERY, false));
```

aravindanr (changes requested):

client/src/main/java/com/netflix/conductor/client/automator/TaskPollExecutor.java
manavvasavada: @aravindanr This should be fixed.

aravindanr: Doesn't the default for the first call be null for the `orElseGet()` to be invoked? If its set as `false`, the else block will not be reached.

`PropertyFactory.getBoolean(worker.getTaskDefName(), OVERRIDE_DISCOVERY, null)`

The code could be simplified like this,

```java
boolean discoveryOverride = Optional.ofNullable(PropertyFactory.getBoolean(worker.getTaskDefName(), OVERRIDE_DISCOVERY, null))
                .orElseGet(() -> Optional.ofNullable(PropertyFactory.getBoolean(ALL_WORKERS, OVERRIDE_DISCOVERY, false));
```

manavvasavada: Yes. I have fixed this. If we remov ethe orElseGet in the end, then we need to return Boolean on the second line and not an Optional.

aravindanr (approved):

Issue #1659, opened 2020-04-28
This does not work in go client 


zhou6675021:
HI，
I made a sample according to the official website, but the task was received but not executed,code：

	c := conductor.NewConductorWorker("http://172.16.249.61:8080/api", 1, 10000)

	c.Start("test1", task.Create_Emr_Task, false)

func Create_Emr_Task(t *task.Task) (taskResult *task.TaskResult, err error) {
  //...//
}

Console log：
![image](https://user-images.githubusercontent.com/23535473/80475399-65cae580-897b-11ea-98f6-a777a404db1c.png)

The task was received but not executed， 
Forgive me newbie

apanicker-nflx:
@zhou6675021 I see that you are using the go client for this purpose. The concept of `ack`ing a task has been deprecated and is no longer used. However, this change has not been propagated to the go client. Once a task is polled, it is expected that it will be executed by the worker and a response will be sent to the server within the `responseTimeoutSeconds`. 

A very simple example to get started with using Conductor is available [here](https://netflix.github.io/conductor/labs/beginner/). As seen here, the task is polled and a response is sent after with the execution result (ack is not sent).

PR #2470, opened 2021-09-21
Remove ack from Go client + pass auth to go http client

CHURLZ:
Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [x] Other (please describe): adapt go client to v3 

Changes in this PR
----
- Removed ack step for go client
- Added constructor for Httpclient to allow token based auth

Based on Issue # https://github.com/Netflix/conductor/issues/1659, v3 no longer requires workers to ack tasks. The ack step in the go client prevents the worker from executing the task, and was therefore removed.

Authentication was added due to my own setup, where workers need to auth to perform any action on the api. The existing http client only took url, and so would not allow any auth.

### DIFF
--- a/polyglot-clients/go/conductorhttpclient.go
+++ b/polyglot-clients/go/conductorhttpclient.go
@@ -32,6 +32,22 @@ func NewConductorHttpClient(baseUrl string) *ConductorHttpClient {
     return conductorClient
 }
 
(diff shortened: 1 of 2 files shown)


CHURLZ:
@aravindanr can we get this through? Current go client is not compatible with v3.

PR #2343, opened 2021-07-06
feat(WT-23007): support es7 in nfx v3

vukor:
Pull Request type
----

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Other (please describe):

Changes in this PR
----

This pr is copied from https://github.com/Netflix/conductor/pull/2324

Issue #2323, opened 2021-06-18
[BUG]: Workflow Status is always in RUNNING if workflow output is greater than 10MB.


vamsibandarupalli:
**Describe the bug**
If the Workflow output is greater than 10MB, conductor is failing the workflow [here](https://github.com/Netflix/conductor/blob/c09ba129c2e57061902831efa06f0abac1903fc7/core/src/main/java/com/netflix/conductor/core/utils/ExternalPayloadStorageUtils.java#L135).

While terminating the workflow [here](https://github.com/Netflix/conductor/blob/c09ba129c2e57061902831efa06f0abac1903fc7/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java#L808) again condcutor is trying to update the workflow output which again throws termination exception and finally conductor just releasing the locks without updating the updated workflow into data store.

Also workflowStatusListner onTerminateWorkflow() is also not getting called.

**Details**
Conductor version: latest

**To Reproduce**
Add any workflow whose output should be greater than 10MB.

**Expected behavior**
Expected to update the workflow in data store, and also should triger workflowStatusListner (onTerminateWorkflow()) method.

kishorebanala:
@vamsibandarupalli Thanks for reporting this. If you happen to fix this yourselves, please send a PR our way.

vamsibandarupalli:
I'm seeing same issue for task as well, workflow status is always RUNNING when task output is greater than 10MB.

MikeTangoEcho:
The 10MB is the default Max Threshold for Conductor to decide where it should store the payload (persistence layer, external storage or just discard it)
https://github.com/Netflix/conductor/blob/8b1c0259b6a8bbced97d6ae8115f13a11c8f5c5b/core/src/main/java/com/netflix/conductor/core/config/ConductorProperties.java#L184-L241

By default, Conductor use the DummyPayloadStorage, so you may have a NullPointerException when it tries to update the workflow by using the payload.

https://github.com/Netflix/conductor/blob/d32ad1622740221398b48589a44079febbcdc5fb/core/src/main/java/com/netflix/conductor/core/config/ConductorCoreConfiguration.java#L55-L60

As a quick fix, you can increase thresholds (in the example here with 20Mb the limit).
```
conductor.app.workflowInputPayloadSizeThreshold=20MB
conductor.app.workflowOutputPayloadSizeThreshold=20MB
conductor.app.taskInputPayloadSizeThreshold=20MB
[...]

PR #2491, opened 2021-09-28
Fix inconsistent workflow status caused by upload payload exception d…

jxu-nflx:
…uring workflow termination

Pull Request type
----

- [x ] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----
Catch exception of external storage upload failures during workflow termination and continue the execution to persist workflow status, etc.
Issue #2343 https://github.com/Netflix/conductor/issues/2323

Alternatives considered
----

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
@@ -837,7 +837,14 @@ public class WorkflowExecutor {
             if (workflow.getWorkflowDefinition() == null) {
                 workflow = metadataMapperService.populateWorkflowWithDefinitions(workflow);
             }
-            deciderService.updateWorkflowOutput(workflow, null);
+
+            try {
+                deciderService.updateWorkflowOutput(workflow, null);
+            } catch (Exception e) {
+                // catch any failure in this step and continue the execution of terminating workflow
+                LOGGER.error("Failed to update output data for workflow: {}", workflow.getWorkflowId(), e);
+                Monitors.error(CLASS_NAME, "terminateWorkflow");
+            }
 
             // update the failed reference task names
             workflow.getFailedReferenceTaskNames().addAll(workflow.getTasks().stream()
--- a/core/src/main/java/com/netflix/conductor/core/utils/ExternalPayloadStorageUtils.java
+++ b/core/src/main/java/com/netflix/conductor/core/utils/ExternalPayloadStorageUtils.java
@@ -132,7 +132,7 @@ public class ExternalPayloadStorageUtils {
                     String errorMsg = String.format(
                         "The output payload size: %dB of workflow: %s is greater than the permissible limit: %dKB",
                         payloadSize, ((Workflow) entity).getWorkflowId(), maxThreshold);
-                    failWorkflow(errorMsg);
+                    failWorkflow(((Workflow) entity), payloadType, errorMsg);
                 }
             } else if (payloadSize > threshold * 1024) {
                 switch (payloadType) {
@@ -194,8 +194,16 @@ public class ExternalPayloadStorageUtils {
         throw new TerminateWorkflowException(errorMsg, Workflow.WorkflowStatus.FAILED, task);
     }
 
-    private void failWorkflow(String errorMsg) {
+    @VisibleForTesting
+    void failWorkflow(Workflow workflow, PayloadType payloadType, String errorMsg) {
         LOGGER.error(errorMsg);
-        throw new TerminateWorkflowException(errorMsg);
+        workflow.setReasonForIncompletion(errorMsg);
+        workflow.setStatus(Workflow.WorkflowStatus.FAILED);
+        if (payloadType == PayloadType.TASK_INPUT) {
+            workflow.setInput(null);
+        } else {
+            workflow.setOutput(null);
+        }
+        throw new TerminateWorkflowException(errorMsg, Workflow.WorkflowStatus.FAILED);
     }
 }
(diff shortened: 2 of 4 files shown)


core/src/main/java/com/netflix/conductor/core/utils/ExternalPayloadStorageUtils.java
```diff
@@ -194,8 +194,16 @@ void failTask(Task task, PayloadType payloadType, String errorMsg) {
         throw new TerminateWorkflowException(errorMsg, Workflow.WorkflowStatus.FAILED, task);
     }
 
-    private void failWorkflow(String errorMsg) {
+    @VisibleForTesting
+    void failWorkflow(Workflow workflow, PayloadType payloadType, String errorMsg) {
         LOGGER.error(errorMsg);
-        throw new TerminateWorkflowException(errorMsg);
+        workflow.setReasonForIncompletion(errorMsg);
```
aravindanr, line 200: Lines L200- L206 does not look necessary as the catch block in [WorkflowExecutor.decide](https://github.com/Netflix/conductor/blob/main/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java#L1162-L1165) and [WorkflowExecutor.terminateWorkflow](https://github.com/Netflix/conductor/blob/main/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java#L849) set the status and reasonForIncompletion

core/src/main/java/com/netflix/conductor/core/utils/ExternalPayloadStorageUtils.java
```diff
@@ -194,8 +194,16 @@ void failTask(Task task, PayloadType payloadType, String errorMsg) {
...
+        workflow.setStatus(Workflow.WorkflowStatus.FAILED);
+        if (payloadType == PayloadType.TASK_INPUT) {
```
aravindanr, line 202: From where the `failWorkflow` is called, the `PayloadType` should always be `WORKFLOW_INPUT` or `WORKFLOW_OUTPUT`.

aravindanr (changes requested):

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -837,7 +837,14 @@ public Workflow terminateWorkflow(Workflow workflow, String reason, String failu
             if (workflow.getWorkflowDefinition() == null) {
                 workflow = metadataMapperService.populateWorkflowWithDefinitions(workflow);
             }
-            deciderService.updateWorkflowOutput(workflow, null);
+
+            try {
+                deciderService.updateWorkflowOutput(workflow, null);
+            } catch (Exception e) {
```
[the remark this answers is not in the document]
jxu-nflx: The `verifyAndUpload` method could also throw `ApplicationException`: https://github.com/Netflix/conductor/blob/main/core/src/main/java/com/netflix/conductor/core/utils/ExternalPayloadStorageUtils.java#L87 So I use the generic catch of all exceptions here as I feel no matter what exception we got at this step, it shouldn't prevent us from continuing the termination execution

core/src/main/java/com/netflix/conductor/core/utils/ExternalPayloadStorageUtils.java
jxu-nflx: I was going to remove that. But then on a second thought, I think if we get any exception from uploading outputdata to external storage during the terminate step, the final status should be `failed` instead of `terminated` with specific error message reflects that. What do you think?

core/src/main/java/com/netflix/conductor/core/utils/ExternalPayloadStorageUtils.java
jxu-nflx: yeah the `payloadType` is passed into the `verifyAndUpload` method: https://github.com/Netflix/conductor/blob/main/core/src/main/java/com/netflix/conductor/core/utils/ExternalPayloadStorageUtils.java#L87 And we always verify both input and output data: https://github.com/Netflix/conductor/blob/main/core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java#L566

core/src/main/java/com/netflix/conductor/core/utils/ExternalPayloadStorageUtils.java
aravindanr: I think the final status should be `TERMINATED`. A workflow is marked as FAILED, because of a Task failure or it timing out, not because of internal errors during the workflow evaluation. The `reasonForCompletion` should also be the original error that triggered the termination, not the internal error.

core/src/main/java/com/netflix/conductor/core/utils/ExternalPayloadStorageUtils.java
aravindanr: Yes. `ExternalPayloadStorageUtils.verifyAndUpload()` is never called with a `Workflow` object and `PayloadType` as `TASK_INPUT` or `TASK_OUTPUT`. Workflow's input/output should not be manipulated here since the caller might be already manipulating it.

For instance, if the workflow definition has a `failure_workflow` defined, the workflow output is manipulated and setting it to `null` here would cause a `NullPointerException` there.

Long story short, i don't see a reason to update the `failWorkflow` method to fix this issue.

aravindanr (changes requested):

core/src/main/java/com/netflix/conductor/core/utils/ExternalPayloadStorageUtils.java
jxu-nflx: I added the logic to setting workflow output or input to `null` in the `failWorkflow` method because in normal circustances (i.e. when input or output size don't exceed threshold), we always do set the input/output to `null`: https://github.com/Netflix/conductor/blob/main/core/src/main/java/com/netflix/conductor/core/utils/ExternalPayloadStorageUtils.java#L162 So to be consistent, I think it makes sense to have the same logic added for `failWorkflow`: https://github.com/Netflix/conductor/blob/main/core/src/main/java/com/netflix/conductor/core/utils/ExternalPayloadStorageUtils.java#L135

core/src/main/java/com/netflix/conductor/core/utils/ExternalPayloadStorageUtils.java
jxu-nflx: Sure ya updated

PR #2494, opened 2021-09-28
Fix rerunWorkflow places synchronous system tasks in the queue

jxu-nflx:
Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----
When rerun from task, if it's a synchronous system task, don't put it in the queue, start directly instead.

Issue #MWI-3456: rerunWorkflow places synchronous system tasks in the queue

Alternatives considered
----

### DIFF
--- a/contribs/src/main/java/com/netflix/conductor/contribs/tasks/json/JsonJqTransform.java
+++ b/contribs/src/main/java/com/netflix/conductor/contribs/tasks/json/JsonJqTransform.java
@@ -30,6 +30,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
@@ -58,7 +59,7 @@ public class JsonJqTransform extends WorkflowSystemTask {
     @Override
     public void start(Workflow workflow, Task task, WorkflowExecutor executor) {
         final Map<String, Object> taskInput = task.getInputData();
-        final Map<String, Object> taskOutput = task.getOutputData();
+        final Map<String, Object> taskOutput = task.getOutputData() != null ? task.getOutputData(): new HashMap<>();
 
         final String queryExpression = (String) taskInput.get(QUERY_EXPRESSION_PARAMETER);
 
@@ -87,6 +88,7 @@ public class JsonJqTransform extends WorkflowSystemTask {
                 taskOutput.put(OUTPUT_RESULT, result.get(0));
                 taskOutput.put(OUTPUT_RESULT_LIST, result);
             }
+            task.setOutputData(taskOutput);
         } catch (final Exception e) {
             LOGGER.error("Error executing task: {} in workflow: {}", task.getTaskId(), workflow.getWorkflowId(), e);
             task.setStatus(Task.Status.FAILED);
--- a/contribs/src/main/java/com/netflix/conductor/contribs/tasks/json/JsonJqTransform.java
+++ b/contribs/src/main/java/com/netflix/conductor/contribs/tasks/json/JsonJqTransform.java
@@ -30,7 +30,6 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
@@ -59,7 +58,7 @@ public class JsonJqTransform extends WorkflowSystemTask {
     @Override
     public void start(Workflow workflow, Task task, WorkflowExecutor executor) {
         final Map<String, Object> taskInput = task.getInputData();
-        final Map<String, Object> taskOutput = task.getOutputData() != null ? task.getOutputData(): new HashMap<>();
+        final Map<String, Object> taskOutput = task.getOutputData();
 
         final String queryExpression = (String) taskInput.get(QUERY_EXPRESSION_PARAMETER);
 
@@ -88,7 +87,6 @@ public class JsonJqTransform extends WorkflowSystemTask {
                 taskOutput.put(OUTPUT_RESULT, result.get(0));
                 taskOutput.put(OUTPUT_RESULT_LIST, result);
             }
-            task.setOutputData(taskOutput);
         } catch (final Exception e) {
             LOGGER.error("Error executing task: {} in workflow: {}", task.getTaskId(), workflow.getWorkflowId(), e);
             task.setStatus(Task.Status.FAILED);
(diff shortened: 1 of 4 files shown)


contribs/src/main/java/com/netflix/conductor/contribs/tasks/json/JsonJqTransform.java
```diff
@@ -58,7 +59,7 @@ public JsonJqTransform(ObjectMapper objectMapper) {
     @Override
     public void start(Workflow workflow, Task task, WorkflowExecutor executor) {
         final Map<String, Object> taskInput = task.getInputData();
-        final Map<String, Object> taskOutput = task.getOutputData();
+        final Map<String, Object> taskOutput = task.getOutputData() != null ? task.getOutputData(): new HashMap<>();
```
aravindanr, line 62: `getOutputData()` does not return `null`. Can you please explain the need for a null check here?

core/src/test/java/com/netflix/conductor/core/execution/TestWorkflowExecutor.java
```diff
@@ -1324,6 +1340,59 @@ public void testRerunWorkflowWithTaskId() {
         assertEquals(new HashSet<>(), workflow.getFailedReferenceTaskNames());
     }
 
+    @Test
+    public void testRerunWorkflowWithSyncSystemTaskId() {
```
aravindanr, line 1344: 👍🏼

test-harness/src/test/groovy/com/netflix/conductor/test/integration/JsonJQTransformSpec.groovy
```diff
@@ -84,4 +85,66 @@ class JsonJQTransformSpec extends AbstractSpecification {
...
+        def workflowInstanceId = workflowExecutor.startWorkflow(JSON_JQ_TRANSFORM_WF, 1,
+                '', invalidInput, null, null, null)
+
+        then: "verify that the workflow and task failed with expected error"
+        with(workflowExecutionService.getExecutionStatus(workflowInstanceId, true)) {
+            status == Workflow.WorkflowStatus.FAILED
+            tasks.size() == 1
+            tasks[0].status == Task.Status.FAILED
+            tasks[0].taskType == 'JSON_JQ_TRANSFORM'
+            tasks[0].reasonForIncompletion as String == "Cannot index string with string \"array\""
```
aravindanr, line 129: The `as String` may not be necessary. Also, using single quotes `'` ensures that Groovy uses a `String` and not a `GString`.

test-harness/src/test/groovy/com/netflix/conductor/test/integration/JsonJQTransformSpec.groovy
```diff
@@ -84,4 +85,66 @@ class JsonJQTransformSpec extends AbstractSpecification {
...
+
+        workflowExecutor.rerun(reRunWorkflowRequest)
+
+        then: "verify that the workflow and task are completed with expected output"
+        with(workflowExecutionService.getExecutionStatus(workflowInstanceId, true)) {
+            status == Workflow.WorkflowStatus.COMPLETED
+            tasks.size() == 1
+            tasks[0].status == Task.Status.COMPLETED
+            tasks[0].taskType == 'JSON_JQ_TRANSFORM'
+            tasks[0].outputData as String == "[result:[out:[a, b, c, d]], resultList:[[out:[a, b, c, d]]]]"
```
aravindanr, line 147: same as above

contribs/src/main/java/com/netflix/conductor/contribs/tasks/json/JsonJqTransform.java
jxu-nflx: The outputData field is reset to null here: https://github.com/Netflix/conductor/blob/main/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java#L1642 I guess we could update that line to just clear the map instead of setting it null, but feel there might be other cases that I might neglected if I changed that. So I added the null check for outputData in the system task instead. And that's the only sync system task that need to add the null check

test-harness/src/test/groovy/com/netflix/conductor/test/integration/JsonJQTransformSpec.groovy
jxu-nflx: Thanks, updated!

test-harness/src/test/groovy/com/netflix/conductor/test/integration/JsonJQTransformSpec.groovy
jxu-nflx: updated as well, the `as String` is necessary for outputData though as it's a map, so kept it, but used single quotes for the match string instead

aravindanr: I missed that it's a map here. In that case, it makes sense to assert the keys of the map rather than converting it to s String and asserting that. The assertion will fail, if the `toString()` logic in Map changes, its unlikely, but possible.

jxu-nflx: I think we can rely on Java's implementation of `toSTring()` of the Map. It's kind of cumbersome to write a loop through the map to assert each key/value pair. What do you think?

contribs/src/main/java/com/netflix/conductor/contribs/tasks/json/JsonJqTransform.java
aravindanr: It makes sense to clear the `outputData` or set it to an empty `HashMap` during `rerun`. If you look at the [Task](https://github.com/Netflix/conductor/blob/main/common/src/main/java/com/netflix/conductor/common/metadata/tasks/Task.java#L158), the outputData field is defaulted to an empty map. So it's natural for the rest of the code to expect a non-null value from `Task.getOutputData()`. `DoWhile` also [expects a non-null output data](https://github.com/Netflix/conductor/blob/main/core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java#L63) and would fail if null is returned.

Also, the fewer nulls we introduce, the less chance of an unexpected `NullPointerException`.

test-harness/src/test/groovy/com/netflix/conductor/test/integration/JsonJQTransformSpec.groovy
aravindanr: Verifying the contents of a Map using its `toString()` is not ideal. It's also harder to understand and refactor. The goal of the test is not to verify if `jackson-jq` is doing its work, so you could 

1. write a simpler `jq` expression that is easy to verify or 
2. remove the `outputData` assertion altogether or
3. assert that `output.result != null` and `output.resultList != null`

contribs/src/main/java/com/netflix/conductor/contribs/tasks/json/JsonJqTransform.java
jxu-nflx: ya updated it to clear the `outputData` instead of setting to null

test-harness/src/test/groovy/com/netflix/conductor/test/integration/JsonJQTransformSpec.groovy
jxu-nflx: Make sense, didn't realize the outputData result is JsonNode. yea, the purpose of the test isn't for those kind of checks. Refactored to only do basic assertions.

PR #2499, opened 2021-10-01
Empty variable is defaulted to empty string

aravindanr:
Errors during variable replacement can fail the decide process and leave the workflow in a RUNNING state indefinitely

Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----
ParametersUtils defaults `${}` or `${   }` to empty string.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/utils/ParametersUtils.java
+++ b/core/src/main/java/com/netflix/conductor/core/utils/ParametersUtils.java
@@ -25,6 +25,7 @@ import com.netflix.conductor.common.metadata.workflow.WorkflowDef;
 import com.netflix.conductor.common.run.Workflow;
 import com.netflix.conductor.common.utils.EnvUtils;
 import com.netflix.conductor.common.utils.TaskUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/netflix/conductor/core/utils/ParametersUtils.java
```diff
@@ -203,12 +204,18 @@ private Object replaceList(List<?> values, String taskId, DocumentContext io) {
     }
 
     private Object replaceVariables(String paramString, DocumentContext documentContext, String taskId) {
-        String[] values = paramString.split("(?=(?<!\\$)\\$\\{)|(?<=\\})");
+        String[] values = paramString.split("(?=(?<!\\$)\\$\\{)|(?<=})");
```
jxu-nflx, line 207: Could we add a comment here for the regex?

jxu-nflx (approved):

core/src/main/java/com/netflix/conductor/core/utils/ParametersUtils.java
aravindanr: That would have been helpful. The original author did not include a comment, i merely removed the redundant escape character.

PR #2503, opened 2021-10-07
Remove dependency on Netflix Servo

taoj-action:
Netflix Servo is a deprecated library. Conductor's dependency on Servo can
be simply removed due to code related to Servo Stopwatch is not used anywhere.

Also need to explicitly add Guava after removing servo dependency.

Build and test pass except for cassandraDaoTest during test teardown for cleanupData.

We will investigate more on explicitly adding guava dependency causing test
failure.

Signed-off-by: Tao Jiang <taoj@vmware.com>

Pull Request type
----

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----
Removed Netflix Servo dependency because related code is unused. This is a small step for cleaning up dependencies.

Issue # 
https://github.com/Netflix/conductor/discussions/2501

Alternatives considered
----
Directly make Monitors from class to interface. 

_Describe alternative implementation you have considered

In order to make Monitors class to Interface need to create an almost identical amount to work already exists in spectator-api which may not worthwhile to do so.

### DIFF
--- a/azureblob-storage/dependencies.lock
+++ b/azureblob-storage/dependencies.lock
@@ -57,6 +57,12 @@
             ],
             "locked": "1.0.0"
         },
+        "com.google.guava:guava": {
+            "firstLevelTransitive": [
+                "com.netflix.conductor:conductor-core"
+            ],
+            "locked": "30.0-jre"
+        },
         "com.google.protobuf:protobuf-java": {
             "firstLevelTransitive": [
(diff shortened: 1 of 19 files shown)


aravindanr:
@taojwmware can you execute the following gradle command and commit the changes?

`./gradlew generateLock saveLock`

PR #2512, opened 2021-10-13
[FEATURE] Add field to update task by reference name

jasondavindev:
Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Changes in this PR
----

Adds a new field to update task by reference name.

Alternatives considered
----

The below case

Untitled Diagram (2)

For now is impossible pass a task id to Kafka for it updates a WAIT task using `POST /api/tasks`. For solve it this PR adds a new field to enable update a task by reference name.

Notes:
- I'm using an external app with Kafka for avoid application polling
- There's the possibility to make a request to `GET /api/workflows/{id}` for retrieve all tasks, but it is expensive

### DIFF
--- a/common/src/main/java/com/netflix/conductor/common/metadata/tasks/TaskResult.java
+++ b/common/src/main/java/com/netflix/conductor/common/metadata/tasks/TaskResult.java
@@ -61,6 +61,8 @@ public class TaskResult {
     @ProtoField(id = 8)
     private Any outputMessage;
 
+    private String taskReferenceName;
+
     private List<TaskExecLog> logs = new CopyOnWriteArrayList<>();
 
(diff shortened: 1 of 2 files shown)


apanicker-nflx:
Updating the task by taskReferenceName is not desirable because different attempts for a given task (possibly being executed in different workers) will have the same task reference name but are guaranteed to have different task ids. Allowing updating task only by id prevents incorrect updates from workers.

I would like to reiterate our [contributing guidelines](https://github.com/Netflix/conductor/blob/main/CONTRIBUTING.md#i-want-to-contribute) which would have helped save your cycles in this case. Please create a discussion to discuss the issue you are facing before starting a PR. Thanks.

PR #2566, opened 2021-11-09
Fixes #2565 - Fix the broken inline task implementation. Added unit tests for Inline task

### LINKED ISSUE #2565: Inline tasks implementation is broken - the validations are incorrectly implemented
**Describe the bug**
Inline tasks implementation is broken - the validations are incorrectly implemented. The validations intends to check for whether the `evaluationType` or `expression` is blank, but instead its checking for `isNotBlank`. There are no tests for the new inline task implementation either.

**Details**
Conductor version: Latest
Persistence implementation: NA
Queue implementation: NA
Lock: NA
Workflow definition: NA
Task definition: NA
Event handler definition: NA

**To Reproduce**
1. Implement an INLINE task and try to run it
2. Fails with the validation error even if the required params are missing

**Expected behavior**
Validations should be implemented correctly

**Screenshots**
NA

**Additional context**
NA

boney9:
Fix the broken inline task implementation. Added unit tests to validate for both value param and javascript evaluators

Pull Request type
----

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----
Fixes the Inline task validation implementation. Without this fix, inline task won't work.

Issue #2565

Alternatives considered
----
None

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/tasks/Inline.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/tasks/Inline.java
@@ -94,7 +94,7 @@ public class Inline extends WorkflowSystemTask {
     }
 
     private void checkEvaluatorType(String evaluatorType) {
-        if (StringUtils.isNotBlank(evaluatorType)) {
+        if (StringUtils.isBlank(evaluatorType)) {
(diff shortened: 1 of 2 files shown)


nvn07 (approved):
LGTM

boney9:
@aravindanr @peterlau - We found this bug and fixed it. FYI.

Issue #2548, opened 2021-10-29
[FEATURE]: [UI] Adjustable width task panel


peterlau:
Please read our [contributor guide](https://github.com/Netflix/conductor/blob/main/CONTRIBUTING.md) before creating an issue.   
Also consider discussing your idea on the [discussion forum](https://github.com/Netflix/conductor/discussions) first.

## Describe the Feature Request
The task panel is currently fixed width. While this works well in most cases, on small screens it would be advantageous to be able to expand the panel.

## Describe Preferred Solution
List of potential improvements in order of ease of implementation:
* Add a vertical slider handle to the existing panel so that the task panel width can be interactively adjusted
* Provide a toggle so that the panel can be 'maximized' most of the way and retracted back to the 'split' width.
* The panel is currently 'docked'. That is, as the task panel expands, the main workflow panel shrinks. Provide a toggle so that users may optionally 'undock' the task panel such that it overlays over the workflow panel.

## Describe Alternatives
Separate task details page. This would require flipping back and forth between the workflow view and tasks disrupting the user experience and making it impossible to browse through a workflow execution smoothly.

boney9:
@peterlau - As we discussed - we can own this task. You can assign it to us.

PR #2572, opened 2021-11-12
[UI] Task json fix: Removed start and finish task from json

debapi94:
Pull Request type
----

- [x ] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
image


ui/src/pages/execution/Execution.jsx
```diff
@@ -22,13 +22,16 @@ import clsx from "clsx";
...
 import RightPanel from "./RightPanel";
 import WorkflowDAG from "../../components/diagram/WorkflowDAG";
 import StatusBadge from "../../components/StatusBadge";
 import { useFetch } from "../../utils/query";
 import { Helmet } from "react-helmet";
 
-const drawerWidth = 650;
+const maxWindowWidth = window.innerWidth;
+const drawerWidth = maxWindowWidth>1000 ? maxWindowWidth/2 : 450;
+console.log(drawerWidth);
```
NickTomlin, line 34: is this log intentional?

ui/src/pages/execution/Execution.jsx
```diff
@@ -110,17 +112,90 @@ const useStyles = makeStyles({
...
+  MOUSE_MOVE: 5,
+}
+
+const initialState = {
+  drawerWidth: drawerWidth,
+  isDrawerFullScreen: false,
+  isResizing: false,
+  lastDownX: 0,
+  newWidth: {},
+  //drawerVarient: drawerWidth <= 450 ? 'temporary' : 'persistent'
```
NickTomlin, line 143: do we want this comment here?

ui/src/pages/execution/Execution.jsx
```diff
@@ -110,17 +112,90 @@ const useStyles = makeStyles({
...
+    dispatch({type:actions.MOUSE_DOWN, clientX:e.clientX});
+  };
+
+  const handleMousemove = useCallback((e) => {
+    // we don't want to do anything if we aren't resizing.
+    if (!state.isResizing) {
+      return;
+    }
+
+    let offsetRight =
```
NickTomlin, line 187: nit, since we don't seem to be mutating these variables, would `const` suffice?

ui/src/pages/execution/Execution.jsx
```diff
@@ -213,26 +309,37 @@ export default function Execution() {
...
-          </IconButton>
-        </div>
-        <Divider />
-        <RightPanel
-          className={classes.drawerContent}
-          selectedTask={selectedTask}
-          dag={dag}
-          onTaskChange={handleSelectedTask}
-        />
+        <div
```
NickTomlin, line 321: nit: indentation seems a little off here. 

Side note: :if we don't have `prettier` for the project we should think about implementing (maybe in a git hook of some sort)

ui/src/components/ReactJson.jsx
```diff
@@ -15,6 +15,9 @@ const useStyles = makeStyles({
 export default function ReactJson({ title, className, style, ...props }) {
   const [clipboardMode, setClipboardMode] = useState(false);
   const classes = useStyles();
+  
+  let modifiedProps = props && {src:{...props.src, tasks:props.src.tasks.filter(t => !["__start", "__final"].includes(t.taskReferenceName))}};
```
NickTomlin, line 19: nit: there's a lot going on here and it might be clearer broken out into a helper function or a few lines.

NickTomlin (commented):
made some minor nits/stylistic comments. Thanks for adding this; I think the adjustable sider will also be really useful.

ui/src/components/ReactJson.jsx
debapi94: thank you, created a separate method to take care of task filtering

ui/src/pages/execution/Execution.jsx
debapi94: fixed in other PR #2548

ui/src/pages/execution/Execution.jsx
debapi94: fixed in other PR #2548

ui/src/pages/execution/Execution.jsx
debapi94: fixed in other PR #2548

peterlau:
This behavior looks to be caused by erroneous mutation of the workflow tasks array in constructing the workflow DAG. Rather than filter it out in the `ReactJson.jsx`, which is a generic display component that we should leave agnostic to the structure of the JSON being passed in, I recommend the following simple fix:

**WorkflowDAG.js:62**
```
    const definedTasks = [...workflowDef.tasks];
 ```

peterlau (changes requested):
See comment in main PR thread regarding alternative approach.

peterlau (changes requested):
Could we pull from master and eliminate the diff in `Execution.jsx`? The handle slider PR has been merged so there should not be any changes here. Thanks

debapi94:
Let me close this PR and create a new one

Issue #2573, opened 2021-11-14
3.3.4 UI Bug.   Task Details tab doesn`t show with javascript error.


vfmuniz:
Hi. I just upgraded to 3.3.4 UI and I noticed that task details tab does not show when you click on it any task. Issue happens both on production mode (when I copy the build result from yarn build command to the server resources folder and access all the pages as static files) and on development mode.

Details
Conductor version:
Persistence implementation: Oracle, Elastic 7
Queue implementation: Dynoqueues, RabbitMQ etc
Lock: Redis or Zookeeper? NO lockin
Workflow definition: Any
Task definition: Any
Event handler definition: Any

To Reproduce

Go to UI homescreen
Click on Search button
Scroll down to any workflow that is COMPLETE or RUNNING or any other state
Once the workflow details screen opens, click on any Task

Expected behavior
The task details tab should open so user can see Input, Output, Logs and so on. Instead, a javascript error happens. 

Screenshot:
![image](https://user-images.githubusercontent.com/32685867/141684095-babc77c5-5568-40f5-a755-579bf17055c7.png)

peterlau:
Looks to be caused by a workflow execution returning no 'taskToDomain' field.

vfmuniz:
Thank you :)

PR #2582, opened 2021-11-16
Bug fix 2573

debapi94:
Pull Request type
----

- [x ] Bugfix
- [ ] Feature
- [ ] Build related changes

#2573


peterlau:
After a discussion with @aravindanr it appears this entire block of code is entirely unnecessary as `domain` is evaluated as part of the `tasks` array where it is available and we should just use the value as-is returned by the API. Could you try if the following works?
Could you try if the following works?


```
  const handleSelectedTask = (task) => {
        setSelectedTask(null);
  }
```

ui/src/pages/execution/Execution.jsx
```diff
@@ -226,9 +226,14 @@ export default function Execution() {
     if (task) {
```
peterlau, line 226: Suggestion: Eliminate all logic in `handleSelectedTask`. `task.domain` should be automatically populated by API already.

debapi94:
Opend a new PR as this one got mixed with changes from other branch.

PR #2610, opened 2021-11-30
added nats stream support for WAIT task queues

astelmashenko:
Pull Request type
----

- [x] Feature

Changes in this PR
----
Added support of nats streaming server for WAIT task. It is similar to SQS implementation.

### DIFF
--- a/contribs/src/main/java/com/netflix/conductor/contribs/queue/nats/config/NATSStreamConfiguration.java
+++ b/contribs/src/main/java/com/netflix/conductor/contribs/queue/nats/config/NATSStreamConfiguration.java
@@ -37,7 +37,7 @@ public class NATSStreamConfiguration {
         return new NATSStreamEventQueueProvider(properties, scheduler);
     }
 
-    @ConditionalOnProperty(name = "conductor.default-event-queue.type", havingValue = "nats_stream", matchIfMissing = true)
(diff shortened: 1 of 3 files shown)


contribs/src/main/java/com/netflix/conductor/contribs/queue/nats/config/NATSStreamConfiguration.java
```diff
@@ -28,4 +36,33 @@
     public EventQueueProvider natsEventQueueProvider(NATSStreamProperties properties, Scheduler scheduler) {
         return new NATSStreamEventQueueProvider(properties, scheduler);
     }
+
+    @ConditionalOnProperty(name = "conductor.default-event-queue.type", havingValue = "nats_stream", matchIfMissing = true)
```
apanicker-nflx, line 40: Please remove the `match_if_missing=true` part. The default value is set to `sqs` and we wouldn't want to change that.

apanicker-nflx (changes requested):
Please also add an entry to https://github.com/Netflix/conductor/blob/main/contribs/src/main/resources/META-INF/additional-spring-configuration-metadata.json#L98

contribs/src/main/java/com/netflix/conductor/contribs/queue/nats/config/NATSStreamConfiguration.java
astelmashenko: done

PR #2612, opened 2021-11-30
Performance improvement by replacing StringBuffer with StringBuilder

larrydiamond:
Pull Request type

- [x ] Refactoring (no functional changes, no api changes)

Changes in this PR
----
Replacing StringBuffer objects used locally with StringBuilder.   StringBuilder does not synchronize and so will run faster.
This PR only improves performance


jxu-nflx (approved):
Thanks for fixing it!

Issue #2564, opened 2021-11-09
Creating a new workflow definition will not populate createTime


buzz08:
**Describe the bug**
When you create a new workflow definition, it doesn't set the create time. This causes issues in the workflow definitions UI page, especially if you try to sort by create time. We need to set this value as a current time when you create the workflow.

**Details**
Conductor version: Latest
Persistence implementation: NA
Queue implementation: NA
Lock:  NA
Workflow definition: Applicable to new workflows
Task definition: NA
Event handler definition: NA

**To Reproduce**
1. Create a new workflow without createTime or updateTime
2. Retrieve the workflow definition, it won't have createTime

**Expected behavior**
createTime should be defaulted to the system time

**Screenshots**
NA

**Additional context**
NA

jxu-nflx:
Hello @buzz08 Do you want to submit a PR to fix this? Thank you!

aravindanr:
I was able to reproduce it. PUT /workflow does not set the `createTime`, but POST /workflow defaults it to the current server time.

apanicker-nflx:
Since we are altering the metadata APIs, we propose that we should make these consistent. 
For example, the task definition APIs allow creation of an array of task definitions but update of one task definition. Further, the update operation is only allowed if the task definition has already been registered.

The proposal is to follow the same convention for workflow definition APIs as well. 
As part of the PR #2617, each update operation is now a read-before-write (getWorkflowDef before update). Further, the create API allows only one workflow definition but the update API allows an array of workflow definitions. Here again, the update workflow definition should only be allowed if the workflow definition has already been registered with the system.

API consistency across both workflow and task definitions would provide for a better user/developer experience and reduce support burden.

PR #2617, opened 2021-12-01
Bug fixed: createTime was null when updating worflow definition

villat:
Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----

Now we are getting the `createTime` from:

- The workflow definition if we want to update it
- The system if the workflow definitions doesn't exists yet

Issue #2564 

Alternatives considered
----

It isn't necessary to pass neither the `createTime` nor the `updateTime` to the POST and PUT **/metadata/workflow** methods


boney9:
@aravindanr @apanicker-nflx - This is from us - just fyi.

apanicker-nflx (dismissed):

apanicker-nflx (changes requested):
See [comment](https://github.com/Netflix/conductor/issues/2564#issuecomment-987317849) in #2564

PR #2631, opened 2021-12-08
[Feature] Support to specify shutdown grace period time for worker

howaric:
Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Changes in this PR
----
Add a new parameter 'shutdownGracePeriodSeconds' to TaskRunnerConfigurer which is the timeout of terminating worker thread pool. Default value is 10 seconds, and can't be changed.
A task may cost longer, so it is better to make users can specify the timeout.
Default value is still 10 seconds with this new feature.

Concern: 
I have removed shutdown() method from TaskPollExecutor, it is a default method, and not used by anywhere in this project, I can't see any reason to keep it, please correct me if there is potential problems, thanks.

Alternatives considered
----

### DIFF
--- a/client-spring/src/main/java/com/netflix/conductor/client/spring/ClientProperties.java
+++ b/client-spring/src/main/java/com/netflix/conductor/client/spring/ClientProperties.java
@@ -33,6 +33,8 @@ public class ClientProperties {
 
     private Map<String, String> taskToDomain = new HashMap<>();
 
+    private int shutdownGracePeriodSeconds = 10;
+
     public String getRootUri() {
         return rootUri;
     }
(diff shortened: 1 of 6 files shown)


aravindanr:
@howaric Thanks for the pull request. Could you also update the documentation in the `docs` folder? https://netflix.github.io/conductor/gettingstarted/client/#taskrunnerconfigurer

aravindanr (changes requested):
@howaric Thanks for the pull request. Could you also update the documentation in the docs folder? https://netflix.github.io/conductor/gettingstarted/client/#taskrunnerconfigurer

apanicker-nflx (approved):

aravindanr (approved):

PR #2630, opened 2021-12-08
[Fix] UI: fix fetching task logs uri

howaric:
Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Other (please describe):

Changes in this PR
----
Issue: '/api' is already contained in root uri, UI can't fetch task logs right now.
Solution: remove '/api' from fetch task logs uri.

Alternatives considered
----

PR #2640, opened 2021-12-14
Fix bug in Task->Logs view

marcocrasso:
Pull Request type
----

- [x] Bugfix

Changes in this PR
----

The Task->Logs view doesn't work, it tries to fetch resources from non existent URL. The problem was that it prefix the target endpoint with /api. Additionally, modified ignore to keep buildable artifacts out of configuration management.

With this fix:
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ buildscan.log
 # JS & UI Related
 node_modules
 /ui/build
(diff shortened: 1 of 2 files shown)


marcocrasso:
In favour of https://github.com/Netflix/conductor/pull/2630

PR #2699, opened 2022-01-14
Support for Retry with LINEAR_BACKOFF

venkag:
* Support for retry with linear backoff.
* Scale factor added to control the slope of the linear backoff.

Pull Request type
----

- [ ] Bugfix
- [ X] Feature

Changes in this PR
----

Support for Retry with LINEAR_BACKOFF
Issue #

Alternatives considered
----

No alternatives available. This support is added just to have a certain retry behavior.

### DIFF
--- a/grpc/src/main/proto/model/taskdef.proto
+++ b/grpc/src/main/proto/model/taskdef.proto
@@ -10,7 +10,8 @@ option go_package = "github.com/netflix/conductor/client/gogrpc/conductor/model"
 message TaskDef {
     enum RetryLogic {
         FIXED = 0;
-        EXPONENTIAL_BACKOFF = 1;
+        LINEAR_BACKOFF = 1;
+        EXPONENTIAL_BACKOFF = 2;
     }
     enum TimeoutPolicy {
         RETRY = 0;
@@ -35,4 +36,5 @@ message TaskDef {
     string execution_name_space = 17;
     string owner_email = 18;
     int32 poll_timeout_seconds = 19;
+    int32 backoff_rate = 20;
 }
--- a/grpc/src/main/proto/model/taskdef.proto
+++ b/grpc/src/main/proto/model/taskdef.proto
@@ -36,5 +36,5 @@ message TaskDef {
     string execution_name_space = 17;
     string owner_email = 18;
     int32 poll_timeout_seconds = 19;
-    int32 backoff_rate = 20;
+    int32 backoff_scale_factor = 20;
 }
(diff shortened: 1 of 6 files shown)


venkag:
I reran this again locally on my system and the build was successful. Is this a problem with then environment?

@apanicker-nflx @aravindanr @kishorebanala

apanicker-nflx:
@venkag We do not see a significant value gain by adding this retry logic feature, specifically since there isn't a requirement from the community. Could you please provide more details on why this needs to be added to Conductor and what specific usecases you are trying to address here?

Also, please note that as part of the [contributing guidelines](https://github.com/Netflix/conductor/blob/main/CONTRIBUTING.md#i-want-to-contribute), we recommend starting a [discussion](https://github.com/Netflix/conductor/discussions) to gather inputs from the community and gauge interest. This also helps to finalize the implementation details for a contribution.

venkag:
Thank you for the feedback @apanicker-nfl! I had a feeling that I was not following the SoP 😄.

Agree that this is not a feature that community did not agree yet. I was hoping that this is non-intrusive change and is complimentary (instead of conflicting with any feature), I assumed, it would be OK to raise a PR directly.

This is something that we need internally. We are trying to keep our internal mirror as close to open source version as possible and thats the reason for submitting the PR in the first place.

Please let me know on the next steps. I see `Convert to draft` link under `Reviewers` section. Should I convert this PR to a draft to start a discussion or should I start a fresh discussion?

Sorry for the confusion.

grpc/src/main/proto/model/taskdef.proto
```diff
@@ -10,7 +10,8 @@ option go_package = "github.com/netflix/conductor/client/gogrpc/conductor/model"
 message TaskDef {
     enum RetryLogic {
         FIXED = 0;
-        EXPONENTIAL_BACKOFF = 1;
+        LINEAR_BACKOFF = 1;
```
apanicker-nflx, line 13: changing the ordinal of the enum fields here would have unforeseen consequences for users using the grpc server and proto classes. A possible fix would be to append the new value at the end of the enum in TaskDef

apanicker-nflx (changes requested):

apanicker-nflx:
Thanks for the explanation and for the PR itself. Agreed that it is not a conflicting feature that breaks compatibility, but we still value community insights into new features to ensure that we cover the widest array of possible use cases. Thanks for starting the discussion as well.

grpc/src/main/proto/model/taskdef.proto
venkag: Adjusted the enum and the `taskdef.proto` is fixed (automatically) as well..

apanicker-nflx (approved):

PR #2702, opened 2022-01-17
Object model separation

apanicker-nflx:
Pull Request type
----
- [x] Other (please describe): 
Separate object models into Data Transfer Object and Domain Object

Changes in this PR
----
A new object model is introduced that will be used by the state machine and all internal processes within Conductor. All external facing APIs and client will continue to use the existing object model. A new class `ModelMapper` is introduced to provide the translation between the domain object and the data transfer object. 

The separation of domain object from the data transfer object completely decouples the conductor state machine from the user-facing APIs. Some internal fields that are only used by the state machine (e.g. executed, subWorkflowChanged etc) are currently exposed to the users because the same object model is used by the state machine and the DTO layer. This separation allows the evolution of the state machine and features therewith, without impacting the user. It also helps abstract away any of the state machine complexities while providing a clean and lean object model for the user to work with.

Additionally, because of the way the object models are intertwined currently, payloads stored in external storage need to be manipulated (uploaded/downloaded) depending on the call stack, either from state machine or api layer, so that large payloads are not sent over the wire and are only available for use within the state machine. The separation of object models enables the state machine to download the payload from external payload storage once during a given workflow evaluation without needing to examine the call stack. The API layer does not need to deal with the large payloads since the model mapper handles the mapping to the DTO.


indiv0:
What's the driving factor behind this change, if you don't mind my asking? DTO/DO separation tends to add a lot of complexity in my experience.

core/src/main/java/com/netflix/conductor/domain/WorkflowDO.java
```diff
@@ -0,0 +1,366 @@
...
+
+import org.apache.commons.lang3.StringUtils;
+
+import com.netflix.conductor.common.metadata.workflow.WorkflowDef;
+
+import com.google.common.base.Preconditions;
+
+import static com.netflix.conductor.domain.WorkflowStatusDO.RUNNING;
+
+public class WorkflowDO {
```
jxu-nflx, line 26: I am wondering should we use a different name convention for xxxDO? I understand it's short for domain object, but isn't that intuitive when first hearing it.

core/src/main/java/com/netflix/conductor/domain/WorkflowStatusDO.java
```diff
@@ -0,0 +1,44 @@
...
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package com.netflix.conductor.domain;
+
+import com.netflix.conductor.common.run.Workflow.WorkflowStatus;
+
+public enum WorkflowStatusDO {
```
jxu-nflx, line 17: I think we should abstract the status out into its own object model, but do we need to separate Status into DO, DTO objects? What used internally for the states machines should be the same for what exposed to users, right? My concern is about keep it consistently and considerate each other if separated. Same goes for TaskStatus

core/src/main/java/com/netflix/conductor/core/dal/DomainMapper.java
```diff
@@ -0,0 +1,213 @@
...
+        this.externalPayloadStorageUtils = externalPayloadStorageUtils;
+    }
+
+    /**
+     * Fetch the fully formed workflow domain object with complete payloads
+     *
+     * @param workflowDO the workflow domain object from the datastore
+     * @return the workflow domain object {@link WorkflowDO} with payloads from external storage
+     */
+    public WorkflowDO getWorkflowDO(WorkflowDO workflowDO) {
```
jxu-nflx, line 51: What about call it something like `populateDataToWorkflowDO` or as a contrary to the `getLeanCopy` method, maybe we can call this `getFatCopy/getDeepCopy`? :) The reason to bring this up is `getWorkflowDO` takes a `WorkflowDO` and then returns a `WorkflowDO` could be confusing for users

core/src/main/java/com/netflix/conductor/core/dal/DomainMapper.java
```diff
@@ -0,0 +1,213 @@
...
+        return workflowDO;
+    }
+
+    public WorkflowDO getLeanCopy(WorkflowDO workflowDO) {
+        WorkflowDO leanWorkflowDO = workflowDO.copy();
+        externalizeWorkflowData(leanWorkflowDO);
+        return leanWorkflowDO;
+    }
+
+    public Workflow getWorkflowDTO(WorkflowDO workflowDO) {
```
jxu-nflx, line 62: Should this just be named to be `getWorkflow` instead since the returned object don't have the DTO suffix, which of course we should have ideally, but since it cannot be done due to backward compatibility. I think we should just eliminate the appear of `DTO` to avoid confusions

core/src/main/java/com/netflix/conductor/domain/WorkflowDO.java
apanicker-nflx: changed to WorkflowModel

core/src/main/java/com/netflix/conductor/domain/WorkflowStatusDO.java
apanicker-nflx: nested the status object inside the model

core/src/main/java/com/netflix/conductor/core/dal/DomainMapper.java
apanicker-nflx: changed to getFullCopy

core/src/main/java/com/netflix/conductor/core/dal/DomainMapper.java
apanicker-nflx: changed

apanicker-nflx:
> What's the driving factor behind this change, if you don't mind my asking? DTO/DO separation tends to add a lot of complexity in my experience.

@indiv0 While it does add some complexity at the beginning, it completely decouples the conductor state machine from the user-facing APIs. Some of the internal fields used by the state machine (eg. executed, subWorkflowChanged etc) are currently exposed to the users because the same object model is used by the state machine and the DTO layer. This separation allows the evolution of the state machine and features therewith, without impacting the user.

build.gradle
```diff
@@ -184,3 +184,14 @@ configure(allprojects - project(':conductor-grpc')) {
         }
     }
 }
+
+["cassandra-persistence", "core", "redis-concurrency-limit", "test-harness"].each {
```
aravindanr, line 188: Should this configuration be done for all projects? We will end up adding Spock tests for most of the modules.

aravindanr (approved):

build.gradle
apanicker-nflx: we can extend this to all java projects, i only added the ones that currently have spock tests

jxu-nflx (approved):

PR #2760, opened 2022-02-08
Reset workflow index when workflow changed status after reached to terminal status

jxu-nflx:
Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] WHOSUSING.md

Changes in this PR
----

- Add reason param to removeWorkflow methods to keep record of why the workflow needs to be removed as we'd like to keep a record of the reasons behind it.
- Reset workflow index when workflow changed status after already reached to terminal status. Possible actions include:  retry/restart/rerun/terminate operations. Reason for that is workflow could've been removed to secondary index already after reach to terminal status, while updateIndex will always create a new index on primary index.

Issue #

Alternatives considered
----

### DIFF
--- a/contribs/src/main/java/com/netflix/conductor/contribs/listener/archive/ArchivingWithTTLWorkflowStatusListener.java
+++ b/contribs/src/main/java/com/netflix/conductor/contribs/listener/archive/ArchivingWithTTLWorkflowStatusListener.java
@@ -29,6 +29,7 @@ public class ArchivingWithTTLWorkflowStatusListener implements WorkflowStatusLis
 
     private static final Logger LOGGER =
             LoggerFactory.getLogger(ArchivingWithTTLWorkflowStatusListener.class);
+    private static final String REASON = "archived";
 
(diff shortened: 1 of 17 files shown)


contribs/src/main/java/com/netflix/conductor/contribs/listener/archive/ArchivingWithTTLWorkflowStatusListener.java
```diff
@@ -29,6 +29,7 @@
 
     private static final Logger LOGGER =
             LoggerFactory.getLogger(ArchivingWithTTLWorkflowStatusListener.class);
+    private static final String ARCHIVE_OPERATION = "terminated";
```
apanicker-nflx, line 32: Did you intend to use this constant?

jxu-nflx: ya was intended to use this to replace some duplicated strings, but forgot. Thanks for spotting that. It is fixed now

PR #2781, opened 2022-02-23
add support to kafka IAM-tls

itzikSorani:
Pull Request type
----

- [ ] Bugfix
- [ X] Feature
- [ ] WHOSUSING.md

Changes in this PR
----

Add ability to interact with Kafka server that uses tls and client authentication, those configuration would be per request
Alternatives considered
----

### DIFF
--- a/contribs/build.gradle
+++ b/contribs/build.gradle
@@ -19,7 +19,7 @@ dependencies {
 
     implementation "com.amazonaws:aws-java-sdk-s3:${revAwsSdk}"
     implementation "com.amazonaws:aws-java-sdk-sqs:${revAwsSdk}"
-    implementation "software.amazon.msk:aws-msk-iam-auth"
+    implementation "software.amazon.msk:aws-msk-iam-auth:1.1.1"
 
     implementation "org.apache.commons:commons-lang3:"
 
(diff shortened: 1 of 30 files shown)


apanicker-nflx:
Please fix build failures.

apanicker-nflx:
Build still fails due to lack of version - 
image

contribs/build.gradle
```diff
@@ -19,6 +19,7 @@ dependencies {
 
     implementation "com.amazonaws:aws-java-sdk-s3:${revAwsSdk}"
     implementation "com.amazonaws:aws-java-sdk-sqs:${revAwsSdk}"
+    implementation "software.amazon.msk:aws-msk-iam-auth:1.1.1"
```
apanicker-nflx, line 22: Looks like this dependency is not used at compile time, why is this added?

apanicker-nflx:
> If you want to use IAM client auth you need this lib so it is used only if you add the IAM authentication in the request which declare the class in this lib 

Can you add this as a `runtimeOnly` dependency in that case?

contribs/build.gradle
```diff
@@ -19,6 +19,8 @@ dependencies {
...
+    implementation "com.amazonaws:aws-java-sdk-core:1.12.80"
```
apanicker-nflx, line 23: Can you comment on the need for this dependency? Also why is it pinned to `1.12.80`?

apanicker-nflx (changes requested):

apanicker-nflx:
If this is still needed, please open PR in https://github.com/Netflix/conductor-community.

Issue #2368, opened 2021-07-21
[BUG]: The tasks in the loop are scheduled concurrently


evgvain:
The loop is a fragment of the complex workflow. The order of the iteration is broken and tasks from the next iteration are dispatched before the previous iteration is completed.
The loop body consists of 3 tasks: set_variable, simple, set_variable. The tasks are executed consecutively but the 33rd iteration is mixed up with the 34th. The task set_dec_loop_completed__33 appears scheduled twice.

Conductor version: 3.0.4

Workflow definition:
![image](https://user-images.githubusercontent.com/6605905/126495641-d0092920-3d70-470e-afc8-bb44a8a471ee.png)

Tasks Details
![image](https://user-images.githubusercontent.com/6605905/126495695-87164d3d-5349-4141-9057-148e89ef61bf.png)

apanicker-nflx:
Hi @evgvain, is this reproducible? Could you please give us a few details about your setup? Also what conductor version are you running? 
If reproducible, please share the workflow definition that you used as well.

evgvain:
Conductor version: 3.0.4. Unfortunately, it is not deterministic - it happens from time to time. 
Please specify what details about the setup do you need.

apanicker-nflx:
> Please specify what details about the setup do you need.

Please provide details about the persistence layer, lock implementation and server deployment. 

Since this bug is not easily reproducible, it would take us some time to get around to it. But in the meantime, please feel free to send in a PR with the fix if you are able to get to it sooner.

evgvain:
I am working with stand alone MySQL db persistence model and stand alone Elasticsearch instance.
I am not sure about the lock implementation - I am using the default configuration. conductor-server, conductor-ui, Elasticsearch and MySQL are running on the same platform in dockers, managed by docker-compose.

evgvain:
This happen again. See below the tasks from the loop extracted from the Elasticsearch index conductor_task. You can see that the last task  (seq_qry_decrypt__9) is scheduled and starts before the previous (seq_qry_decrypt__8) ends. The task seq_qry_decrypt__8 even does not appear in conductor GUI. The order of 8th and 9th iteration is messed up.
`
[...]

PR #2814, opened 2022-03-04
Local only lock fix

manan164:
Cancel thread scheduled for stopping starvation in case of actually releasing a lock.
Background, 
As and when the lock is acquired, there is one [thread](https://github.com/Netflix/conductor/blob/main/contribs/src/main/java/com/netflix/conductor/contribs/lock/LocalOnlyLock.java#L78) is scheduled to release a lock. This is to stop the starvation scenario.
Now the issue is, when the operation is complete, the scheduled thread must be stopped otherwise after the interval it will take the lock preemptively.

Pull Request type
----

- [X] Bugfix
- [ ] Feature
- [ ] WHOSUSING.md

Changes in this PR
----
Canceled the thread to stop the starvation in a normal scenario.

Issue #https://github.com/Netflix/conductor/issues/2368

Alternatives considered
----

### DIFF
--- a/contribs/src/main/java/com/netflix/conductor/contribs/lock/LocalOnlyLock.java
+++ b/contribs/src/main/java/com/netflix/conductor/contribs/lock/LocalOnlyLock.java
@@ -41,7 +41,8 @@ public class LocalOnlyLock implements Lock {
                     return new Semaphore(1, true);
                 }
             };
-    private static final ConcurrentHashMap<String, ScheduledFuture<?>> scheduledFutures = new ConcurrentHashMap<>();
+    private static final ConcurrentHashMap<String, ScheduledFuture<?>> scheduledFutures =
+            new ConcurrentHashMap<>();
     private static final LoadingCache<String, Semaphore> lockIdtoSemaphoreMap =
             CacheBuilder.newBuilder().build(LOADER);
     private static final ThreadGroup THREAD_GROUP = new ThreadGroup("LocalOnlyLock-scheduler");
@@ -78,7 +79,8 @@ public class LocalOnlyLock implements Lock {
                 unit);
         if (acquireLock(lockId, timeToTry, unit)) {
             LOGGER.trace("Releasing {} automatically after {} {}", lockId, leaseTime, unit);
-            scheduledFutures.put(lockId, SCHEDULER.schedule(() -> releaseLock(lockId), leaseTime, unit));
+            scheduledFutures.put(
+                    lockId, SCHEDULER.schedule(() -> releaseLock(lockId), leaseTime, unit));
             return true;
         }
         return false;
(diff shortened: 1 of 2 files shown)


manan164:
Hi @v1r3n  Please review.

contribs/src/test/java/com/netflix/conductor/contribs/lock/LocalOnlyLockTest.java
```diff
@@ -12,107 +12,84 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
 
-@Ignore
-// Test causes "OutOfMemoryError: GC overhead limit reached" error during build
 public class LocalOnlyLockTest {
 
-    //    // Lock can be global since it uses global cache internally
```
jxu-nflx, line 21: Since you are on to fix the LocalOnlyLock, could you also take a look whether we can run those tests instead of comment out?

contribs/src/main/java/com/netflix/conductor/contribs/lock/LocalOnlyLock.java
```diff
@@ -39,7 +41,9 @@ public Semaphore load(String key) {
                     return new Semaphore(1, true);
                 }
             };
-    private static final LoadingCache<String, Semaphore> CACHE =
+    private static final ConcurrentHashMap<String, ScheduledFuture<?>> scheduledFutures =
```
aravindanr, line 44: Please use capitals for `scheduledFutures` and `lockIdtoSemaphoreMap` since they are constants.

contribs/src/main/java/com/netflix/conductor/contribs/lock/LocalOnlyLock.java
```diff
@@ -75,33 +79,47 @@ public boolean acquireLock(String lockId, long timeToTry, long leaseTime, TimeUn
...
+        lockIdtoSemaphoreMap.invalidate(lockId);
     }
 
     @VisibleForTesting
     LoadingCache<String, Semaphore> cache() {
-        return CACHE;
+        return lockIdtoSemaphoreMap;
+    }
+
+    ConcurrentHashMap<String, ScheduledFuture<?>> scheduledFutures() {
```
aravindanr, line 122: if this in package scope for testing purposes, please annotate with `@VisibleForTesting`.

aravindanr (changes requested):
@manan164 could you add more details about the bug that is fixed in this PR?

contribs/src/test/java/com/netflix/conductor/contribs/lock/LocalOnlyLockTest.java
manan164: Hi @jxu-nflx , Please see below they are uncommented. Only one test I have removed because that can cause the memory issue.

manan164: Hi @jxu-nflx , Please clarify on this.

jxu-nflx: I see, never mind, I saw those tests are added back.

jxu-nflx (approved):

aravindanr (approved):

PR #2816, opened 2022-03-04
Add previousStatus filed to WorkflowModel

jxu-nflx:
Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] WHOSUSING.md

Changes in this PR
----
Add previousStatus to WorkflowModel so it can be used to capture status change

Issue #

Alternatives considered
----

### DIFF
--- a/core/src/main/java/com/netflix/conductor/model/WorkflowModel.java
+++ b/core/src/main/java/com/netflix/conductor/model/WorkflowModel.java
@@ -116,7 +116,10 @@ public class WorkflowModel {
     }
 
     public void setStatus(Status status) {
-        setPreviousStatus(this.status);
+        // update previous status if current status changed
+        if (this.status != status) {
+            setPreviousStatus(this.status);
+        }
         this.status = status;
     }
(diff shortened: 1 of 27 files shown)


apanicker-nflx (approved):

core/src/main/java/com/netflix/conductor/model/WorkflowModel.java
```diff
@@ -101,11 +101,22 @@ public boolean isSuccessful() {
...
+    public void setPreviousStatus(Status status) {
+        this.previousStatus = status;
+    }
+
     public Status getStatus() {
         return status;
     }
 
     public void setStatus(Status status) {
+        setPreviousStatus(this.status);
```
aravindanr, line 119: Should we worry about `previousStatus` getting overridden if `setStatus` is called twice?

jxu-nflx: Good catch! Added a check for that

Issue #1677, opened 2020-05-11
Accessing DO_WHILE 'iteration' value inside the the looped tasks.


krishnaputhran:
Hi, anyone know how to access the iterator value inside the tasks, which are running as part of the DO_WHILE loop. For eg: I want to read an array value inside the running tasks. The array would be an input to the looped task from another task output, which was executed before the loop started. I am not sure how we can use the DO_WHILE 'iteration' variable for reading the array values.

I want to use the 'iteration' value something like below inside a task defined inside the DO_WHILE loop. Here 'preOrder_service' is task which runs before the the loop starts and 'products' is an array.

preOrder_service_get.output.response.body..products['iteration'].pluCode

manan164:
Hi @krishnaputhran ,
Please check the documentation here https://netflix.github.io/conductor/configuration/systask/#do-while-task.
You can use the LAMBDA task to parse the input array that is coming from the above task.
Here is the sample json that might help.
```
{
    "name": "test_loopTask",
    "taskReferenceName": "test_loopTask",
    "inputParameters": {
        "orders_count": "${workflow.input.orders.size}"
    },
    "type": "DO_WHILE",
    "loopCondition": "if ($.test_loopTask['iteration'] <= $.orders_count) { true; } else { false; }",
    "loopOver": [
        {
            "name": "get_order_at_index",
            "taskReferenceName": "get_order_at_index",
            "inputParameters": {
                "orders": "${workflow.input.orders}",
                "iterator": "$.test_loopTask['iteration']",
                "scriptExpression": "return $.orders.get($.iterator);"
            },
            "type": "LAMBDA"
        }
    ]
}
```

krishnaputhran:
Hi @manan164  thank you very much for the response and I appreciate your timely response. I have tried reading the do-while-task documentation, but couldn't get answer for my problem, but somehow I didn't consider the use fo LAMDA tasks.
I tried to follow your suggestion. But somehow, I feel that LAMDA task inside the loop is getting executed only once and hence same output from the first execution is getting passed in subsequent executions without getting the next value of the array. Please see my test workflow def and wrorkflow start scripts.

WorkfFlow Definition
==============
{
  "name": "test_do_while_workflow",
  "description": "test_do_while_workflow",
  "tasks": [
		{
            "name": "LoopTask",
            "taskReferenceName": "LoopTask",
            "type": "DO_WHILE",
            "inputParameters": {
[...]

PR #2881, opened 2022-03-29
Fix for accessing iteration value for loop over task.

manan164:
Pull Request type

 Bugfix
Changes in this PR
Whenever a task gets scheduled, ParamterUtils compute the task input and then [decider schedules](https://github.com/Netflix/conductor/blob/main/core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java#L820) the task.
Paramterutils relies on the [workflow ](https://github.com/Netflix/conductor/blob/main/core/src/main/java/com/netflix/conductor/core/utils/ParametersUtils.java#L102)to compute/populate the input for the task to be scheduled.
Now the issue is since the do_while task is also [yet to be scheduled ](https://github.com/Netflix/conductor/blob/main/core/src/main/java/com/netflix/conductor/core/execution/mapper/DoWhileTaskMapper.java#L105)so it is not added to workflow yet. So If any loopover task wants to use do_while task output/first iteration then it won't be able to use.
The fix is while figuring out the input for the loopover tasks, find all the pending do_while tasks and put iteration values as 1.

Describe the new behavior from this PR, and why it's needed
Issue #https://github.com/Netflix/conductor/issues/1677

Alternatives considered
Describe alternative implementation you have considered
Tried adding a do_while task in the workflow itself when the do_while task mapper is called but that lead to an unstable state of workflow. So decided not to go with this approach.


test-harness/src/test/groovy/com/netflix/conductor/test/integration/DoWhileSpec.groovy
```diff
@@ -124,6 +125,29 @@ class DoWhileSpec extends AbstractSpecification {
         }
     }
 
+    def "Test workflow with a iteration fix Do While task"() {
```
apanicker-nflx, line 128: Please name the spec so that it is easy to consume later.

test-harness/src/test/groovy/com/netflix/conductor/test/integration/DoWhileSpec.groovy
```diff
@@ -124,6 +125,29 @@ class DoWhileSpec extends AbstractSpecification {
...
+        given: "Number of iterations of the loop is set to 2"
+        def workflowInput = new HashMap()
+        workflowInput['loop'] = 2
+
+        when: "A do_while workflow is started"
+        def workflowInstanceId = workflowExecutor.startWorkflow("Do_While_Workflow_Iteration_Fix", 1, "looptest", workflowInput, null, null)
+
+        then: "Verify that the workflow has started"
```
apanicker-nflx, line 136: `Verify that the workflow has COMPLETED with tasks in expected state`

apanicker-nflx (approved):

aravindanr:
Hello @manan164, thanks for taking the time to fix this issue.  Placing the logic of finding all DO_WHILE tasks and setting an `iteration` is not ideal. But we understand the alternative that you tried and why it didn't work.

We think that it's best that the first `loopOverTask` is created **and persisted** in the `start` method of the [DoWhile](https://github.com/Netflix/conductor/blob/main/core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java) class. [WorkflowExecutor.scheduleTask persists the TaskModel and calls DoWhile.start after](https://github.com/Netflix/conductor/blob/main/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java#L1718-L1745). So if the `loopOverTask` is created in the start method, the DoWhile's task model is already persisted and available in the `WorkflowModel` for the `ParametersUtils` to use.

Let us know what you think.

aravindanr:
> Whenever do-while task mapper gets called it schedules the first loop over the task and to figure out the input required parameter utils gets called.

We are suggesting that the "first loop over" task is created and persisted in the `DoWhile.start()`.  The `DoWhileTaskMapper` should only create the TaskModel for the DO_WHILE task. [WorkflowExecutor.scheduleTask persists the TaskModel and calls DoWhile.start after](https://github.com/Netflix/conductor/blob/main/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java#L1718-L1745). So if the `loopOverTask` is created in the start method, the DoWhile's task model is already persisted and available in the `WorkflowModel` for the `ParametersUtils` to use. Let me know if this makes sense.

manan164:
Hi @aravindanr , 
```We are suggesting that the "first loop over" task is created and persisted in the DoWhile.start(). The DoWhileTaskMapper should only create the TaskModel for the DO_WHILE task.```
To calculate the first loop -over task also requires the task mapper for that task and the moment it gets called there can be a case where it requires iteration as input. Consider a case where the first task is the fork_join task and one of the tasks needs iteration inside the fork_join task.  Now writing a logic in task mapper will be late because for the first task everything would have been computed.

aravindanr:
@manan164 Sorry about the delay. Analyzing the right way to fix this bug, sent me down a rabbit hole. It turns out the `WorkflowExecutor.decide()` needs to be refactored for my above suggestion to work. I am working on refactoring and updating the tests. Hoping to have a PR by end of this week.

aravindanr:
Please see https://github.com/Netflix/conductor/pull/2968

manan164:
Thanks @aravindanr for coming up with better approach. Closing this PR since https://github.com/Netflix/conductor/pull/2968 is raised.

PR #2883, opened 2022-03-29
Event handler for loop over task.

manan164:
Pull Request type
----

- [X] Bugfix

Changes in this PR
----
Whenever the tasks are being retrieved for the event handler, the loop over tasks does not come since they are being appended by iteration value. This change enables performing an action on loop over task using an event handler.

Issue https://github.com/Netflix/conductor/discussions/2873

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/events/SimpleActionProcessor.java
+++ b/core/src/main/java/com/netflix/conductor/core/events/SimpleActionProcessor.java
@@ -12,10 +12,14 @@
  */
 package com.netflix.conductor.core.events;
 
+import java.util.Comparator;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
+import com.netflix.conductor.common.utils.TaskUtils;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/netflix/conductor/core/events/SimpleActionProcessor.java
```diff
@@ -123,6 +127,23 @@ public SimpleActionProcessor(
...
+            List<TaskModel> loopOverTaskList =
+                    workflow.getTasks().stream()
+                            .filter(
+                                    t ->
+                                            TaskUtils.removeIterationFromTaskRefName(
+                                                            t.getReferenceTaskName())
+                                                    .equals(taskRefName))
+                            .collect(Collectors.toList());
+            if (!loopOverTaskList.isEmpty()) {
+                // Find loopover task with the highest iteration value
```
manan164, line 140: The assumption here is loop over task with the highest iteration only will be waiting for an event. Since older iteration gets completed.

jxu-nflx:
Hello @manan164 can you also include an integration test for this?

apanicker-nflx (approved):

jxu-nflx (approved):

PR #2910, opened 2022-04-10
Allow Overrding of IDGenerator 

v1r3n:
Pull Request type
----

- [ ] Bugfix
- [ x] Feature
- [ ] WHOSUSING.md

Changes in this PR
----
IDGenerator is a fixed class today with a static generate() method to generate ID for workflows and tasks.
This makes it impossible to override the id generation logic for cases where users want to have a separate scheme for id generation.

This change allows users to override the IDGenerator class while keeping the default implementation intact.

Issue #

Alternatives considered
----

### DIFF
--- a/azureblob-storage/src/main/java/com/netflix/conductor/azureblob/config/AzureBlobConfiguration.java
+++ b/azureblob-storage/src/main/java/com/netflix/conductor/azureblob/config/AzureBlobConfiguration.java
@@ -19,6 +19,7 @@ import org.springframework.context.annotation.Configuration;
 
 import com.netflix.conductor.azureblob.storage.AzureBlobPayloadStorage;
 import com.netflix.conductor.common.utils.ExternalPayloadStorage;
+import com.netflix.conductor.core.utils.IDGenerator;
 
 @Configuration(proxyBeanMethods = false)
 @EnableConfigurationProperties(AzureBlobProperties.class)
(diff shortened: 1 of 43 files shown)


jxu-nflx:
Hello @v1r3n currently the Cassandra module requires an uuid for workflow and tasks: https://github.com/Netflix/conductor/blob/main/cassandra-persistence/src/main/java/com/netflix/conductor/cassandra/dao/CassandraBaseDAO.java#L152
So if a custom ID generator is used, user needs to be aware of that might not work with Cassandra module. Could you please add comment or document it somewhere?

Also, the id generator is all over the code bases, would you be interested to refactor that so to move everything to the storage layer?

v1r3n:
> Hello @v1r3n currently the Cassandra module requires an uuid for workflow and tasks: https://github.com/Netflix/conductor/blob/main/cassandra-persistence/src/main/java/com/netflix/conductor/cassandra/dao/CassandraBaseDAO.java#L152 So if a custom ID generator is used, user needs to be aware of that might not work with Cassandra module. Could you please add comment or document it somewhere?
> I have added a note on the Id Generator class to indicate that C* needs IDs to be valid UUIDs and the overriding should be done with careful consideration on the storage layer.
> Also, the id generator is all over the code bases, would you be interested to refactor that so to move everything to the storage layer?
Happy to do this, I did consider this earlier, but that will require significant refactoring of the code - I can send another PR after this one is merged so we can break this down in multiple pushes.

PR #2923, opened 2022-04-15
Remove guava from common module; Handle exceptions in sweeper

apanicker-nflx:
Pull Request type
----

- [x] Bugfix
- [x] Refactoring (no functional changes, no api changes)

Changes in this PR
----
- Removed guava from common module
- Fixed exception handling within the sweeper


core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -1160,33 +1159,15 @@ public void updateTask(TaskResult taskResult) {
             case IN_PROGRESS:
             case SCHEDULED:
                 try {
-                    String postponeTaskMessageDesc =
-                            "Postponing Task message in queue for taskId: " + task.getTaskId();
-                    String postponeTaskMessageOperation = "postponeTaskMessage";
-
-                    new RetryUtil<>()
-                            .retryOnException(
```
aravindanr, line 1168: Good call on removing these 👏🏼

core/src/main/java/com/netflix/conductor/core/events/DefaultEventProcessor.java
```diff
@@ -254,30 +256,21 @@ protected CompletableFuture<List<EventExecution>> executeActionsForEventHandler(
...
-                                                    action,
-                                                    payload,
-                                                    eventExecution.getEvent(),
-                                                    eventExecution.getMessageId()),
-                                    this::isTransientException,
-                                    null,
-                                    RETRY_COUNT,
-                                    description,
-                                    methodName);
+                    retryTemplate.execute(
```
aravindanr, line 267: Instead of using `RetryTemplate`, can we annotate `SimpleActionProcessor.execute` as `@Retryable`?

core/src/main/java/com/netflix/conductor/core/config/ConductorCoreConfiguration.java
```diff
@@ -47,6 +48,7 @@
 import static java.util.function.Function.identity;
 
 @Configuration(proxyBeanMethods = false)
+@EnableRetry
```
aravindanr, line 51: Is this needed? I see `RetryTemplate` is created per class as needed. I think `@EnableRetry` is only needed if we use `@Retryable`?

aravindanr:
1. Wherever `RetryTemplate` is created, i suggest that they are created with a qualifier in a Spring `@Configuration` class. The `RetryUtil` encapsulated the guava retry boilerplate from the classes that need to use them. The `createRetryTemplate()` method brings the boilerplate back to the classes that use it. 
2. I am not sure if we need blind retries over ES client calls, but that's a separate discussion.

core/src/main/java/com/netflix/conductor/core/events/DefaultEventProcessor.java
apanicker-nflx: no, because we want to inspect the cause of the exception and retry only in specific cases.

core/src/main/java/com/netflix/conductor/core/config/ConductorCoreConfiguration.java
apanicker-nflx: good catch, this was a leftover from when i was trying to use `@Retryable`

apanicker-nflx:
>I am not sure if we need blind retries over ES client calls, but that's a separate discussion.

I removed the retries over single request ES client calls, and only added retries on the bulk request which can have partial failures.

cassandra-persistence/src/main/java/com/netflix/conductor/cassandra/dao/CassandraExecutionDAO.java
```diff
@@ -775,18 +748,9 @@ public void removeTaskFromLimit(TaskModel task) {
...
-                                    session.execute(
-                                            deleteTaskDefLimitStatement.bind(
-                                                    task.getTaskDefName(),
-                                                    UUID.fromString(task.getTaskId()))),
-                            null,
-                            null,
-                            3,
-                            "Deleting from task_def_limit",
-                            "removeTaskFromLimit");
+            session.execute(
```
jxu-nflx, line 751: Although we don't use those anymore, but would it be a concern that removing the retries from `addTasktoLimit` and `removeTaskFromLimit` cause other user to start observer issues?

client/src/main/java/com/netflix/conductor/client/automator/TaskPollExecutor.java
```diff
@@ -259,7 +254,7 @@ private void executeTask(Worker worker, Task task) {
                 worker.getClass().getSimpleName(),
                 worker.getIdentity(),
                 result.getStatus());
-        updateWithRetry(updateRetryCount, task, result, worker);
+        updateTaskResult(updateRetryCount, task, result, worker);
```
jxu-nflx, line 257: updateRetryCount is a property that we expose to our client that can be configed. Our internal spring boot client library also expose that. Why do we want to remove the retries here?

core/src/main/java/com/netflix/conductor/core/reconciliation/WorkflowSweeper.java
```diff
@@ -89,12 +85,13 @@ public void sweep(String workflowId) {
...
+                return;
             }
         } catch (Exception e) {
-            queueDAO.setUnackTimeout(
-                    DECIDER_QUEUE, workflowId, properties.getWorkflowOffsetTimeout().toMillis());
             Monitors.error(CLASS_NAME, "sweep");
             LOGGER.error("Error running sweep for " + workflowId, e);
         }
+        queueDAO.setUnackTimeout(
+                DECIDER_QUEUE, workflowId, properties.getWorkflowOffsetTimeout().toMillis());
```
jxu-nflx, line 95: Before the change, if any exception occur during any step of this method, we will call setUnackTimeout` to extend lease. With the change, it is still seem to be the same thing. Did I miss anything? What is the bug that we are trying to fix?

es6-persistence/src/main/java/com/netflix/conductor/es6/config/ElasticSearchV6Configuration.java
```diff
@@ -122,15 +125,29 @@ public RestClientBuilder restClientBuilder(ElasticSearchProperties properties) {
...
-            Client client, ElasticSearchProperties properties, ObjectMapper objectMapper) {
-        return new ElasticSearchDAOV6(client, properties, objectMapper);
+            Client client,
+            @Qualifier("es6RetryTemplate") RetryTemplate retryTemplate,
+            ElasticSearchProperties properties,
+            ObjectMapper objectMapper) {
+        return new ElasticSearchDAOV6(client, retryTemplate, properties, objectMapper);
+    }
+
+    @Bean(name = "es6RetryTemplate")
```
jxu-nflx, line 144: This can also be replaced with `@Retrayable(backoff=@Backoff(delay=1000))` if prefer

common/src/main/java/com/netflix/conductor/common/run/Workflow.java
```diff
@@ -328,7 +320,9 @@ public void setPriority(int priority) {
      * @return the workflow definition name.
      */
     public String getWorkflowName() {
-        Preconditions.checkNotNull(workflowDefinition, "Workflow definition is null");
+        if (workflowDefinition == null) {
+            throw new IllegalArgumentException("Workflow definition is null");
```
aravindanr, line 324: `Preconditions.checkNotNull` throws `NullPointerException`, i think we should keep the same here.

contribs/src/main/java/com/netflix/conductor/contribs/storage/S3PayloadStorage.java
```diff
@@ -149,7 +150,7 @@ public InputStream download(String path) {
             S3Object s3Object = s3Client.getObject(new GetObjectRequest(bucketName, path));
             return s3Object.getObjectContent();
         } catch (SdkClientException e) {
-            String msg = "Error communicating with S3";
+            String msg = String.format("Error downloading from S3 - path:%s", path);
```
aravindanr, line 153: +1

core/src/main/java/com/netflix/conductor/core/events/DefaultEventProcessor.java
```diff
@@ -336,4 +313,26 @@ private Object getPayloadObject(String payload) {
         }
         return payloadObject;
     }
+
+    private RetryTemplate createRetryTemplate() {
```
aravindanr, line 317: Should this move to `ConductorCoreConfiguration`?

core/src/main/java/com/netflix/conductor/core/utils/Utils.java
```diff
@@ -119,10 +111,26 @@ public static void checkNotNullOrEmpty(String input, String errorMessage) {
      * @throws ApplicationException if input object is not valid
      */
     public static void checkNotNull(Object object, String errorMessage) {
-        try {
-            Preconditions.checkNotNull(object, errorMessage);
-        } catch (NullPointerException exception) {
-            throw new ApplicationException(ApplicationException.Code.INVALID_INPUT, errorMessage);
+        if (object == null) {
+            throw new IllegalArgumentException(errorMessage);
```
aravindanr, line 115: `NullPointerException`

es6-persistence/src/main/java/com/netflix/conductor/es6/config/ElasticSearchV6Configuration.java
aravindanr: change method name to `es6RetryTemplate()`?

es6-persistence/src/main/java/com/netflix/conductor/es6/dao/index/ElasticSearchRestDAOV6.java
```diff
@@ -130,12 +126,15 @@ public class ElasticSearchRestDAOV6 extends ElasticSearchBaseDAO implements Inde
...
 
+    private final RetryTemplate retryTemplate;
+
     static {
         SIMPLE_DATE_FORMAT.setTimeZone(GMT);
     }
 
     public ElasticSearchRestDAOV6(
             RestClientBuilder restClientBuilder,
+            RetryTemplate retryTemplate,
```
aravindanr, line 137: Should this have a `@Qualifier`?

test-harness/build.gradle
```diff
@@ -15,6 +15,8 @@ dependencies {
     testImplementation project(':conductor-client')
     testImplementation project(':conductor-grpc-client')
 
+    testImplementation "org.springframework.retry:spring-retry"
```
aravindanr, line 18: Don't we need also need this in `server/build.gradle` as `implementation`?

cassandra-persistence/src/main/java/com/netflix/conductor/cassandra/dao/CassandraExecutionDAO.java
apanicker-nflx: I had added this more as a safeguard than an actual problem, so i think it would be ok to remove this. I do not foresee this issue to manifest and if it does it will fail the operation instead.

test-harness/build.gradle
apanicker-nflx: I was able to bring up the server locally using `gradlew bootRun`. What am I overlooking?

core/src/main/java/com/netflix/conductor/core/reconciliation/WorkflowSweeper.java
apanicker-nflx: >Before the change, if any exception occur during any step of this method, we will call setUnackTimeout` to extend lease.

Actually, when an ApplicationException was thrown with a `Code != NOT_FOUND` the method setUnackTimeout was not invoked. So, in case of transient errors (Code == BACKEND_ERROR), the message would be removed from the decider queue. This change rectifies this behavior.

common/src/main/java/com/netflix/conductor/common/run/Workflow.java
apanicker-nflx: ahh good point, i missed that

core/src/main/java/com/netflix/conductor/core/utils/Utils.java
apanicker-nflx: fixed

es6-persistence/src/main/java/com/netflix/conductor/es6/dao/index/ElasticSearchRestDAOV6.java
apanicker-nflx: no, the bean creation happens in ElasticsearchV6Configuration and the Qualifier is used there.

client/src/main/java/com/netflix/conductor/client/automator/TaskPollExecutor.java
apanicker-nflx: good catch, will need to add a retry loop here

core/src/main/java/com/netflix/conductor/core/events/DefaultEventProcessor.java
apanicker-nflx: changed

es6-persistence/src/main/java/com/netflix/conductor/es6/config/ElasticSearchV6Configuration.java
apanicker-nflx: The method to be retried is a private method and ideally should not be exposed. The annotation only works on public methods, hence I chose to use RetryTemplate. Open to change if you feel otherwise

jxu-nflx (approved):

client/src/main/java/com/netflix/conductor/client/automator/TaskPollExecutor.java
apanicker-nflx: added a retry loop for update operations

core/src/main/java/com/netflix/conductor/core/utils/Utils.java
aravindanr: Shouldn't this be `NullPointerException`?

es6-persistence/src/main/java/com/netflix/conductor/es6/dao/index/ElasticSearchDAOV6.java
```diff
@@ -117,12 +114,15 @@ public class ElasticSearchDAOV6 extends ElasticSearchBaseDAO implements IndexDAO
...
 
+    private final RetryTemplate retryTemplate;
+
     static {
         SIMPLE_DATE_FORMAT.setTimeZone(GMT);
     }
 
     public ElasticSearchDAOV6(
             Client elasticSearchClient,
+            RetryTemplate retryTemplate,
```
aravindanr, line 125: Doesn't this need a `@Qualifier`?

aravindanr (approved):

test-harness/build.gradle
apanicker-nflx: Added to server's build.gradle

Issue #2927, opened 2022-04-18
AMQP ack failed when an event task is executed and not consumed, while the workflow is terminated


amirpopovich:
**Describe the bug**

**File**: Event Task (core\src\main\java\com\netflix\conductor\core\execution\tasks\Event.java).

**Issue**: The `start` method creates a `Message` object using `task.getTaskId()` as the receipt property. 
The value of `task.getTaskId()` is a _UUID_ represented as _String_.
When the task's `cancel` method is called, it tries to ack the message, probably in order to avoid the consumption of the event when a workflow is terminated.
In the `AMQPObservableQueue.ack` method, there is a call to ` chn.basicAck(Long.parseLong(message.getReceipt()), false);` which expects the `message.getReceipt()` to be parse as a _Long_, while the current underlying value is a _UUID_.
This causes an exception and the message will stay in the amqp queue.

**Details**
Conductor version: 3.6.1
Persistence implementation:Dynomite
Queue implementation: Dynoqueues 
Lock: Redis

**To Reproduce**
1. Create a workflow with an event task
2. Publish the event without consuming it
3. Cancel the workflow

**Expected behavior**
The message created should have a receipt with a _Long_ value to avoid errors.

amirpopovich:
After looking into the issue, I think that the problem is with the ack call on the task's cancel method.
My guess is that the ack call was made to avoid consumption of the event when the task is terminated. 

There are two problems in terms of _AMQP_ (specifically, _RabbitMQ_):

1. The message's delivery tag is managed by the channel and cannot be set externally.
Additionally, the delivery tag is not saved anywhere (i.e. on the task itself or something) after the message is published.
Therefore, future ack's can be made only by consumers which will read the delivery tag from the consumed message's envelope.
Trying to ack a specific message by a generated delivery tag (`task.getTaskId()`) like in the task's cancel method isn't supported by _AMQP_.
[Reference - section 1.1](https://www.rabbitmq.com/resources/specs/amqp-xml-doc0-9-1.pdf )

2. The cancel method can be called while the message was already consumed and acked by some kind of external process.
Re-acking it again will cause a "delivery-tag doesn't exist error".

An attempt to ack a non-existing delivery tag will result in the channel being shutdown with the following error:

[...]

PR #2928, opened 2022-04-18
Fix message delivery tag to be of long type (and not UUID)

AvitalOfstein:
Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] WHOSUSING.md

Changes in this PR
----

Issue #2927 
Fix message delivery tag to be of long type (and not GUID) since can be parsed later for ack on a message, on AMQPObservableQueue class -> ack method -> chn.basicAck(Long.parseLong(message.getReceipt()), false)

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/tasks/Event.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/tasks/Event.java
@@ -63,7 +63,7 @@ public class Event extends WorkflowSystemTask {
 
         try {
             String payloadJson = objectMapper.writeValueAsString(payload);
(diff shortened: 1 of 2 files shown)


jxu-nflx:
@AvitalOfstein From what I read _https://github.com/Netflix/conductor/issues/2927_, the issue is specific to AMQP. So instead of limit that to Long in the core module, please make the change on the AMQP module.

AvitalOfstein:
We updated the issue (#2927 ), this won't solve our problem, since we can't set actually the delivery tag. Closing this PR, as not relevant.

Issue #2533, opened 2021-10-25
[QUESTION]:  How to set redis password


dzx2018:
There is no redis password attribute in RedisProperties. Can't I set the redis password?

PR #2950, opened 2022-04-26
Redis password authentication support

jmigueprieto:
Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] WHOSUSING.md

Changes in this PR
----
Added Redis password authentication support for `conductor.db.type`
- `redis_standalone` 
- `redis_cluster`
- `redis_sentinel`

Issue #2533

---
The password is set as the 4th param in the host configuration. E.g.:

```properties
conductor.redis.hosts=localhost:6379:us-east-1c:my_str0ng_pazz
```

For Cluster and Sentinel configuration the password from the first host is used. 

In Sentinel configuration sentinels and redis nodes use the same password.

image

Alternatives considered
----

Considered changing `RedisProperties` to add a new password property un `conductor.redis`. That would have made it more explicit but since `host.getPassword()` already exists it seems like an OK alternative if documented properly.

### DIFF
--- a/server/src/main/resources/application.properties
+++ b/server/src/main/resources/application.properties
@@ -18,8 +18,9 @@ conductor.db.type=memory
 
 conductor.indexing.enabled=false
 
-#Dynomite Cluster details.
+#Redis configuration details.
 #format is host:port:rack separated by semicolon
+#Auth is supported. Password is taken from host[0]. format: host:port:rack:password
 conductor.redis.hosts=host1:port:rack;host2:port:rack:host3:port:rack
 
 #namespace for the keys stored in Dynomite/Redis
--- a/docs/docs/how-tos/redis.md
+++ b/docs/docs/how-tos/redis.md
@@ -1,0 +1,38 @@
+# Redis Configuration
+
+By default conductor runs with an in-memory Redis mock. However, you
+can change the configuration by setting the properties `conductor.db.type` and `conductor.redis.hosts`.
+
+## `conductor.db.type`
+
+| Value                          | Description                                                                            |
+|--------------------------------|----------------------------------------------------------------------------------------|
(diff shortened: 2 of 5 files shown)


apanicker-nflx:
Please add documentation about usage of this new field. As such, users would need to read through the code to make use of this feature.

jmigueprieto:
@apanicker-nflx sure - I'll add some documentation.

What would be the best place for it and its extent?

The only thing I was able to find related to `conductor.redis.hosts` is a comment in `application.properties`

```
#format is host:port:rack separated by semicolon
conductor.redis.hosts=host1:port:rack;host2:port:rack:host3:port:rack
```

Would it be enough just adding to that? or would you prefer having something like "Configuring Redis" under `/docs/how-tos` and an entry in `spring-configuration-metadata.json`?

apanicker-nflx:
A comment in `application.properties` would help devs.
Additionally, a section in the docs to elaborate would provide more clarity. @dougsillars Could you point to the best place for adding this documentation?

dougsillars:
docs/docs/how-tos/redis.md is probably the best place, IMO.

server/src/main/resources/application.properties
```diff
@@ -18,8 +18,9 @@ conductor.db.type=memory
 
 conductor.indexing.enabled=false
 
-#Dynomite Cluster details.
```
jmigueprieto, line 21: Changed this comment because these properties apply not only to Dynomite. They apply to other Redis configurations.

jmigueprieto:
Added a Redis Configuration page to docs with documentation related to the scope of this PR.

Screen Shot 2022-05-04 at 11 59 45

We can later work with @dougsillars to improve it.

jmigueprieto:
Some tests are failing after rebasing changes, I'll take a look.

jmigueprieto:
nvm, the build/test issue was fixed before I got a chance to look at it

Issue #2884, opened 2022-03-29
Is it possible to check empty input values and change them with default within the ${} expression?


AliakseiBuhakou:
### Discussed in https://github.com/Netflix/conductor/discussions/2878

<div type='discussions-op-text'>

<sup>Originally posted by **AliakseiBuhakou** March 28, 2022</sup>
Hello. Could you please guide me with following question.
There is an workflow example.
`{
			"name": "task1",
			"taskReferenceName": "task1",
			"inputParameters": {
				"http_request": {
					"uri": "some_uri",
					"method": "POST",
					"body": { some_body
					},
					"connectionTimeOut": "600000",
					"readTimeOut": "600000",
					"headers": {
						"header1": "${workflow.input.header1}"
					}
				}
			}`

The value for "header1": "${workflow.input.header1}" is optional. Is it possible to check the value and change it with some default string? It can be solved with some JSON_JQ_TRANSFORM probably, is there a way to do this without lambda? The reason of the question is that on Conductor v3+ it fails with NPE if value is not provided, Conductor v2+ works fine</div>

jxu-nflx:
Duplicate in thread: https://github.com/Netflix/conductor/discussions/2878 So close this one

PR #2964, opened 2022-05-03
Fix for NPE when header values are null

muralicmr:
Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] WHOSUSING.md

Changes in this PR
----
Fix NPE when optional headers like below are not provided.

```
"headers": {
    "header1": "${workflow.input.header1}"
}
```
Issue #2884 
In the new version of conductor HTTP tasks fail with NPE is a header value is not provided. As HTTP task is generic it is a good practice to support header values which are defined in workflow but not sent while making a call to execute the workflow.
Also its good to note that this feature was working in 2.x version of conductor. 

Alternatives considered
----
No alternatives considered


apanicker-nflx:
Please run `./gradlew spotlessApply` to fix formatting issues.

Issue #313, opened 2017-08-29
 get /tasks/poll/{tasktype}  problem


lfozvye:
*environment：
es
dynomite
redis

*The detail workflow
taskdefs
[
  {    
    "name": "test",
    "retryCount": 1,
    "timeoutSeconds": 120,
    "inputKeys": [
        "build_input"
    ],
    "outputKeys": [
        "data",
        "error",
        "msg",
        "status"
    ],
    "timeoutPolicy": "TIME_OUT_WF",
    "retryLogic": "FIXED",
    "retryDelaySeconds": 0,
    "responseTimeoutSeconds": 360
  }
]

metadata/workflow
{
    "name": "test",
    "description": "my workflow for ci flow",
    "version": 1,
    "tasks": [
        {
            "name": "test",
            "taskReferenceName": "node1",
            "inputParameters": {
                "build_input": "${workflow.input.build_input}"
            },
            "type": "SIMPLE",
            "startDelay": 0
        }
    ],
    "outputParameters": {
        "build_data": "${node1.output.data}",
        "build_error": "${node1.output.error}",
        "build_status": "${node1.output.status}",
        "build_msg": "${node1.output.msg}"
    },
    "schemaVersion": 2
}

[workflow name]
{}  

[exec task]
![image](https://user-images.githubusercontent.com/6927559/29814040-8a89f20e-8cde-11e7-9a87-b2ba24fa37ff.png)

[conductor output]
289400 [qtp1701436909-32] DEBUG org.eclipse.jetty.io.ChannelEndPoint  - filled 0 SelectChannelEndPoint@157ef56f{/10.200.0.181:59151<->8080,Open,in,out,-,-,0/30000,HttpConnection@61662ab2}{io=0/0,kio=0,kro=1}
289401 [qtp1701436909-32] DEBUG org.eclipse.jetty.io.ChannelEndPoint  - filled 0 SelectChannelEndPoint@157ef56f{/10.200.0.181:59151<->8080,Open,in,out,-,-,0/30000,HttpConnection@61662ab2}{io=0/0,kio=0,kro=1}
[...]

PR #2982, opened 2022-05-10
Fix update with large payload task result

apanicker-nflx:
Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] WHOSUSING.md

Changes in this PR
----
Fixes a bug where large payload location in task result was not being propagated as part of task result by the client.

### DIFF
--- a/client/src/main/java/com/netflix/conductor/client/automator/TaskPollExecutor.java
+++ b/client/src/main/java/com/netflix/conductor/client/automator/TaskPollExecutor.java
@@ -279,18 +279,16 @@ class TaskPollExecutor {
         try {
             TaskResult finalResult =
                     retryOperation(
-                            (TaskResult taskResult) -> {
-                                taskClient.evaluateAndUploadLargePayload(
-                                        taskResult, task.getTaskType());
-                                return null;
-                            },
+                            (TaskResult taskResult) ->
+                                    taskClient.evaluateAndUploadLargePayload(
+                                            taskResult.getOutputData(), task.getTaskType()),
                             count,
                             result);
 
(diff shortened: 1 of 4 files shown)


client/src/main/java/com/netflix/conductor/client/automator/TaskPollExecutor.java
```diff
@@ -306,13 +304,22 @@ private void updateTaskResult(int count, Task task, TaskResult result, Worker wo
...
-            Function<TaskResult, Void> operation, int count, TaskResult result) {
+            Function<TaskResult, Optional<String>> operation, int count, TaskResult result) {
         int index = 0;
         while (index < count) {
             try {
-                TaskResult taskResult = result.copy();
-                operation.apply(taskResult);
-                return taskResult;
+                Optional<String> optionalExternalStorageLocation = operation.apply(result);
+                if (optionalExternalStorageLocation.isPresent()) {
```
jxu-nflx, line 312: This is confusing. Can we create two retryOperation method? one for the uploadtoexternalstorage, which takes a taskresult output map and returns an externalstoragepath, another one for the updatetask, which takes the final taskresult from last step, and return void

jxu-nflx: If we want to keep just one `retryoperation` method, we can just remove line #313 `TaskResult taskResult = result.copy();` from it. I don't see the point of why we want to make a copy of it for every retry

aravindanr (approved):

jxu-nflx (approved):

PR #3009, opened 2022-05-26
Fix: Enable extra task fields in Java-SDK

lj-ovo:
I have been trying to set the `Start Delay` and `Optional` fields in the Java SDK.  I can't find a whole load of documentation on what the SDK supports, but a quick read through the code seems to suggest the `Start Delay` and `Optional` fields are not copied into the `WorkflowTask`, meaning when I register by Workflows/Tasks, there is no current way to set these values except through the API/UI.

* Enables `Start Delay` field for a Task.
* Enables `Optional` field for a Task.

PLEASE NOTE:

If this is indeed an issue, I will create two new tests which:
* Creates a `Simple Task`, sets the `startDelay` field and asserts against the value given in the workflow to make sure equals the expected value.
* * Creates a `Simple Task`, sets the `optional` field and asserts against the value given in the workflow to make sure equals the expected value.

Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] WHOSUSING.md

Changes in this PR
----

Issue #

Alternatives considered
----

### DIFF
--- a/java-sdk/src/main/java/com/netflix/conductor/sdk/workflow/def/tasks/Task.java
+++ b/java-sdk/src/main/java/com/netflix/conductor/sdk/workflow/def/tasks/Task.java
@@ -205,6 +205,8 @@ public abstract class Task<T> {
         workflowTask.setWorkflowTaskType(type);
         workflowTask.setDescription(description);
         workflowTask.setInputParameters(input);
+        workflowTask.setStartDelay(startDelay);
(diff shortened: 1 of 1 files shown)


jxu-nflx:
Hello @lj-ovo the build failed because of the spotless checks, could you update your PR by  running './gradlew :conductor-java-sdk:spotlessApply' to fix these violations.

PR #3035, opened 2022-06-07
[WIP] Nashorn engine cleanup.

manan164:
Pull Request type
----
- [X] Bugfix

Changes in this PR
----
The current implementation does not allow injecting script engine from outside. Also, the methods are static so it is not possible to use it with another scripting engine.
Issue #


manan164:
Closing the PR since this changes are not needed as of now.

PR #3036, opened 2022-06-08
Nashorn cleanup.

manan164:
- [X] Bugfix
Do not allow external binding.
Do not allow java scope


manan164:
By mistake raised so closing.

PR #3055, opened 2022-06-21
Bugfix/jointask output

jxu-nflx:
Pull Request type
----
- [x] Bugfix
- [ ] Feature
- [ ] WHOSUSING.md

Changes in this PR
----
* re-throw TerminateWorkflowException for task update
* Add only non-empty task output for Join

Issue #

Alternatives considered
----

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/tasks/Join.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/tasks/Join.java
@@ -60,7 +60,10 @@ public class Join extends WorkflowSystemTask {
             if (hasFailures) {
                 failureReason.append(forkedTask.getReasonForIncompletion()).append(" ");
             }
-            task.addOutput(joinOnRef, forkedTask.getOutputData());
+            // Only add to task output if it's not empty
+            if (!forkedTask.getOutputData().isEmpty()) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/netflix/conductor/core/execution/tasks/Join.java
```diff
@@ -60,7 +60,10 @@ public boolean execute(
             if (hasFailures) {
                 failureReason.append(forkedTask.getReasonForIncompletion()).append(" ");
             }
-            task.addOutput(joinOnRef, forkedTask.getOutputData());
+            // Only add to task output if it's not empty
+            if (!forkedTask.getOutputData().isEmpty()) {
```
aravindanr, line 64: Does `getOutputData()` always return a non-null value?

aravindanr (approved):

core/src/main/java/com/netflix/conductor/core/execution/tasks/Join.java
jxu-nflx: yes, we always init it to empty map

v1r3n: FYI this breaks backward compatibility if anyone relied on this to be present.

PR #3058, opened 2022-06-22
Parameterized support for duration and until in wait task.

manan164:
Pull Request type
----
- [X] Bugfix

Changes in this PR
----
Parameterized input support for wait task. Removed the static constraint from the wait task while saving workflow definition.


v1r3n (approved):

core/src/main/java/com/netflix/conductor/validations/WorkflowTaskTypeConstraint.java
```diff
@@ -271,19 +270,6 @@ private boolean isWaitTaskValid(
                 context.buildConstraintViolationWithTemplate(message).addConstraintViolation();
                 valid = false;
             }
-
-            try {
```
aravindanr, line 275: Don't we still need this check if the values are not parameters? Instead of removing the check, can this be changed to check if the value of `duration` or `until` is a parameter and if its not, then perform these checks?

core/src/main/java/com/netflix/conductor/validations/WorkflowTaskTypeConstraint.java
```diff
@@ -271,19 +270,6 @@ private boolean isWaitTaskValid(
...
-                if (StringUtils.isNotBlank(duration)) {
-                    DateTimeUtils.parseDuration(duration);
-                } else if (StringUtils.isNotBlank(until)) {
-                    DateTimeUtils.parseDate(until);
-                }
-            } catch (Exception e) {
```
aravindanr, line 281: Please update this to catch the correct exceptions from the methods, `NumberFormatException`, `IllegalArgumentException` and `ParseException`.

aravindanr (changes requested):

manan164:
Hi @aravindanr  Please review.

apanicker-nflx (approved):

PR #3070, opened 2022-06-28
Performance improvements in Decider

v1r3n:
Pull Request type
----
- [ ] Bugfix
- [ ] Feature
- [ ] WHOSUSING.md
- [x] Other (please describe):
**Performance improvements**
* Use the workflowModel loaded from the database at the start of the `decide()` method calls during recursion reducing the number of times workflow is loaded from database.
* Improve check for workflowCompletion reducing the no. of times the tasks are evaluated which can be slow for large workflows
* Add a state inside DeciderOutcome to record if the workflow has a terminal task and capture the task to avoid checking later.
* Reduce and avoid extra recursion for the forks - which has proper nesting of fork/joins now if there are nested forks.
* Enable tests for the test-harness.  
* Note: 2 tests which depends on the client changes are marked as @Ignore as they need to be updated with new client changes.  These can be re-enabled.

With the above changes we have been able to measure marked performance improvements and reduced CPU load when running large workloads.

Changes in this PR
----

Issue #

Alternatives considered
----


aravindanr:
@v1r3n The tests are fixed. You can rebase from main

core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
```diff
@@ -882,6 +892,7 @@ public static class DeciderOutcome {
         List<TaskModel> tasksToBeScheduled = new LinkedList<>();
         List<TaskModel> tasksToBeUpdated = new LinkedList<>();
         boolean isComplete;
+        Optional<TaskModel> terminateTask = Optional.empty();
```
aravindanr, line 895: `Optional` is typically not used as a field. https://blog.joda.org/2014/11/optional-in-java-se-8.html

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -611,7 +613,8 @@ public void restart(String workflowId, boolean useLatestDefinitions) {
             throw e;
         }
 
-        decide(workflowId);
+        metadataMapperService.populateWorkflowWithDefinitions(workflow);
+        decide(workflow.getWorkflowId());
```
jxu-nflx, line 617: can `workflow` be passed in here instead of just workflowId, so it won't need to reload the workflow model again?

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -1270,28 +1264,40 @@ public List<String> getRunningWorkflowIds(String workflowName, int version) {
...
+            executionLockService.releaseLock(workflowId);
+        }
+    }
+
+    /**
+     * @param workflow the workflow to evaluate the state for
+     * @return true if the workflow has completed (success or failed), false otherwise.
+     * @throws ApplicationException If there was an error - caller should retry in this case.
+     */
+    public WorkflowModel decide(WorkflowModel workflow) {
```
jxu-nflx, line 1294: If this method called directly, there is no lock acquire required?

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
v1r3n: keeping it consistent with other re* methods like re-run, retry etc to always re-load frrom the database and evaluate.  Given these are not very frequently used methods.

jxu-nflx (approved):

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -1248,34 +1241,44 @@ public List<String> getRunningWorkflowIds(String workflowName, int version) {
...
-            return decide(workflowId);
+
+            WorkflowModel workflow = executionDAOFacade.getWorkflowModel(workflowId, true);
+            if (workflow == null) {
+                // This can happen if the workflowId is incorrect
+                return null;
+            }
+            // FIXME Backwards compatibility for legacy workflows already running.
+            // This code will be removed in a future version.
+            workflow = metadataMapperService.populateWorkflowWithDefinitions(workflow);
```
jxu-nflx, line 1259: These code lines are removed from main repo now

test-harness/src/test/java/com/netflix/conductor/test/integration/grpc/GrpcEndToEndTest.java
```diff
@@ -13,12 +13,14 @@
...
 
 import org.junit.Before;
+import org.junit.Ignore;
 
 import com.netflix.conductor.client.grpc.EventClient;
 import com.netflix.conductor.client.grpc.MetadataClient;
 import com.netflix.conductor.client.grpc.TaskClient;
 import com.netflix.conductor.client.grpc.WorkflowClient;
 
+@Ignore
```
apanicker-nflx, line 23: > 2 tests which depends on the client changes are marked as @ignore as they need to be updated with new client changes. These can be re-enabled.

Do these need to be re-enabled in this PR? Or will that be done as part of client change PR?

apanicker-nflx (approved):

test-harness/src/test/java/com/netflix/conductor/test/integration/grpc/GrpcEndToEndTest.java
v1r3n: Yes re-enabled them now that client is fixed.

PR #3109, opened 2022-07-14
Update start time in updateTask if it is unset

jun-he:
Update start time in updateTask if it is unset.

Pull Request type
----
- [x] Bugfix
- [ ] Feature
- [ ] WHOSUSING.md

Changes in this PR
- Update start time in updateTask if it is unset.


jun-he:
@aravindanr Can you please help to review it? Thanks.

jun-he:
Wondering if we should only check start time without checking `IN_PROGRESS`. So in case if the job failed directly, we won't leave start time unset.

aravindanr (approved):

aravindanr:
> Wondering if we should only check start time without checking `IN_PROGRESS`. So in case if the job failed directly, we won't leave start time unset.

If status is not checked, the startTime will be set when the task is in SCHEDULED state.

PR #3120, opened 2022-07-20
Async system task should get timeout.

manan164:
----
- [X] Bugfix

Changes in this PR
----
Async system task was never getting timedout. This is the behavior change introduced in the [PR](https://github.com/Netflix/conductor/pull/2181). Earlier the behavior was to let async system task to be timed out.
Issue #

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
@@ -745,7 +745,7 @@ public class DeciderService {
             return false;
         }
 
-        if (task.getStatus().isTerminal() || isAyncCompleteSystemTask(task)) {
+        if (task.getStatus().isTerminal() || task.getTaskType().equals(SUB_WORKFLOW.name()) {
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


venkathangella (approved):
Fix the build failures and merge this.

venkathangella (approved):

core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
```diff
@@ -745,7 +746,7 @@ boolean isResponseTimedOut(TaskDef taskDefinition, TaskModel task) {
             return false;
         }
 
-        if (task.getStatus().isTerminal() || isAyncCompleteSystemTask(task)) {
+        if (task.getStatus().isTerminal() || task.getTaskType().equals(SUB_WORKFLOW.name())) {
```
apanicker-nflx, line 749: @manan164 The contract for the async complete system tasks is that the response for this task will be sent asynchronously (either via API or events). For such tasks, response time out will not apply. 

However, the task timeout will still be applied and is enforced as per the `checkTaskTimeout` method. 

Could you please provide more info about this change?

PR #3154, opened 2022-08-06
Change TaskType to be string in order to define custom task and taskMappers

manan164:
Pull Request type
----

Using String task to enable custom tasks and taskMapper to be implemented in future.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/config/ConductorCoreConfiguration.java
+++ b/core/src/main/java/com/netflix/conductor/core/config/ConductorCoreConfiguration.java
@@ -30,7 +30,6 @@ import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.retry.support.RetryTemplate;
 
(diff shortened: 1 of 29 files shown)


apanicker-nflx (approved):

james-deee:
Just my 2 cents here, but loosening this type to a String doesn't seem quite like the correct thing to do here. Wonder if there's a better thing we can do to achieve your goal.

@manan164 can you give everyone a more descriptive, in depth explanation of what you are trying to get by doing this? I can't quite tell by your comment above.

manan164:
Hi @james-deee , The motivation behind this PR is to make the conductor task more extensible. Consider a case where we are using a conductor jar and want to add a few system tasks. As of now, there is no way to do that since TaskType is an enum and that can't be extended. This PR enables that functionality. So to add a system task two things are needed (assuming PR merged),
1. Extend a WorkflowSystemTask
2. Create TaskMapper for that Task
Hope it is clear now. Happy to answer any further questions.

PR #3196, opened 2022-08-23
Adding parameterUtils in do-while task to compute input at each iteration.

manan164:
Pull Request type
----
- [X] Bugfix

Changes in this PR
----
Do-While condition evaluation flow does not rely on parameterutils so compute the task input. Doing so will result in an inconsistent state when `set_variable` task is used as one of `loopOver` task and its output is used in `loopCondition`.
In fact, in all the tasks we need parameterUtils to be injected/used because any task can use input that relies on workflow variables.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java
@@ -218,8 +218,7 @@ public class DoWhile extends WorkflowSystemTask {
     }
 
     @VisibleForTesting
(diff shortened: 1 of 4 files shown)


manan164:
Hi @aravindanr @apanicker-nflx , I will add one integration test for this use case once we finalize the approach.

manan164:
Hi @aravindanr @apanicker-nflx , Please review.

manan164:
Hi @aravindanr  @apanicker-nflx any update on this please?

aravindanr (approved):

apanicker-nflx (approved):

PR #3216, opened 2022-09-01
During start, decide is performed inside a lock

aravindanr:
Pull Request type
----
- [x] Bugfix
- [ ] Feature
- [ ] WHOSUSING.md

Changes in this PR
----

_A recent refactoring caused **decide during startWorkflow** to be performed outside a lock. This caused decide to be performed concurrently on two threads. This typically results in the first task of the workflow to be scheduled twice._

Alternatives considered
----

_None_

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
@@ -376,10 +376,7 @@ public class WorkflowExecutor {
         }
 
         try {
-            createWorkflow(workflow);
-            executionDAOFacade.populateWorkflowAndTaskPayloadData(workflow);
-            // then decide to see if anything needs to be done as part of the workflow
-            decide(workflow);
+            createAndDecide(workflow);
             Monitors.recordWorkflowStartSuccess(
                     workflow.getWorkflowName(),
                     String.valueOf(workflow.getWorkflowVersion()),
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -1620,7 +1619,7 @@ public void skipTaskFromWorkflow(
             taskToBeSkipped.setOutputMessage(skipTaskRequest.getTaskOutputMessage());
         }
         executionDAOFacade.createTasks(Collections.singletonList(taskToBeSkipped));
-        decide(workflow);
+        decide(workflow.getWorkflowId());
```
apanicker-nflx, line 1622: What if we wrap invoking `decide(WorkflowModel)` within a lock here? Since we already have the workflow loaded in memory here, we dont need to perform this again.

aravindanr: `decide(workflowId)` calls `decide(WorkflowModel)` and the former is already has a lock. `decide(WorkflowModel)` is also called recursively which means the lock needs to be re-entrant. Since `skipTaskFromWorkflow` is used sparingly, the additional call to load `WorkflowModel` shouldn't have a big impact.

jxu-nflx (approved):

apanicker-nflx (approved):

Issue #3211, opened 2022-08-30
Variables are cleared once the WAIT task is started


vzubyk:
**Bug description:**
`Variables` are cleared once the `WAIT` task is started.

**Details:**
Conductor version: 3.11.0
Workflow definition:
```
{
  "name": "Set_Variable_Workflow",
  "description": "Set a value to a variable and wait",
  "version": 1,
  "tasks": [
    {
      "name": "Set_Name",
      "taskReferenceName": "Set_Name",
      "type": "SET_VARIABLE",
      "inputParameters": {
        "name": "Foo"
      }
    },
    {
      "name": "Test_Wait",
      "taskReferenceName": "Test_Wait",
      "type": "WAIT",
      "inputParameters": {
        "duration": "30 s"  
      } 
    }
  ],
  "restartable": true,
  "ownerEmail":"abc@example.com",
  "workflowStatusListenerEnabled": true,
  "schemaVersion": 2
}
```

**Steps to reproduce the behavior:**
1. Create provided above workflow definition
2. Start a new workflow for the `Set_Variable_Workflow` definition
3. Open the workflow execution page
4. Check the `Variables`

**Actual behavior:**
No variables exist.

**Expected behavior:**
The variable `name` with the value `Foo`, created by the `Set_Name` task exists.

----------
**Screenshots:**
<img width="1679" alt="image" src="https://user-images.githubusercontent.com/65384250/187446059-067b316f-32fc-43fe-9895-a98b116f4f28.png">
<img width="1679" alt="image" src="https://user-images.githubusercontent.com/65384250/187446213-f77dd0be-33fd-4cd7-8b0e-38af8b36c21f.png">

manan164:
Hi @vzubyk , Thanks for reporting this. I am able to reproduce this. We will raise a fix for this.

manan164:
HI @vzubyk , PR for the fix https://github.com/Netflix/conductor/pull/3221

PR #3221, opened 2022-09-07
Update workflow in DAO in case of update of workflow related attribute.

manan164:
Pull Request type
----
- [X] Bugfix

Changes in this PR
----
Update workflow DAO in case of any workflow-related attribute updates. For example workflow variables. 

Issue #3211

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
@@ -413,10 +413,10 @@ public class WorkflowExecutor {
                     workflow.getWorkflowName(),
                     workflow.getWorkflowId());
             executionDAOFacade.populateWorkflowAndTaskPayloadData(workflow);
-            int hashCode = workflow.hashCode();
+            int hashCodeBeforeDecider = workflow.hashCode();
             WorkflowModel workflowModel = decide(workflow);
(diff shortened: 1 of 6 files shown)


astelmashenko:
@manan164, thank you for the fix. Does anything prevent merging of this fix?

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -1367,10 +1373,6 @@ WorkflowModel decide(WorkflowModel workflow) {
                 return decide(workflow);
             }
 
-            if (!outcome.tasksToBeUpdated.isEmpty() || !tasksToBeScheduled.isEmpty()) {
-                executionDAOFacade.updateWorkflow(workflow);
-            }
-
             return workflow;
```
jxu-nflx, line 1376: Couple of issues:
1. the fix is not going to work for cases the set_variable task is not the first task of the workflow
2. hashcode computation of workflow on every execution is not going to be efficient
3. remove block (i.e. line 1370-1373 above) from the decide method could cause side effect

manan164: Hi @jxu-nflx , Updated as per the comment.

aravindanr (approved):

apanicker-nflx (approved):

PR #3222, opened 2022-09-07
Allow updating task status and output via POST call by workflowId and taskRefName

c4lm:
Pull Request type
----
- [x] Feature

Changes in this PR
----
Allow updating task status using a POST call for situations where workflowId and taskReferenceName are already known ( before task is assigned a taskId) and sent to some external service which is supposed to update a WAIT task.

### DIFF
--- a/rest/src/main/java/com/netflix/conductor/rest/controllers/TaskResource.java
+++ b/rest/src/main/java/com/netflix/conductor/rest/controllers/TaskResource.java
@@ -83,6 +83,25 @@ public class TaskResource {
         return taskService.updateTask(taskResult);
     }
 
+    @PostMapping(value = "/{workflowId}/{taskRefName}/{status}", produces = TEXT_PLAIN_VALUE)
+    @Operation(summary = "Update a task By Ref Name")
+    public String updateTask(
+            @PathVariable("workflowId") String workflowId,
+            @PathVariable("taskRefName") String taskRefName,
+            @PathVariable("status") TaskResult.Status status,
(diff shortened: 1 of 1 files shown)


rest/src/main/java/com/netflix/conductor/rest/controllers/TaskResource.java
```diff
@@ -83,6 +83,25 @@ public String updateTask(@RequestBody TaskResult taskResult) {
...
 
+    @PostMapping(value = "/{workflowId}/{taskRefName}/{status}", produces = TEXT_PLAIN_VALUE)
+    @Operation(summary = "Update a task By Ref Name")
+    public String updateTask(
+            @PathVariable("workflowId") String workflowId,
+            @PathVariable("taskRefName") String taskRefName,
+            @PathVariable("status") TaskResult.Status status,
+            @RequestBody Map<String, Object> output) {
+
+        Task pending = taskService.getPendingTaskForWorkflow(workflowId, taskRefName);
```
apanicker-nflx, line 94: In the current implementation, this is an expensive DB operation and the DB queries are not optimized to serve this well.
Additionally, the API itself is designed to be too constrained to only allow updating status and output. It should ideally allow updating the TaskResult.

Having said that, we are working on a proposal to overhaul the DB schema model to better support large workflows and enable support for these types of queries. We will revisit this API at that time.

c4lm:
Closed, can still be used as reference for whoever needs it.

Issue #2312, opened 2021-06-15
Replace Nashorn engine with GraalVM


apanicker-nflx:
With the release of Java 11, Nashorn is deprecated, and has been removed from JDK 15 onwards. The engine needs to be replaced with GraalVM.

[Helpful article for migration](https://medium.com/graalvm/oracle-graalvm-announces-support-for-nashorn-migration-c04810d75c1f)

ijusti:
is there any updates?

martonbtoth:
Is there any reason the PR resolving this was closed? It would make sense to support less ancient java versions.

PR #3233, opened 2022-09-11
Replace Nashorn with GraalVM JS Engine

BlasiusSecundus:
Pull Request type
----

- [x] Feature

Changes in this PR
----
Issue #2312 replace Nashorn with GraalVM JS Engine

Nashorn is deprecated in JDK 11 and removed in JDK 15, resulting in runtime errors (and test failures) on JDK 15+

### DIFF
--- a/java-sdk/src/main/java/com/netflix/conductor/sdk/workflow/def/tasks/Javascript.java
+++ b/java-sdk/src/main/java/com/netflix/conductor/sdk/workflow/def/tasks/Javascript.java
@@ -43,7 +43,7 @@ public class Javascript extends Task<Javascript> {
 
     private static final String EVALUATOR_TYPE_PARAMETER = "evaluatorType";
 
-    private static final String ENGINE = "nashorn";
+    private static final String ENGINE = "graal.js";
 
     /**
      * Javascript tasks are executed on the Conductor server without having to write worker code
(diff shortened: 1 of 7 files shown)


apanicker-nflx (approved):

java-sdk/src/main/java/com/netflix/conductor/sdk/workflow/def/tasks/Javascript.java
```diff
@@ -105,6 +105,7 @@ public Javascript validate() {
             LOGGER.error("missing " + ENGINE + " engine.  Ensure you are running supported JVM");
             return this;
         }
+        scriptEngine.put("polyglot.js.allowHostAccess", true);
```
boney9, line 108: Why do we need this?

BlasiusSecundus: This is required to access Java objects bound as input. (Nashorn allows this by default, but GraalVM requires explicit setting for security reasons: https://www.graalvm.org/22.2/reference-manual/js/NashornMigrationGuide/#secure-by-default). You can  validate this by removing the line and running the tests: it will result in errors like this:  `TypeError: Cannot read property 'testKey1' of undefined`

This is equivalent to `HostAccess.ALL` (see https://www.graalvm.org/22.2/reference-manual/js/ScriptEngine/) which should be identical to the Nashorn behavior. 

An alternative would be to use `HostAccess.EXPLICIT` for greater security, but that would result in a breaking change.

To my understanding, workflows are considered part of the source code of the host application and can be trusted at the same level, thus this should not pose a security risk.

cwensel:
Is there any technical reason this cannot be merged?

Java 11 active support ends in two months: https://endoflife.date/java

PR #3272, opened 2022-10-05
re-throw TerminateWorkflowException if payload exceed max threshold s…

jxu-nflx:
…o to fail workflow and task

Pull Request type
----
- [x] Bugfix
- [ ] Feature
- [ ] WHOSUSING.md

Changes in this PR
----
Fix the issue that if a task output exceed threshold, the workflow appears to be stuck in RUNNING state.

Issue #

Alternatives considered
----

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/utils/ExternalPayloadStorageUtils.java
+++ b/core/src/main/java/com/netflix/conductor/core/utils/ExternalPayloadStorageUtils.java
@@ -192,6 +192,8 @@ public class ExternalPayloadStorageUtils {
             }
         } catch (TransientException te) {
             throw te;
+        } catch (TerminateWorkflowException twe) {
+            throw twe;
         } catch (Exception e) {
             LOGGER.error(
                     "Unable to upload payload to external storage for workflow: {}", workflowId, e);
--- a/test-harness/src/test/java/com/netflix/conductor/test/utils/MockExternalPayloadStorage.java
+++ b/test-harness/src/test/java/com/netflix/conductor/test/utils/MockExternalPayloadStorage.java
@@ -168,4 +168,19 @@ public class MockExternalPayloadStorage implements ExternalPayloadStorage {
         }
         return new HashMap<>();
     }
+
+    public Map<String, Object> createLargePayload(int repeat) {
+        Map<String, Object> largePayload = new HashMap<>();
+        try {
+            InputStream inputStream = readOutputDotJson();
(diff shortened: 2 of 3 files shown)


test-harness/src/test/java/com/netflix/conductor/test/utils/MockExternalPayloadStorage.java
```diff
@@ -168,4 +168,19 @@ public Map<String, Object> downloadPayload(String path) {
         }
         return new HashMap<>();
     }
+
+    public Map<String, Object> createLargePayload(int repeat) {
```
apanicker-nflx, line 172: Does this need to be a parameterized method? Why not use a static large payload?

core/src/main/java/com/netflix/conductor/core/utils/ExternalPayloadStorageUtils.java
```diff
@@ -192,6 +192,8 @@ public <T> void verifyAndUpload(T entity, PayloadType payloadType) {
             }
         } catch (TransientException te) {
             throw te;
+        } catch (TerminateWorkflowException twe) {
```
aravindanr, line 195: I presume that this fails the workflow, but the task whose payload is large will not be marked as FAILED. Should we try to fail the respective task?

test-harness/src/test/java/com/netflix/conductor/test/utils/MockExternalPayloadStorage.java
jxu-nflx: yes could do that as well, but I find this way is more flexible to get the payload size I want for different tests. Like in this example, the fork join task only spawn 2 tasks, so I can play around with the parameter there to make sure the join output exceed the size limit. If for other tests, there are more tasks spawn from fork join, I could always use different parameter to achieve the purpose instead of having to create precise static large payload for different workflows

core/src/main/java/com/netflix/conductor/core/utils/ExternalPayloadStorageUtils.java
jxu-nflx: It will fail the task as well if the terminateworkflowexception is due to task exceed size limit: https://github.com/Netflix/conductor/blob/ecfa7b10e620b6edc230d10d3eeed932e9f04784/core/src/main/java/com/netflix/conductor/core/utils/ExternalPayloadStorageUtils.java#L220

There is a test case added verified that as well

apanicker-nflx (approved):

aravindanr (approved):

PR #3273, opened 2022-10-05
Fail workflow if input to dynamic fork join task is invalid

jxu-nflx:
Pull Request type
----
- [x] Bugfix
- [ ] Feature
- [ ] WHOSUSING.md

Changes in this PR
----
Fix the issue that uncaught exception thrown when the task input of FORK_JOIN_DYNAMIC is invalid, and task seem to be stuck. Also startWorkflow will fail with 500 in this case if FORK_JOIN_DYNAMIC task is synchronous evaluated when workflow starts.

Issue #

Alternatives considered
----

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/mapper/ForkJoinDynamicTaskMapper.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/mapper/ForkJoinDynamicTaskMapper.java
@@ -195,9 +195,17 @@ public class ForkJoinDynamicTaskMapper implements TaskMapper {
             }
 
             for (TaskModel forkedTask : forkedTasks) {
-                Map<String, Object> forkedTaskInput =
-                        tasksInput.get(forkedTask.getReferenceTaskName());
-                forkedTask.addInput(forkedTaskInput);
+                try {
+                    Map<String, Object> forkedTaskInput =
+                            tasksInput.get(forkedTask.getReferenceTaskName());
+                    forkedTask.addInput(forkedTaskInput);
+                } catch (Exception e) {
+                    String reason =
+                            String.format(
+                                    "Tasks could not be dynamically forked due to invalid input: %s",
+                                    e.getMessage());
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/netflix/conductor/core/execution/mapper/ForkJoinDynamicTaskMapper.java
```diff
@@ -195,9 +195,17 @@ public List<TaskModel> getMappedTasks(TaskMapperContext taskMapperContext)
...
-                        tasksInput.get(forkedTask.getReferenceTaskName());
-                forkedTask.addInput(forkedTaskInput);
+                try {
+                    Map<String, Object> forkedTaskInput =
+                            tasksInput.get(forkedTask.getReferenceTaskName());
+                    forkedTask.addInput(forkedTaskInput);
+                } catch (Exception e) {
+                    String reason =
+                            String.format(
+                                    "Dynamic tasks could not be created due to invalid task input: %s ",
```
apanicker-nflx, line 205: nit: Since, `DYNAMIC` is a task type already, this would probably read better as ```Tasks could not be dynamically forked due to invalid input: %s```

apanicker-nflx (approved):

test-harness/src/test/groovy/com/netflix/conductor/test/base/AbstractSpecification.groovy
```diff
@@ -61,4 +61,8 @@ abstract class AbstractSpecification extends Specification {
 
         workflowExecutor.startWorkflow(input)
     }
+
+    protected String startWorkflow(StartWorkflowInput startWorkflowInput) {
```
aravindanr, line 65: nit: instead of this method in the base class, the sub-classes can directly call `workflowExecutor.startWorkflow()`. It adds a layer of indirection, which is not necessary.

aravindanr (approved):

PR #3284, opened 2022-10-11
JOIN task is made async

apanicker-nflx:
Pull Request type
----
- [ ] Bugfix
- [x] Feature
- [ ] WHOSUSING.md

Changes in this PR
----
The JOIN task is made async and will be evaluated by the async system task executor and the system task workers off the JOIN task queue. This improves the efficiency of workflow evaluations because the workflow will now be evaluated only after the JOIN task reaches a terminal state rather than for every forked task update leading to empty evaluations.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
@@ -66,7 +66,13 @@ public class WorkflowExecutor {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(WorkflowExecutor.class);
     private static final int EXPEDITED_PRIORITY = 10;
-
+    private static final String CLASS_NAME = WorkflowExecutor.class.getSimpleName();
+    private static final Predicate<TaskModel> UNSUCCESSFUL_TERMINAL_TASK =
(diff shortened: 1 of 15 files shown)


jxu-nflx (approved):

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutor.java
```diff
@@ -870,10 +866,7 @@ public void updateTask(TaskResult taskResult) {
                     task.getTaskDefName(), lastDuration, false, task.getStatus());
         }
 
-        // sync evaluate workflow only if the task is not within a forked branch
-        if (isLazyEvaluateWorkflow(workflowInstance.getWorkflowDefinition(), task)) {
-            expediteLazyWorkflowEvaluation(workflowId);
-        } else {
+        if (!isLazyEvaluateWorkflow(workflowInstance.getWorkflowDefinition(), task)) {
```
franrull, line 869: @apanicker-nflx What's the intent of this change? It seems like an oversight cause if the condition is false (in other words, the workflow `isLazyEvalute`), nothing happens anymore.

Holmesus:
Hi~ Make JOIN task async causing a problem.Can anyone fix it or just make JOIN sync ?#3528

PR #3289, opened 2022-10-11
Set unack/postpone duration based on task state

skorse:
Pull Request type
----
- [ ] Bugfix
- [X] Feature
- [ ] WHOSUSING.md

Changes in this PR
Set unack/postpone duration based on task state. 

For workflows with a task in IN_PROGRESS state, the postpone duration is set as (responseTimeoutSeconds + 1). Multiple tasks in IN_PROGRESS state, use the first task. If responseTimeoutSeconds is 0, use a default non-zero value as postpone duration.

For workflows with a task in SCHEDULED state, the postpone duration is set as (pollTimeoutSeconds + 1). If pollTimeoutSeconds is not set, the workflowDef.timeoutSeconds is used for postponeDuration. If the workflowDef.timeoutSeconds is not set, a default non-zero value is used the postpone duration. Multiple tasks in SCHEDULED state, use the first task's value. For deducing the default value, the queue_wait_time metric can be used.

For workflows with a WAIT task in IN_PROGRESS state, the postpone duration is set as the wait timeout seconds computed on the task. If wait timeout is not set, this task is waiting for an external trigger to complete the task. In such cases, the postpone duration will be a default non-zero value.

Alternatives considered
----

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/reconciliation/WorkflowSweeper.java
+++ b/core/src/main/java/com/netflix/conductor/core/reconciliation/WorkflowSweeper.java
@@ -22,6 +22,7 @@ import org.springframework.scheduling.annotation.Async;
 import org.springframework.stereotype.Component;
 
 import com.netflix.conductor.common.metadata.tasks.TaskDef;
+import com.netflix.conductor.common.metadata.tasks.TaskType;
 import com.netflix.conductor.common.metadata.workflow.WorkflowTask;
 import com.netflix.conductor.core.WorkflowContext;
 import com.netflix.conductor.core.config.ConductorProperties;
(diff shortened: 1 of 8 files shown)


aravindanr:
Can you add a test for the changes?

core/src/main/java/com/netflix/conductor/core/reconciliation/WorkflowSweeper.java
```diff
@@ -88,7 +96,66 @@ public void sweep(String workflowId) {
             Monitors.error(CLASS_NAME, "sweep");
             LOGGER.error("Error running sweep for " + workflowId, e);
         }
+        if (workflow != null) {
+            unack(workflow);
+        } else {
+            queueDAO.setUnackTimeout(
```
apanicker-nflx, line 102: Let's log here to detect what causes this and fix thereafter.

core/src/main/java/com/netflix/conductor/core/reconciliation/WorkflowSweeper.java
```diff
@@ -88,7 +96,66 @@ public void sweep(String workflowId) {
...
+                            (taskModel.getResponseTimeoutSeconds() != 0)
+                                    ? taskModel.getResponseTimeoutSeconds() + 1
+                                    : properties.getWorkflowOffsetTimeout().getSeconds();
+                }
+                break;
+            }
+            if (taskModel.getStatus() == Status.SCHEDULED) {
+                Optional<TaskDef> taskDefinition = taskModel.getTaskDefinition();
+                if (taskDefinition.isEmpty()) {
+                    taskDefinition =
```
apanicker-nflx, line 129: If task definition is empty within the task model, it would not exist within the workflow task either. You can skip the `isEmpty` check here

apanicker-nflx (approved):

BrandonDotLin:
This feature actually made my workflow always stuck for a long time if Response Timeout for including task is long...

Issue #1505, opened 2020-01-22
Tasks not removed from ElasticSearch when deleting workflow


NLRemco:
When removing a workflow instance using the CRUD API. The workflow index itself is removed from ElasticSearch, but the attached tasks are not removed.

When doing a `\DELETE` on the `http://conductor/api/workflow/<workflow-uuid>/remove?archiveWorkflow=false` endpoint, a `204 No Content` is returned: all the occurrences are removed from Redis, the workflow index is removed from ElasticSearch, but the tasks of that workflow are not.

apanicker-nflx:
@RemcoBuddelmeijer Thank you for reporting this issue. We will look into fixing this in a future release, in the meantime please feel free to contribute the fix for this if you need this sooner.

annair-dev:
@RemcoBuddelmeijer True! 
Memory cleanup has to be addressed @apanicker-nflx . 
We need a bulk delete API which would delete the Workflow along with the task entries from the DB and index. 
Do we have one now?

apanicker-nflx:
@annair-dev This improvement is yet to be made to the remove workflow API. Please feel free to contribute a fix.

james-deee:
Just wanted to see if we could re-open, since it's not fixed. Anyone working on a fix for this?

RemcoBuddelmeijer:
Seeing that it has been two years and have flown under my radar, I wouldn't mind taking a look at this myself. So if no one else is working on this I'll be creating a PR. @james-deee

apanicker-nflx:
@RemcoBuddelmeijer that would be great. Thank you

RemcoBuddelmeijer:
@apanicker-nflx Can the issue be reopened? Currently working on the PR but would like for the progress to be followed here as well.

RemcoBuddelmeijer:
Still working on the issue. Been a couple of very busy weeks. Currently only fixing harness tests then the PR will be posted.
@james-deee @apanicker-nflx

apanicker-nflx:
Closed by #3300

james-deee:
Still waiting on the PR https://github.com/Netflix/conductor/issues/1505 to be merged

james-deee:
Re-opening because it still hasn't merged

NLRemco:
PR has been merged. Closing issue. Thank you!

PR #3300, opened 2022-10-17
Remove tasks from Elasticsearch when workflow is removed

NLRemco:
Pull Request type
----
- [x] Bugfix
- [x] Feature
- [ ] WHOSUSING.md

Changes in this PR
----
* Implement removal of tasks when calling /workflow/remove
* Written tests for the new behaviour
  - Introduced tests for WorkflowServiceTest
  - Introduced tests for WorkflowResourceTest
  - Introduced tests for ExecutionDAOFacadeTest
  - Introduced tests for QueueResilienceSpec.groovy
  - Introduced tests for TestElasticSearchRestDAOV6 
  - Introduced tests for TestElasticSearchDAOV6

* Updated Javadocs where necessary to reflect on the new changes

Issue #1505 

Tasks were not removed from Elasticsearch once the workflow was removed. This is necessary to not leave hanging objects in Elasticsearch that are not connected to any existing instance of a workflow. This change allows you to specify to also remove the tasks.

Alternatives considered
----
None

Additional notes
----
Options for removing tasks and archiving tasks have both been considered. In the final draft only the option to remove tasks was introduced. This to keep the changes clear and communicate to the user that tasks are removed now. On top of this all of this was done as it's unknown if the users are using this undefined behaviour of tasks not being removed. This was later on removed in the PR as it can be trusted that users do not rely on unidentified behaviour.

Right now to minimize the impact of the change the option to toggle on removing tasks or not is present. Only methods being called by the WorkflowResource class for workflow removal allows you to delete tasks, any other class such as WorkflowServiceImpl does **not** implement this yet. This can be done in a future PR.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/dal/ExecutionDAOFacade.java
+++ b/core/src/main/java/com/netflix/conductor/core/dal/ExecutionDAOFacade.java
@@ -564,8 +564,8 @@ public class ExecutionDAOFacade {
                 indexDAO.updateTask(
                         workflow.getWorkflowId(),
                         task.getTaskId(),
-                        new String[] {RAW_JSON_FIELD, ARCHIVED_FIELD},
-                        new Object[] {objectMapper.writeValueAsString(task), true});
+                        new String[] {ARCHIVED_FIELD},
+                        new Object[] {true});
             } else {
                 throw new IllegalArgumentException(
                         String.format(
--- a/core/src/main/java/com/netflix/conductor/core/dal/ExecutionDAOFacade.java
+++ b/core/src/main/java/com/netflix/conductor/core/dal/ExecutionDAOFacade.java
@@ -333,27 +333,18 @@ public class ExecutionDAOFacade {
      * Removes the workflow from the data store.
      *
      * @param workflowId the id of the workflow to be removed
-     * @param archiveWorkflow if true, the workflow will be archived in the {@link IndexDAO} after
-     *     removal from {@link ExecutionDAO}. Next to this if removeTasks is true, the tasks
-     *     associated with the workflow will also be archived in the {@link IndexDAO} after removal
-     *     from {@link ExecutionDAO}.
-     * @param removeTasks if true, the tasks associated with the workflow will be removed
+     * @param archiveWorkflow if true, the workflow and associated tasks will be archived in the
+     *     {@link IndexDAO} after removal from {@link ExecutionDAO}.
      */
-    public void removeWorkflow(String workflowId, boolean archiveWorkflow, boolean removeTasks) {
-        if (!removeTasks) {
-            LOGGER.info("Not removing tasks of workflow: {}", workflowId);
-        }
-
+    public void removeWorkflow(String workflowId, boolean archiveWorkflow) {
         WorkflowModel workflow = getWorkflowModelFromDataStore(workflowId, true);
         List<TaskModel> tasks = workflow.getTasks();
 
         executionDAO.removeWorkflow(workflowId);
-        if (removeTasks) {
-            tasks.forEach(
-                    task -> {
-                        executionDAO.removeTask(task.getTaskId());
-                    });
-        }
+        tasks.forEach(
+                task -> {
+                    executionDAO.removeTask(task.getTaskId());
+                });
 
         try {
             removeWorkflowIndex(workflow, archiveWorkflow);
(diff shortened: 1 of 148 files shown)


RemcoBuddelmeijer:
Once reviewed I will resolve the merge conflicts.

es6-persistence/src/main/java/com/netflix/conductor/es6/dao/index/ElasticSearchRestDAOV6.java
```diff
@@ -810,6 +810,95 @@ public CompletableFuture<Void> asyncUpdateWorkflow(
...
+                                .must(QueryBuilders.termQuery("taskId", taskId))
+                                .toString(),
+                        "*",
+                        0,
+                        1,
+                        null);
+
+        if (taskSearchResult.getTotalHits() == 0) {
+            LOGGER.error("Task: {} does not belong to workflow: {}", taskId, workflowId);
+        }
```
ghoshabhi, line 831: why do we continue execution even after we don't find the task with the given id ?

RemcoBuddelmeijer: You are correct, I will go ahead and extend the tests for the changes. I will create a separate PR to fix this behaviour in `#removeWorkflow` alongside some other changes that need to be done. Thank you for noticing!

jxu-nflx (approved):

RemcoBuddelmeijer:
I have gone ahead and resolved the merge conflicts @jxu-nflx @apanicker-nflx @ghoshabhi

apanicker-nflx:
@RemcoBuddelmeijer Thanks for creating the PR and 💯 on adding a wide slew of tests.

While I understand the reasoning for adding a flag `removeTasks` to the API and service layers, it should be the default behavior. The current implementation where tasks are left orphaned when workflow is removed is really a bug and your PR directly addresses this. Hence, I would suggest removing the flag and making it the default behavior.

core/src/main/java/com/netflix/conductor/core/dal/ExecutionDAOFacade.java
```diff
@@ -509,6 +555,29 @@ public void removeTask(String taskId) {
         executionDAO.removeTask(taskId);
     }
 
+    private void removeTaskIndex(WorkflowModel workflow, TaskModel task, boolean archiveTask)
+            throws JsonProcessingException {
+        if (archiveTask) {
+            if (task.getStatus().isTerminal()) {
+                // Only allow archival if task is in terminal state
+                // DO NOT archive async, since if archival errors out, task data will be lost
+                indexDAO.updateTask(
```
apanicker-nflx, line 564: The archival process for task does not need a rawJSON field because the archival process for workflow, contains all of the tasks within in the stringified JSON blob stored in its rawJSON field.

RemcoBuddelmeijer:
> @RemcoBuddelmeijer Thanks for creating the PR and 💯 on adding a wide slew of tests.
> 
> While I understand the reasoning for adding a flag `removeTasks` to the API and service layers, it should be the default behavior. The current implementation where tasks are left orphaned when workflow is removed is really a bug and your PR directly addresses this. Hence, I would suggest removing the flag and making it the default behavior.

Amazing! Nice to know that we are on the same line and that no complicated logic has to be implemented just for task archival. I'll keep it as it is and remove it from the description of the PR! Thank you! 😄

core/src/main/java/com/netflix/conductor/core/dal/ExecutionDAOFacade.java
RemcoBuddelmeijer: Removed! Will keep the tasks just for auditing purposes.

RemcoBuddelmeijer:
@apanicker-nflx It seems like that after updating the QueueResilienceSpec test to also include archiving workflows that it doesn't terminate tasks associated with the workflow model. I tried querying for the tasks after removal but this didn't seem to have any impact. Is this perhaps a bug on the Redis/Cassandra side of the ExecutionDAO?

core/src/main/java/com/netflix/conductor/core/dal/ExecutionDAOFacade.java
```diff
@@ -332,19 +333,46 @@ public void removeFromPendingWorkflow(String workflowType, String workflowId) {
...
+
+        List<TaskModel> tasks = workflow.getTasks();
         try {
             removeWorkflowIndex(workflow, archiveWorkflow);
         } catch (JsonProcessingException e) {
             throw new TransientException("Workflow can not be serialized to json", e);
         }
-
-        executionDAO.removeWorkflow(workflowId);
+        tasks.forEach(
```
james-deee, line 350: Why loop through the tasks twice? A regular for loop might be easier to read here, and couldn't you do both operations through a single loop of the tasks?

james-deee:
The only other thing I would call out is that the Community Version will need to be updated as well for the V7 ElasticSearch IndexDAO. So whoever releases this fix (thanks for fixing btw!!!!!), will need to be aware when they bump the version over in Community @apanicker-nflx

core/src/main/java/com/netflix/conductor/core/dal/ExecutionDAOFacade.java
RemcoBuddelmeijer: Personally I found it more readable to have two separate loops with separate error handling. When it comes to a traditional for loop: I wanted to keep in line with the style of the file where a for each is used instead of a traditional for loop. Either way since it’s unnecessary to loop over something twice and now that the order of method has been changed, I’ll merge the loops.

RemcoBuddelmeijer:
> The only other thing I would call out is that the Community Version will need to be updated as well for the V7 ElasticSearch IndexDAO. So whoever releases this fix (thanks for fixing btw!!!!!), will need to be aware when they bump the version over in Community @apanicker-nflx

@james-deee I can always merge this logic into the community edition as well.

james-deee:
@ghoshabhi and @apanicker-nflx when is an ETA that we can expect this to get merged and for a new release of Conudctor to go out? 

I know there is also some batch polling that was added to the Java client that we'd also like to take advantage of. Thanks!

james-deee:
@ghoshabhi or @apanicker-nflx sorry to ping again, just wanted to see if there was an update on this. Thanks!

apanicker-nflx (approved):

PR #3331, opened 2022-11-02
AsyncComplete event task publishing event multiple times.

manan164:
Revert event task-related changes from the PR "Split event system task implementation between start and execute (#3187)"

Pull Request type
----
- [X] Bugfix

Changes in this PR
----

Issue #
Reverting the PR #3187

Alternatives considered
----
1. We can put a condition check so that `EVENT` task execute method will be called once but that will be kind of not proper fix.


apanicker-nflx:
@manan164 To guarantee consistency of event task execution, the splitting of validation and publishing is necessary. By reverting the change, we will run into the race conditions again. We believe it will be best to make the event task an async system task and let this be handled by the AsyncSystemTaskExecutor.

PR #3343, opened 2022-11-14
PX-1012: Fixing the Xstream Library Security leak.

hchepey-clari:
Pull Request type
----
- [x] Bugfix
- [ ] Feature
- [ ] WHOSUSING.md

Changes in this PR
----

This PR is about fixing the security leak vulnerability in the xsteam package which could enable for a potential RCE attack. So, upgraded Netflix Eureka library which was using this xstream package.

Alternatives considered
----

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,7 @@
 buildscript {
 	repositories {
         jcenter()
+        mavenCentral()
     }
 
     dependencies {
@@ -26,10 +27,11 @@ allprojects {
 
     repositories {
         jcenter()
+        mavenCentral()
 
         // oss-candidate for -rc.* verions:
         maven {
-            url "https://dl.bintray.com/netflixoss/oss-candidate"
+            url "https://mvnrepository.com/artifact"
         }
     }
 }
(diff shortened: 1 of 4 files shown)


grpc/build.gradle
```diff
@@ -18,7 +18,6 @@ dependencies {
     compile project(':conductor-common')
     compile project(':conductor-core')
 
-    protobuf 'io.chaossystems.grpc:grpc-healthcheck:1.0.+:protos'
```
Phalgun-Git, line 21: Why did we remove this?

build.gradle
```diff
@@ -1,6 +1,7 @@
 buildscript {
 	repositories {
         jcenter()
+        mavenCentral()
```
Phalgun-Git, line 4: Are these changes the solution to the build issue that you were facing with building the docker image?

Phalgun-Git (commented):
Requesting clarifications.

build.gradle
hchepey-clari: Yes.

grpc/build.gradle
hchepey-clari: this protobuf was hosted at dl.bintray. dl.bintray no longer exists and it is not hosted anywhere else. Hence tried removing that from the project reference and still project build was successful. Hence, I believe, someone added this reference for trying out something and then forgot to remove its reference after the completion of the exercise.

PR #3371, opened 2022-12-01
Support Nack functionality for avoiding loosing messages on errors

AvitalOfstein:
Pull Request type
----
- [X] Feature

Changes in this PR
----

Support Nack functionality for avoiding loosing messages when handling messages and getting unexpected errors

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/events/DefaultEventProcessor.java
+++ b/core/src/main/java/com/netflix/conductor/core/events/DefaultEventProcessor.java
@@ -134,6 +134,10 @@ public class DefaultEventProcessor {
                 queue.publish(Collections.singletonList(msg));
                 LOGGER.debug("Message: {} published to queue: {}", msg.getId(), queue.getName());
             }
+            else {
+                queue.nack(Collections.singletonList(msg));
+                LOGGER.debug("Message: {} nacked on queue: {}", msg.getId(), queue.getName());
+            }
             Monitors.recordEventQueueMessagesHandled(queue.getType(), queue.getName());
         }
     }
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/netflix/conductor/core/events/DefaultEventProcessor.java
```diff
@@ -134,6 +134,10 @@ public void handle(ObservableQueue queue, Message msg) {
                 queue.publish(Collections.singletonList(msg));
                 LOGGER.debug("Message: {} published to queue: {}", msg.getId(), queue.getName());
             }
+            else {
```
manan164, line 137: Hi @AvitalOfstein , Thanks for submitting the PR. Should we also need to change the logic at line [127](https://github.com/Netflix/conductor/pull/3371/files#diff-80301592e5dca75c05d28238ee04eac98b16865d5d42ec57f2f4f32a2435277eL127) to not ack when executionFailed is true?

AvitalOfstein: Hi @manan164, Yes, but we would prefer to split the executionFailed issue handling to different PR (bug fix), from this one (adding Nack feature), since there are already many users of this code, with different queue types, and we wouldn't like to cause potential regression in changing the conditions in this logic in the same PR

aravindanr: hi @AvitalOfstein , could you explain what the `nack` will do? The default implementation is empty. It would also help to which queue implementation are you planning to use `nack` with.

AvitalOfstein: Hi @aravindanr ,
Sure, in case we have an error while processing the event (e.g. line 194), it won't get into the first 2 conditions (queue.ack and queue.publish).
In RabbitMQ implementation, the consumner will do nothing with the message (will be idle) and after some time it will get disconnected.
So in AMQP implementation (community repo) I would like to add the specific implementation of Nack with possibility not to requeue (so it can go to DLQ if defined), other implementations can keep with current behavior.

manan164:
Hi @aravindanr , can you please review the changes? cc @AvitalOfstein

PR #3374, opened 2022-12-06
Do not fail workflow but populate task definition as defaut

v1r3n:
Pull Request type
----
- [ ] Bugfix
- [ ] Feature
- [ ] WHOSUSING.md
- [X] Other (please describe):

Changes in this PR
----
Currently, the workflow is terminated if the task definition is missing in the workflow.  The task definitions are populated at the time of starting a workflow, this fix handles the case where the task was not populated correctly and workflow fails.

Issue #

Alternatives considered
----

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/mapper/SimpleTaskMapper.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/mapper/SimpleTaskMapper.java
@@ -14,7 +14,6 @@ package com.netflix.conductor.core.execution.mapper;
 
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/netflix/conductor/core/execution/mapper/SimpleTaskMapper.java
```diff
@@ -68,18 +67,15 @@ public List<TaskModel> getMappedTasks(TaskMapperContext taskMapperContext)
...
-                        .orElseThrow(
-                                () -> {
-                                    String reason =
-                                            String.format(
-                                                    "Invalid task. Task %s does not have a definition",
-                                                    workflowTask.getName());
-                                    return new TerminateWorkflowException(reason);
-                                });
-
+        TaskDef taskDefinition = workflowTask.getTaskDefinition();
```
aravindanr, line 70: This case seems to be addressed [in MetadataMapperService](https://github.com/Netflix/conductor/blob/main/core/src/main/java/com/netflix/conductor/core/metadata/MetadataMapperService.java#L120-L133). Is there a scenario where this exception happens?

PR #3435, opened 2023-01-10
Make orkes queue default

manan164:
Deprecate dyno queue.

### DIFF
--- a/server/src/main/resources/application.properties
+++ b/server/src/main/resources/application.properties
@@ -23,7 +23,8 @@ conductor.indexing.enabled=false
 #Redis configuration details.
 #format is host:port:rack separated by semicolon
 #Auth is supported. Password is taken from host[0]. format: host:port:rack:password
-conductor.redis.hosts=host1:port:rack;host2:port:rack:host3:port:rack
+#conductor.redis.hosts=host1:port:rack;host2:port:rack:host3:port:rack
+conductor.redis.hosts=localhost:6379:us-east-1c
 
 #namespace for the keys stored in Dynomite/Redis
 conductor.redis.workflowNamespacePrefix=
--- a/server/src/main/resources/application.properties
+++ b/server/src/main/resources/application.properties
@@ -16,6 +16,7 @@ springdoc.api-docs.path=/api-docs
 loadSample=true
 
 conductor.db.type=memory
+conductor.queue.type=redis_standalone
 
 conductor.indexing.enabled=false
 
--- a/README.md
+++ b/README.md
@@ -20,15 +20,15 @@ The latest version is [![Github release](https://img.shields.io/github/v/release
 ## Community Contributions
 The modules contributed by the community are housed at [conductor-community](https://github.com/Netflix/conductor-community). Compatible versions of the community modules are released simultaneously with releases of the main modules.
 
-[Discussion Forum](https://github.com/Netflix/conductor/discussions) Please use the forum for questions and discussing ideas and join the community.
+[Discussion Forum](https://github.com/Netflix/conductor/discussions): Please use the forum for questions and discussing ideas and join the community.
 
-[List of Conductor community projects](/docs/docs/resources/related.md) - Backup tool, Cron like workflow starter, Docker containers...
+[List of Conductor community projects](/docs/docs/resources/related.md): Backup tool, Cron like workflow starter, Docker containers and more.
 
 ## Getting Started - Building & Running Conductor
-### Docker
+###  Using Docker:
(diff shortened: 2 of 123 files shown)


server/src/main/resources/application.properties
```diff
@@ -16,6 +16,7 @@ springdoc.api-docs.path=/api-docs
 loadSample=true
 
 conductor.db.type=memory
+conductor.queue.type=redis_standalone
```
aravindanr, line 19: Does this add redis standalone dependency to starting Conductor?

manan164: yes @aravindanr , this will add the dependency on redis standalone

missedone:
@manan164 , i got the following error when starting conductor on the last main branch, how can I configure `MeterRegistry`?

```
[com/netflix/conductor/core/config/ConductorCoreConfiguration.class]: Unsatisfied dependency expressed through method 'getEventQueueProviders' parameter 0; nested exception is 
org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'conductorEventQueueProvider' defined in URL [jar:file:/Users/xuqingtan/workspace/platform/conductor/core/build/libs/conductor-core-3.14.0-SNAPSHOT.jar!/com/netflix/conductor/core/events/queue/ConductorEventQueueProvider.class]: Unsatisfied dependency expressed through constructor parameter 0; nested exception is 
org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'getQueueDAOStandalone' defined in class path resource [io/orkes/conductor/queue/config/RedisQueueConfiguration.class]: Unsatisfied dependency expressed through method 'getQueueDAOStandalone' parameter 1; nested exception is 
org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'io.micrometer.core.instrument.MeterRegistry' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {}
```

manan164:
Hi @missedone , Thanks for reporting. Can you please share the properties used other than those mentioned above? I tried but could not be able to reproduce it.

missedone:
i tried with the following properties as well, but didn't work either:
```
management.endpoint.metrics.enabled=true
management.endpoints.web.exposure.include=*
management.metrics.export.simple.enabled=true
```

missedone:
thanks @manan164  for the help, the issue might be caused by the dirty cache in IntelliJ.
it works once I updated to springboot 2.7.9, but don't think the version matters, as it works still if I revert the springboot version to 2.7.3.

manan164:
Thanks @missedone , glad the issue is solved.

server/src/main/resources/application.properties
chidov: @manan164 is there a chance you could update this [doc](https://conductor.netflix.com/devguide/running/source.html) to mention we need to run redis standalone before we start server locally?

lijia-rengage: if i didn't define this config `conductor.queue.type`, will I still use the orkes_queue?

manan164: No @lijia-rengage , We need to mention the queue type otherwise it will use dynomite.

lijia-rengage:
I got this error when I deployed dynomite and conductor-server on Kubernetes.
MDRT1Zm7jO
This is my config:
```
# Servers.
conductor.grpc-server.enabled=false

# Database persistence model.
conductor.db.type=redis_standalone

# Dynomite Cluster details.
# format is host:port:rack separated by semicolon
conductor.redis.hosts=redis.conductor.svc.cluster.local:6379:us-west-2a

# Namespace for the keys stored in Dynomite/Redis
conductor.redis.workflowNamespacePrefix=conductor

# Namespace prefix for the dyno queues
conductor.redis.queueNamespacePrefix=conductor_queues

# No. of threads allocated to dyno-queues (optional)
queues.dynomite.threads=10

# By default with dynomite, we want the repairservice enabled
conductor.app.workflowRepairServiceEnabled=true

# Non-quorum port used to connect to local redis.  Used by dyno-queues.
# When using redis directly, set this to the same port as redis server
# For Dynomite, this is 22122 by default or the local redis-server port used by Dynomite.
conductor.redis.queuesNonQuorumPort=6379

# Elastic search instance indexing is disabled.
conductor.indexing.enabled=true
conductor.elasticsearch.url=http://elasticsearch.conductor.svc.cluster.local:9200
conductor.elasticsearch.indexReplicasCount=0

# Load sample kitchen sink workflow
loadSample=true

conductor.elasticsearch.clusterHealthColor=yellow
```
Could you plz give my some insights?

PR #3444, opened 2023-01-13
[#3442] fix: can't click task card to open the right panel while using FireFox

nhandt2021:
Pull Request type
----
- [x] Bugfix

Changes in this PR
----

Issue #3442

User can't click on task at flow diagram to open the right side panel while using FireFox browser.

https://user-images.githubusercontent.com/83625759/212276351-150a8a03-3f0f-4772-b79f-003b7c49e8c4.mov

### DIFF
--- a/ui/src/components/diagram/WorkflowGraph.jsx
+++ b/ui/src/components/diagram/WorkflowGraph.jsx
@@ -338,8 +338,21 @@ class WorkflowGraph extends React.Component {
     inner.selectAll("g.node").on("click", this.handleClick);
   };
 
+  /**
+   * Get the taskRef id base on browsers
+   * @param e
+   * @returns {string | undefined} The id of the task ref
+   */
+  getTaskRef = (e) => {
+    const isFirefox = navigator.userAgent.toLowerCase().indexOf("firefox") > -1;
+    if (isFirefox) {
+      return e.target?.parentNode?.id;
+    }
+    return e?.path[1]?.id || e?.path[2]?.id; // could be 2 layers down
+  };
+
   handleClick = (e) => {
-    const taskRef = e.path[1].id || e.path[2].id; // could be 2 layers down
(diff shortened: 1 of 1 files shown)


ui/src/components/diagram/WorkflowGraph.jsx
```diff
@@ -338,8 +338,21 @@ class WorkflowGraph extends React.Component {
...
   };
 
+  /**
+   * Get the taskRef id base on browsers
+   * @param e
+   * @returns {string | undefined} The id of the task ref
+   */
+  getTaskRef = (e) => {
+    const isFirefox = navigator.userAgent.toLowerCase().indexOf("firefox") > -1;
+    if (isFirefox) {
```
arketec, line 348: This fix should also work in the latest version of Chrome:
image

ui/src/components/diagram/WorkflowGraph.jsx
```diff
@@ -338,8 +338,21 @@ class WorkflowGraph extends React.Component {
...
+      return e.target?.parentNode?.id;
+    }
+    return e?.path[1]?.id || e?.path[2]?.id; // could be 2 layers down
+  };
+
   handleClick = (e) => {
-    const taskRef = e.path[1].id || e.path[2].id; // could be 2 layers down
+    const taskRef = this.getTaskRef(e);
```
ceisen1, line 355: I believe this could be made simpler with `event.composedPath`. Chrome recommends updating to this for the now deprecated `event.path` API https://chromestatus.com/feature/5726124632965120.

It also seems to be compatible with all major browsers https://developer.mozilla.org/en-US/docs/Web/API/Event/composedPath#browser_compatibility

ceisen1: added the fix here https://github.com/Netflix/conductor/pull/3459

PR #3484, opened 2023-02-08
Task reset feature

manan164:
Pull Request type
----
- [X] Feature

Introduced a new feature called reset tasks.
Tasks can be reset and will get executed again. This is like restarting at task level. Multiple tasks can be restarted together. The task which comes first in the schedule will get executed first.
Workflow execution will proceed only when all reset tasks are completed.


manan164:
Comment to make PR alive.

PR #3492, opened 2023-02-15
Add spring support for annotated workers

v1r3n:
Pull Request type
----
- [ ] Bugfix
- [x] Feature
- [ ] WHOSUSING.md

Changes in this PR
----
1. Make it possible to add `@Component` to the classes that has worker methods and they will get auto-initialized.
2. Allow more control in the annotated workers so users can use `callbackAfter` for long running tasks
3. New exception that allows user to indicate the task has failed and should not be retried
4. Ability to control polling interval - useful for long running tasks
5.  `TaskContext` allows getting access to the task and task result for inspecting task level information

Issue #

Alternatives considered
----

### DIFF
--- a/client-spring/src/main/java/com/netflix/conductor/client/spring/ConductorWorkerAutoConfiguration.java
+++ b/client-spring/src/main/java/com/netflix/conductor/client/spring/ConductorWorkerAutoConfiguration.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2023 Netflix, Inc.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package com.netflix.conductor.client.spring;
+
+import java.util.Map;
+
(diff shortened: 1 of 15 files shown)


client-spring/src/main/java/com/netflix/conductor/client/spring/ConductorWorkerAutoConfiguration.java
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright 2020 Netflix, Inc.
```
mytang0, line 2: 2020 or 2023?

client-spring/src/main/java/com/netflix/conductor/client/spring/ConductorWorkerAutoConfiguration.java
```diff
@@ -0,0 +1,48 @@
...
+public class ConductorWorkerAutoConfiguration {
+
+    @Autowired private ApplicationContext applicationContext;
+
+    @Autowired private TaskClient taskClient;
+
+    @ConditionalOnMissingBean
+    @Bean
+    public AnnotatedWorkerExecutor annotatedWorkerExecutor(ClientProperties clientProperties) {
+
```
mytang0, line 37: Do we need to make sure that this bean is created later than other beans to avoid omissions?

client-spring/src/main/java/com/netflix/conductor/client/spring/ConductorWorkerAutoConfiguration.java
v1r3n: Fixed.  Thanks for pointing out.

client-spring/src/main/java/com/netflix/conductor/client/spring/ConductorWorkerAutoConfiguration.java
v1r3n: Addressed this by doing the initialization using context initialization event.

Issue #3494, opened 2023-02-16
Complete WAIT task inside for loop by TaskRef


astelmashenko:
See discussion https://github.com/Netflix/conductor/discussions/2873 
Seems like it was partially fixed https://github.com/Netflix/conductor/pull/2883 for eventhandler actions, but did not get fixed for default WAIT task handlers.

astelmashenko:
@manan164 , could you please review?

astelmashenko:
@manan164 , There is already PR, please see https://github.com/Netflix/conductor/pull/3495

PR #3495, opened 2023-02-16
removed task ref name when processing event queue

astelmashenko:
Pull Request type
----
- [x] Bugfix

----
Fixes https://github.com/Netflix/conductor/issues/3494

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/events/queue/DefaultEventQueueProcessor.java
+++ b/core/src/main/java/com/netflix/conductor/core/events/queue/DefaultEventQueueProcessor.java
@@ -14,6 +14,7 @@ package com.netflix.conductor.core.events.queue;
 
 import java.util.*;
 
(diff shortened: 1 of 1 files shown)


manan164:
Hi @astelmashenko , Changes looks good. Can you please add a small unit test around this? Please check [reference](core/src/test/java/com/netflix/conductor/core/events/TestSimpleActionProcessor.java)

astelmashenko:
@manan164 , I quickly checked the reference test. DefaultEventQueueProcessor is wirtten a way it will require a lot of ceremony to setup a test. To make it easier to test DefaultEventQueueProcessor should be refactored e.g. into testable service with logic to process events and processor with code that subscribes to observable/etc.
What do you think about the reafactoring?

astelmashenko:
@manan164 , why do not we merge this?

v1r3n (approved):

PR #3544, opened 2023-03-22
ci: use Temurin distributions

yeikel:
Host runners include Temurin by default as part of the [hosted tool cache ](https://github.com/actions/setup-java/blob/main/docs/advanced-usage.md#hosted-tool-cache)

Using Temurin speeds up builds as there is no need to download and configure the Java SDK with every build.

### DIFF
--- a/.github/workflows/update-gradle-wrapper.yml
+++ b/.github/workflows/update-gradle-wrapper.yml
@@ -13,7 +13,7 @@ jobs:
       - name: Set up Zulu JDK 11
         uses: actions/setup-java@v3
         with:
-          distribution: 'zulu'
+          distribution: 'temurin'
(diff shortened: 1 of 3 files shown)


.github/workflows/update-gradle-wrapper.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       - name: Set up Zulu JDK 11
         uses: actions/setup-java@v3
         with:
-          distribution: 'zulu'
+          distribution: 'temurin'
```
mytang0, line 16: The 'name' also needs to be changed.

image

yeikel: Thank you. I updated it

PR #3565, opened 2023-04-04
elastic search with postgres.

manan164:
Changing the docker file to use the elastic search with postgres and Redis.  The image will contain ES and Reids and based on configuration either both or none or one is required.


thxmike (approved):
OK

v1r3n (approved):

PR #3573, opened 2023-04-11
Fix disappearing outputs from long running tasks

marosmars:
with external storage enabled

In case of a:
1. Long running task
2. With big output (externalized)
3. With output growing over time
4. Causing multiple externalize / internalize executions
5. ... such as a join task collecting outputs of all forked tasks
6. Lost some of its outputs when finally completed

This issue was caused by / because:
1. On an Nth execution of a task (such as described above)
2. The task internalized its intermediate output from external storage
3. The task was executed and it updated its output to current value in memory
4. The task tried to externalize the new version of its output
5. ... but while doing so, the outputPayload (last externalized value) was combined with outputData (current, in-memory value) in a way where output payload over-wrote the latest values
6. Thus, newly calculated outputs have been lost

Pull Request type
----
- [x] Bugfix
- [ ] Feature
- [ ] WHOSUSING.md

Changes in this PR
----

Issue #

Alternatives considered
----

### DIFF
--- a/core/src/main/java/com/netflix/conductor/model/TaskModel.java
+++ b/core/src/main/java/com/netflix/conductor/model/TaskModel.java
@@ -398,7 +398,12 @@ public class TaskModel {
     @JsonIgnore
     public Map<String, Object> getOutputData() {
         if (!outputPayload.isEmpty() && !outputData.isEmpty()) {
-            outputData.putAll(outputPayload);
+            // Combine payload + data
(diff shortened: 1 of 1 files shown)


marosmars:
@v1r3n hey, we found this issue when running with external storage. Please take a look, it's just a 1 liner

v1r3n (approved):

PR #3585, opened 2023-04-19
Handling messages loss case on exceptions

AvitalOfstein:
Pull Request type
----
- [ X ] Bugfix

Changes in this PR
----

When message has no transient failures, only logically unhandled exception - message was being Acked. In our case - message that it's handler metadata wasn't fetched successfully (for example, Connection with Redis was down).
In this case we would expect it to-
 1.Be retried (transient failure) - Handled with adding transient exception catching in RedisEventHandlerDao 
 2. In general, we would like only messages that passed successfully, without any type of exception - will be Acked. Messages that we want by config to be republished on failure OR have exceptions that are transient - to be published back to queue. Only other failing messages - to be Nacked.


v1r3n:
@AvitalOfstein can you run the spotless and update the PR?

PR #3594, opened 2023-04-26
Join task callBackDuration

manan164:
Pull Request type

Changes in this PR
----
The implemented approach of exponential backoff in join task calbackDuration.

Issue #
https://github.com/Netflix/conductor/discussions/3436

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/tasks/Join.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/tasks/Join.java
@@ -13,6 +13,7 @@
 package com.netflix.conductor.core.execution.tasks;
 
 import java.util.List;
+import java.util.Optional;
 import java.util.stream.Collectors;
 
 import org.springframework.stereotype.Component;
@@ -96,6 +97,15 @@ public class Join extends WorkflowSystemTask {
             return true;
         }
         return false;
+    }
+
+    @Override
+    public Optional<Long> getEvaluationOffset(TaskModel taskModel, long defaultOffset) {
+        int index = taskModel.getPollCount() > 0 ? taskModel.getPollCount() - 1 : 0;
+        if (index == 0) {
+            return Optional.of(0L);
+        }
+        return Optional.of(Math.min((long) Math.pow(2, index), defaultOffset));
     }
 
     public boolean isAsync() {
--- a/core/src/main/java/com/netflix/conductor/core/execution/AsyncSystemTaskExecutor.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/AsyncSystemTaskExecutor.java
@@ -153,11 +153,14 @@ public class AsyncSystemTaskExecutor {
                 LOGGER.debug("{} removed from queue: {}", task, queueName);
             } else {
                 task.setCallbackAfterSeconds(systemTaskCallbackTime);
+                systemTask.getEvaluationOffset(task, systemTaskCallbackTime)
+                        .ifPresentOrElse(task::setCallbackAfterSeconds,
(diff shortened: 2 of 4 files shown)


apanicker-nflx (approved):

v1r3n:
cc: @james-deee

core/src/main/java/com/netflix/conductor/core/execution/tasks/Join.java
```diff
@@ -98,6 +99,15 @@ public boolean execute(
         return false;
     }
 
+    @Override
+    public Optional<Long> getEvaluationOffset(TaskModel taskModel, long defaultOffset) {
+        int index = taskModel.getPollCount() > 0 ? taskModel.getPollCount() - 1 : 0;
+        if (index == 0) {
+            return Optional.of(0L);
+        }
+        return Optional.of(Math.min((long) Math.pow(2, index), defaultOffset));
```
james-deee, line 108: What happens if someone has their systemTaskWorkerCallbackDuration to one second though? It will always still be 1 second, so this won't be an exponential backoff in that scenario. It feels like `defaultOffset` should be a configurable maximum offset or something like that. Maybe the default 30 seconds?

manan164: Hi @james-deee , I think if we are setting systemTaskWorkerCallbackDuration to 1 second then ideally we don't want the exponential backoff thing. Here we are trying to exponential backoff from 1 to 30 seconds.

james-deee: One hand, I see what you're saying, but in *practice* what we are seeing is that having 1 second doesn't alleviate the problem. The issue is that we have a lot of long running workflows (thus JOINs) in the queue.  If we do NOT do an exponential backoff from our setting of 1 second, then our JOIN tasks don't get processed fast enough because the async task worker is trying to process say 10ks JOINs.......... which then makes JOINs that you want to process immediately be backed up and slowed down.

With my proposal, and with what we changed locally, fixes all of our issues with processing the async JOINS. 

In our situation we use a duration of 1 second, but we have a cap of 30 seconds in place of that `defaultOffset` and it completely solves our issue. The result of it is that JOINs on fast running workflows get processed almost instantaneously, and JOINs in the long running workflows get pushed back to the max of 30 seconds..... which is what frees up the processing of the immediate ones.

I really think these are 2 different levers/values that need to be present. Just using 1 second always won't work in heavy workfload Conductor instances where there might be a LOT of joins to process.

james-deee: Or maybe what you're saying is that we should be using a setting of 30 seconds for the `systemTaskWorkerCallbackDuration` always, and then it sounds like no one should ever use 1 second for the `systemTaskWorkerCallbackDuration` value.

Then, I think that this could get confusing, because it would feel like someone would want to set the value to 1 for fast processing, but they wouldnt know that actually using 30 seconds will get the fastest/best processing?

james-deee: @manan164 maybe a compromise is to use a hardcoded 30 seconds for that `defaultOffset`. I really think it will be confusing and could cause issues for people to use 1 second for that setting then, because that doesn't actually give them the best performance.

shekarpcs:
is it merged to the main?

franrull:
Any estimate on when this can be expected to be merged? It is crippling the performance of our workflows 🙏🏼

shekarpcs:
Exactly, our system is throttled. Database is also at high utilization.

PR #3602, opened 2023-05-05
Accept skipTaskRequest payload in conductor client skipTaskFromWorkfow api

supriyanta:
Pull Request type
----
- [ ] Bugfix

Changes in this PR
----
Add skipTaskRequest payload in conductor client skipTaskFromWorkfow api

Issue 
----
Conductor client skipTaskFromWorkfow api doesn't accept skipTaskRequest payload


client/src/main/java/com/netflix/conductor/client/http/WorkflowClient.java
```diff
@@ -357,14 +358,15 @@ public void resumeWorkflow(String workflowId) {
      * @param workflowId the id of the workflow instance
      * @param taskReferenceName the reference name of the task to be skipped
      */
-    public void skipTaskFromWorkflow(String workflowId, String taskReferenceName) {
```
v1r3n, line 360: can you create an overloaded method and deprecate the existing method so we do not break backward compatibility?

supriyanta: yes, sure

v1r3n (approved):

PR #3146, opened 2022-08-03
do not synchronously eval workflow during update task for select criteria

apanicker-nflx:
Pull Request type
----
- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes (Please run `./gradlew generateLock saveLock` to refresh dependencies)
- [ ] WHOSUSING.md
- [ ] Other (please describe):

**NOTE**: Please remember to run `./gradlew spotlessApply` to fix any format violations.

Changes in this PR
----

_Describe the new behavior from this PR, and why it's needed_
Issue #

Alternatives considered
----

_Describe alternative implementation you have considered_

PR #3640, opened 2023-06-07
Ensure that JOIN tasks wait for all retries before terminating

bjpirt:
Credit to @james-deee who authored the patch that this PR implements.

This fixes the bug where a JOIN wasn't waiting for all of the retries for a task to have been performed

Pull Request type
----
- [x] Bugfix

Changes in this PR
----
This fixes the bug where a JOIN wasn't waiting for all of the retries for a task to have been performed

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/tasks/Join.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/tasks/Join.java
@@ -56,7 +56,11 @@ public class Join extends WorkflowSystemTask {
                 break;
             }
             TaskModel.Status taskStatus = forkedTask.getStatus();
-            hasFailures = !taskStatus.isSuccessful() && !forkedTask.getWorkflowTask().isOptional();
+            hasFailures =
+                    !forkedTask.isStillInNeedOfProcessing()
+                            && !taskStatus.isSuccessful()
+                            && !forkedTask.getWorkflowTask().isOptional();
+
             if (hasFailures) {
                 failureReason.append(forkedTask.getReasonForIncompletion()).append(" ");
             }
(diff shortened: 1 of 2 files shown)


manan164:
Hi @bjpirt , Changes look fine. Can you please write few tests around this change?

core/src/main/java/com/netflix/conductor/core/execution/tasks/Join.java
```diff
@@ -56,15 +56,19 @@ public boolean execute(
                 break;
             }
             TaskModel.Status taskStatus = forkedTask.getStatus();
-            hasFailures = !taskStatus.isSuccessful() && !forkedTask.getWorkflowTask().isOptional();
+            hasFailures =
```
v1r3n, line 59: the code at line 52 should already fetch the join task that is retried, so it will be in progress and will wait for it to complete.

core/src/main/java/com/netflix/conductor/model/TaskModel.java
```diff
@@ -587,6 +587,19 @@ public void setWaitTimeout(long waitTimeout) {
...
     }
 
+    /**
+     * If the task is in a failed state, we need to make sure that all retries have been attempted
+     * for the task. If not, then the task is not in a "final" terminal state yet.
+     */
+    @JsonIgnore
+    public boolean isStillInNeedOfProcessing() {
+        if (Status.FAILED.equals(getStatus())) {
+            final int taskDefRetryCount = getTaskDefinition().map(TaskDef::getRetryCount).orElse(0);
```
manan164, line 597: This can create NPE when task definition is not present.

manan164:
Hi @bjpirt , I have raised a fix for this change. https://github.com/Netflix/conductor/pull/3655. 
Ideally, changing the task model is not a good idea. Since this behavior was introduced in https://github.com/Netflix/conductor/pull/3146 we should fix that.

bjpirt:
Closing in favour of https://github.com/Netflix/conductor/pull/3655

Issue #1982, opened 2020-11-24
How to use a dynamic subworkflow?


smurakozi:
According to the [doc of SubWorkflow](https://netflix.github.io/conductor/configuration/systask/#sub-workflow) the workflow parameter "Allows starting a subworkflow with a dynamic workflow definition.". This is the solution that andyyumiao recommended to [make dynamic fork serializer](https://github.com/Netflix/conductor/issues/1588).

I see how you can inline a workflow def but I have no clue how to make it dynamic. For me dynamic means you should be able to define the subworkflow "on the fly", when the parent workflow/the subworkflow task is running.
 
I tried to use the output of a previous task, but when I try to create the main workflow (PUT metadata/workflow) it's rejected with an error `INTERNAL_ERROR - Cannot construct instance of `com.netflix.conductor.common.metadata.workflow.WorkflowDef` (although at least one Creator exists): no String-argument constructor/factory method to deserialize from String value`. 
I tried to find related doc pages (e.g. in kitchensink) issues, stackoverflow questions..., but I couldn't find any examples. I also checked the code of Conductor but found no hint how to do this.

Are there any examples how to use a dynamic subworkflow? Am I misunderstanding what the doc means on a dynamic workflow?

Background: I need to process a long list of items in batches. My idea was to create a task for each batch, and to create these tasks as a dynamic sub workflow.
I also considered using a dynamic forkjoin, but I have to process the batches in a sequential fashion. I also checked the do-while task, but it would restart all batches in case of an error + `Do while task does NOT support domain or isolation group execution` (that we plan to use).

Is there another alternative I did not consider?

These are the tasks in my latest attempt to create the sub workflow dynamically:
```
  {
    "name": "create_batches",
    "taskReferenceName": "create_batches",
    "inputParameters": {
    },
    "type": "SIMPLE"
  },
  {
    "name": "sub_workflow_task",
    "taskReferenceName": "sub_workflow_task",
    "type": "SUB_WORKFLOW",
    "inputParameters": {
      "workflowDefinition": "${create_batches.output}"
    },
    "subWorkflowParam": {
      "name": "sub_workflow",
      "version": 1,
      "taskToDomain": {
        "*": "mydomain"
      },
      "workflowDefinition": "workflowDefinition"
    }
  }
```
The output of the first task is a workflow definition with name sub_workflow. I also tried to use this output directly in subWorkflowParam.workflowDefinition, with the same result. 

Thanks,
Sandor

[...]

PR #3648, opened 2023-06-09
Add support for subworkflow tasks which have SubworkflowParam's workflowDef inline

c4lm:
Pull Request type
----
- [x] Bugfix

Changes in this PR
----

Currently unsupported.
Issue [#1982](https://github.com/Netflix/conductor/issues/1982)
This PR will allow setting subWorkflowParams like this:
```
"subWorkflowParam": {
        "workflowDefinition": "${workflow.input.tst}"
      }
```
so that parameter resolution will happen, and workflowDefinition, if available, will take priority over name.

Alternatives considered
----

The alternative was to keep everything in place, but alter SubWorkflowTaskMapper and SubWorkflow so that name/version/definition resolution actually comes from task input instead of `subWorkflowParams`.


c4lm:
Do not merge yet. There are some unhandled (de)serialization cases.

PR #3695, opened 2023-07-15
Updates for podman

jd-brennan:
Changes needed to run with podman:

1) don't use "links" in docker-compose.yaml (podman doesns't support this)
2) use arm version of elasticsearch (not podman related)
3) use node 16 (not podman related)

Steps to set up:
```
brew install podman-desktop
# Default of -m 2048 is not enough for Conductor containers (so use 5GB)
podman machine init -m 5120 --disk-size 200 --cpus 2 
sudo podman-mac-helper install # make podman engine compatible with `docker` command!
podman machine start

brew install docker
brew install docker-compose
brew install docker-credential-helper
```

Running conductor:
```
cd conductor
cd docker
# Can't build two images at once - it causes a race condition failure, so build UI first
docker-compose build conductor-ui 
docker-compose up -d
```

### DIFF
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -18,9 +18,6 @@ services:
       interval: 60s
       timeout: 30s
       retries: 12
-    links:
-      - elasticsearch:es
-      - redis:rs
     depends_on:
       elasticsearch:
         condition: service_healthy
@@ -44,8 +41,6 @@ services:
       - internal
     ports:
       - 5000:5000
-    links:
-      - conductor-server
     stdin_open: true
 
   redis:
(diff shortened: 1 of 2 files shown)


docker/docker-compose.yaml
```diff
@@ -44,33 +41,35 @@ services:
...
     healthcheck:
       test: [ "CMD", "redis-cli","ping" ]
 
   elasticsearch:
-    image: elasticsearch:6.8.15
+    image: jazzdev/elasticsearch:6.8.15-arm
     container_name: elasticsearch
     environment:
-      - "ES_JAVA_OPTS=-Xms512m -Xmx1024m"
+      - "ES_JAVA_OPTS=-Xms1024m -Xmx1024m"
```
jd-brennan, line 63: elasticsearch outputs a warning if these aren't the same value.
Could set them both to 512 and maybe run the VM with only 4GB instead of 5GB.

jd-brennan:
I didn't mean to make this public. Closing.

Issue #3741, opened 2023-08-21
[BUG]: ./gradlew: not found error when trying to build and run docker compose on Windows


dhruv2196:
## What are you missing in the docs
When we try to build docker image for conductor server on Windows, it throws -
> [conductor:server builder 4/4] RUN ./gradlew build -x test --stacktrace:
#24 0.398 /bin/sh: 1: ./gradlew: not found
So I found a solution to do that and I have added steps for the same below.
#### ./gradlew: not found when running on Windows 
1. Switch the line endings to "LF" from "CLRF" in gradlew file. 
In IntelliJ - From the main menu, select File | File Properties | Line Separators, and then select LF - Unix and MacOS(\n).
In Eclipse - Go to Window | Preferences | General | Workspace | New Text file line delimiter -> Other -> Unix
2. From the `docker` directory,
    ```
    docker build --no-cache -t conductor:server -f server/Dockerfile ../

dhruv2196:
I have added a solution for it. Please let me know if I can assign it to myself and get it merged.

dhruv2196:
I found another similar issue that when we are trying to build docker compose on windows. I was facing an error ./gradlew: not found.
I fixed it by changing the line seperator in gradlew so docker build was able to built it successfully but when I tried to run the server it was again throwing some exceptions. Reason for this issue is that when we clone code in windows machine it by default uses CRLF as line separator and due to this all scripts have CRLF line separators as EOL and they broke.

PR #3742, opened 2023-08-21
Added support for building and running docker image for windows in documentation.

dhruv2196:
Pull Request type
----
- [x] Bugfix and Documentation for Windows

Changes in this PR
----
Added support for windows in documentation 

Issue #3741
./gradlew: not found error when trying to build docker compose on Windows

### DIFF
--- a/docs/docs/gettingstarted/docker.md
+++ b/docs/docs/gettingstarted/docker.md
@@ -153,6 +153,11 @@ Reference: [Issue 2262][issue2262]
 #### Changes in config-*.properties do not take effect
 Config is copy into image during docker build. You have to rebuild the image or better, link a volume to it to reflect new changes.
 
+#### ./gradlew: not found when running on Windows 
+Switch the line endings to "LF" from "CLRF" in gradlew file. 
+In IntelliJ - From the main menu, select File | File Properties | Line Separators, and then select LF - Unix and MacOS(\n).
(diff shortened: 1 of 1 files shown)


dhruv2196:
I have updated the solution as I found another issue when trying to build and run docker compose on windows. 
Initial issue - I was facing an error ./gradlew: not found.
I fixed it by changing the line seperator. So it was able to built it successfully but when I tried to run the server it was again throwing some exceptions. So I found a solution to run these 2 commands and then clone it -
Run git config --global core.autocrlf false;
Run git config --global core.eol lf; (Small L and f)
git clone https://github.com/Netflix/conductor.git
The problem was that when I was trying to clone it on windows then it automatically picks all the files with CRLF line separator. So we need to change it to LF so that scripts can run smoothly via windows on docker.

dhruv2196:
Can anyone please review this PR?

Robban1980:
Running 
```
Run git config --global core.autocrlf false;
Run git config --global core.eol lf; (Small L and f)
```
on windows is not good as it changes the global settings for all repos and will cause issues for any other repo

in a `.gitattributes` file add the following
`gradlew eol=lf
*.gradle eol=lf
*.java eol=lf
*.groovy eol=lf
spring.factories eol=lf
*.sh eol=lf`

then the line endings will not change when the code is pulled.

dhruv2196:
> 

Thanks @Robban1980 for reviewing this. Your suggestion sounds better to me. I will make changes, test and update the PR accordingly.

PR #3745, opened 2023-08-22
security fix for missing env variable replacement

egandro:
Pull Request type
----
- [x] Bugfix

Changes in this PR
----

The current implementation of the Conductor properties file reader, didn't substitute environment variables.
This is a major security and design flaw leading to bad setups and insecure installations in the field.

Alternatives considered
----

The properties system is a bit outdated and requires some work. For now we go with the security fix.

### DIFF
--- a/server/src/main/java/com/netflix/conductor/Conductor.java
+++ b/server/src/main/java/com/netflix/conductor/Conductor.java
@@ -14,6 +14,8 @@ package com.netflix.conductor;
 
 import java.io.IOException;
 import java.util.Properties;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


v1r3n:
This approach is useful when constructing strings with env variables which is not the use case here.  A typical k8s based deployment environment supports sending environment variables which is a recommended way to handle deployments for production use cases.

egandro:
> This approach is useful when constructing strings with env variables which is not the use case here. A typical k8s based deployment environment supports sending environment variables which is a recommended way to handle deployments for production use cases.

I agree and that is what I am using in my Docker containers. 

https://github.com/egandro/conductor-containers/blob/e4583d0dbbeab494a2440a6d4ad7cc9912362ecc/docker/bin/startup.sh#L23

I specifically disable the `CONDUCTOR_CONFIG_FILE` and go with a `--spring.config.location=/app/config/application.properties`.

However - you need to make a decission. If you want to have  `CONDUCTOR_CONFIG_FILE` either fix it the way I suggest - or - explicitly use a piece of documentation and make it a major security issue to use in production enviornment (that part of the documentation is really missing) the docker-compose files in this repo are even encuraging people going the super super insecure was - if - they are not Java Spring experts.

Fix it - document it - or delete it! Even for development mode it make no sense to have a settings and some "advanced light settings overirde - in a bad security style" enabled.

PR #3770, opened 2023-09-12
Allow for version upgrading of log4j2

james-deee:
Pull Request type
----
- [X] Bugfix

Changes in this PR
Allow for Log4j upgrading by regenerating the lock files. The strictly version will still ensure that the versions are within the range.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -98,8 +98,6 @@ allprojects {
     dependencies {
         implementation('org.apache.logging.log4j:log4j-core') {
             version {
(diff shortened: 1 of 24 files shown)


build.gradle
```diff
@@ -98,44 +98,34 @@ allprojects {
     dependencies {
         implementation('org.apache.logging.log4j:log4j-core') {
             version {
-                // this is the preferred version this library will use
```
james-deee, line 101: This was preventing log4j from actually being upgraded

PR #3774, opened 2023-09-14
Unlock Common and Client from the log4j2 version locking for Consumers

james-deee:
Pull Request type
----
- [X] Bugfix
- [X] Build related changes (Please run `./gradlew generateLock saveLock` to refresh dependencies)

Changes in this PR
----
This does a couple of things. First, it runs an update on the locks, which upgraded a couple of other items. Next, it removes the lock of log4j for both the conductor-client and conductor-common modules. These are used by consumers, and can cause all sorts of havoc if we lock a version. Consumers should be able to upgrade log4j as they see fit, and I believe that this will still keep the strictly restriction in place for Conductor itself, so no one would be able to set a version below 2.17.2 in Conductor itself.

### DIFF
--- a/client/build.gradle
+++ b/client/build.gradle
@@ -15,6 +15,16 @@ configurations.all {
     exclude group: 'amazon', module: 'aws-java-sdk'
 }
 
+dependencyLock {
+    skippedDependencies = [
+            "org.apache.logging.log4j:log4j-api",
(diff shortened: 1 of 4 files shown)


james-deee:
Don't merge this, I'm not sure this solves the consumer strict issue

Issue #3403, opened 2022-12-13
Wait takes much longer than expected


Nostras:
**Describe the bug**
Wait duration is precise, but highly inaccurate. Setting a duration of 3 seconds actually results in a delay of 30 seconds. A delay of 30 seconds is actually 115 seconds. These results are consistent.

**Details**
Conductor version: 3.13.2
Persistence implementation: Postgres
Queue implementation: NA (stock or postgres)
Lock: NA (stock)
`
    {
      "name": "Internal wait",
      "taskReferenceName": "Internal_wait",
      "inputParameters": {
        "duration": "30 seconds"
      },
      "type": "WAIT",
      "startDelay": 0,
      "optional": false,
      "asyncComplete": false
    },
`

**To Reproduce**
Just create a workflow with a WAIT type and see it take longer than expected.

**Expected behavior**
The behaviour matches the duration specified.

**Screenshots**
30 seconds:
![image](https://user-images.githubusercontent.com/4203238/207400571-1132ff45-70d7-467a-9c8a-aa1ffd016f76.png)
Actual duration:
![image](https://user-images.githubusercontent.com/4203238/207400621-dad9389b-41e7-41b2-ad13-5aec054c5363.png)

I'm also having an issue where this WAIT is ignored while in a DO_WHILE loop, but I'm not sure if this is related or an issue at all. Still trying to identify it. 
![image](https://user-images.githubusercontent.com/4203238/207400863-6745a072-2d1d-4ba1-9a62-ccc08d564eba.png)

boney9:
cc: @v1r3n

boney9:
Hi @Nostras - this is sort of by design. The WAIT tasks are completed by the decision queues, which is scheduled once every 30 seconds with the default configurations. There is an improvement coming for this, but it's not out yet.  cc: @v1r3n

Nostras:
> Hi @Nostras - this is sort of by design. The WAIT tasks are completed by the decision queues, which is scheduled once every 30 seconds with the default configurations. There is an improvement coming for this, but it's not out yet. cc: @v1r3n

Hi Boney, could you please point out where to change this? If need be I can edit the source code too.
To add to the issue with the wait in switch, it seems that the switch is spawning a WAIT which is then orphaned and ignored (except that it shows up in the task list). Adding a join outside of the switch resolves this, but this breaks pretty badly as the join spins infinitely if the wait was not executed/created, which defeats the purpose.
[...]

PR #3775, opened 2023-09-15
Make wait task async.

manan164:
Pull Request type
----
- [X] Feature

Changes in this PR
----
Currently, the waiting task can not be less than 30 seconds because the wait task relies on the sweeper to get completed and the sweeper runs every 30 seconds for the workflow. 
Now we can decrease the `workflowOffsetTimeout` to some lower value so that workflow will get evaluated faster. 
However, this approach will put a lot of load on the platform since workflows are evaluated faster.
This PR changes the wait task from sync to async mechanism. 

Issue #
https://github.com/Netflix/conductor/issues/3403
Alternatives considered
----
Using workflowOffsetTimeout with a lower value.


v1r3n (approved):

core/src/main/java/com/netflix/conductor/core/execution/mapper/WaitTaskMapper.java
```diff
@@ -69,6 +79,52 @@ public List<TaskModel> getMappedTasks(TaskMapperContext taskMapperContext) {
...
+            long seconds = timeDuration.getSeconds();
+            task.setCallbackAfterSeconds(seconds);
+
+        } else if (StringUtils.isNotBlank(until)) {
+            try {
+
+                Date expiryDate = parseDate(until);
+                long timeInMS = expiryDate.getTime();
+                long now = System.currentTimeMillis();
+                long seconds = ((timeInMS - now) / 1000) + 1;
```
Young-Zen, line 113: @manan164 Hello, I don't understand why we need to add 1 to the "seconds" variable. It seems to occasionally cause testWaitUntil to fail: https://github.com/Netflix/conductor/actions/runs/6294685428/job/17086871264?pr=3779

manan164: Hi @Young-Zen , good point.  I think it may be for rounding purposes. I will change this. Thanks for pointing this out.

manan164: Hi @Young-Zen , I have raised fix for this, https://github.com/Netflix/conductor/pull/3786

PR #1113, opened 2019-04-19
Enable retry on optional tasks

andrea11:
Hello, 
after the discussion #1060 about the possibility to enable retry in case of optional tasks, I did this PR.
Let me know if I missed something, or if you expect some refactoring.

Thank you

PR #3824, opened 2023-10-26
Upgrade Conductor to Spring3

LuisLainez:
Pull Request type
----
- [ ] Bugfix
- [ ] Feature
- [X ] Build related changes (Please run `./gradlew generateLock saveLock` to refresh dependencies)
- [ ] WHOSUSING.md

Changes in this PR
Upgrading Conductor to Spring 3 - **WARNING - Server doesn't start successfully.**
The main changes are:

1. Upgrade Spring version to 3.1.4
2. Add org.elasticsearch.client is not in the bom - has to be self managed
3. Groovy + Spock - Version 4.0.9 (changed from org.codehaus to org.apache )
4. Add com.google.googlejavaformat as it was not in bom
5. WorkflowDefConstraintTest -> Need to delete the ApacheValidator because bval-jsr does not allow the new jakarta stuff - and version 3x is not yet public.
6. There is an issue with grpc libraries - they generate the ‘Generated’ tag in javax - and then it can’t compile. As per the thread here [protobuf-maven-plugin supports jakarta instead of deprecated javax.annotation · Issue #1113 · grpc/grpc.io](https://github.com/grpc/grpc.io/issues/1113) , we will need to also add the javax annotation - we should not use it inside the project though:  implementation 'javax.annotation:javax.annotation-api:1.3.2'
7. In HTTPTask , the method field used to be HTTP Method - which was an enum and serialized fine, it is no longer an enum and serializes bad - I am decoupling the internal model from the Spring Framework so it doesn’t break constantly.
8. I had to add the dependency [Maven Repository: org.apache.httpcomponents.client5 » httpclient5 » 5.2.1](https://mvnrepository.com/artifact/org.apache.httpcomponents.client5/httpclient5/5.2.1)  because it’s needed in org.springframework.http.client
9. Had to add the explicit jakarta annotations in grpc-client as it depends on grpc and that has the javax as well - just didnt compile.
10. In DefaultRestTemplateProvider , the method setReadTimeout is now deprecated and no longer has an effect - so we had to do changes in there. We are no longer user a restTemplate in the constructor, but a builder, and we set up the timeout in there taking into account the input from the task. **Alert: This will create more RestTemplates potentially in the thread**, but it’s the only way we can set up the timeout PER request. An alternative solution is to create 1 default one and then a custom one per request.
11. Use GRPC version 1.57.2 - Since that works, keeping it 1+ caused compatibility issues with protobuf.

----

Issue #
Spring 2 is reaching its end of life and will no longer be maintained.
Alternatives considered
----


annotations-processor/build.gradle
```diff
@@ -8,8 +8,8 @@ dependencies {
     api 'com.google.guava:guava:31.1-jre'
     api 'com.squareup:javapoet:1.13.+'
     api 'com.github.jknack:handlebars:4.3.+'
-    api 'com.google.protobuf:protobuf-java:3.21.7'
-    api 'javax.annotation:javax.annotation-api:1.3.2'
+    api 'com.google.protobuf:protobuf-java:3.21.12'
+    api 'jakarta.annotation:jakarta.annotation-api:2.1.1'
     api gradleApi()
```
LuisLainez, line 13: Add jakarta annotation. We move from javax -> jakarta

annotations-processor/dependencies.lock
```diff
@@ -1,7 +1,7 @@
 {
     "annotationProcessor": {
         "org.springframework.boot:spring-boot-configuration-processor": {
-            "locked": "2.7.3"
+            "locked": "3.1.4"
```
LuisLainez, line 4: Upgrade Spring to 3.1.4

PR #3836, opened 2023-11-03
Fix double execution of Async system tasks when RepairService is enabled

VerstraeteBert:
EDIT: there's another race condition somewhere that we haven't been able to pinpoint. The reasoning below still holds, but does not fully solve the problem described.

Pull Request type
----
- [x] Bugfix
- [ ] Feature
- [ ] WHOSUSING.md

Changes in this PR
----

Issue Summary:
There's a race condition in the system involving async system tasks and the WorkflowRepairService. For example, when a SUB_WORKFLOW task starts, the WorkflowRepairService sometimes erroneously reinserts the task into the processing queue because it perceives the task as out-of-sync between the ExecutorDAO and the queueDAO. This issue stems from the AsyncSystemTaskExecutor updating a task's status only after it removes it from the queue, creating a window where the WorkflowRepairService can wrongly assess the task state. This leads to duplicate subworkflows/http/… tasks being executed concurrently, which complicates maintaining idempotency of Tasks.

Proposed Solution:
To resolve the issue, it's suggested that the AsyncSystemTaskExecutor should update the status of tasks before removing them from the queue. This should close the window where the WorkflowRepairService can misidentify the task state and prevent unnecessary re-queuing of tasks. An edge case we’ve considered is if the process crashes after the task is updated but before it's removed from the queue. If that happens, the executor will simply remove the task from the queue the next time it runs, thereby not affecting system correctness.

Alternatives considered
----

Making the SubWorkflow and HTTP task sync, which would put more pressure on the decide loop.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/AsyncSystemTaskExecutor.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/AsyncSystemTaskExecutor.java
@@ -110,6 +110,7 @@ public class AsyncSystemTaskExecutor {
         }
 
         boolean hasTaskExecutionCompleted = false;
+        boolean shouldRemoveTaskFromQueue = false;
         String workflowId = task.getWorkflowInstanceId();
         // if we are here the Task object is updated and needs to be persisted regardless of an
         // exception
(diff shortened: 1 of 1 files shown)


v1r3n (approved):

dcore94:
Might this also be solving [this](https://github.com/Netflix/conductor-community/issues/282 )?

PR #3847, opened 2023-11-15
Adding Redis db index and username properties

Anupchat:
Pull Request type
----
Feature

Changes in this PR
----
This adds properties for database index and username to the redis properties for standalone and sentinel.

Issue/Discussion
----
https://github.com/Netflix/conductor/discussions/3713

### DIFF
--- a/redis-persistence/src/main/java/com/netflix/conductor/redis/config/RedisSentinelConfiguration.java
+++ b/redis-persistence/src/main/java/com/netflix/conductor/redis/config/RedisSentinelConfiguration.java
@@ -63,7 +63,24 @@ public class RedisSentinelConfiguration extends JedisCommandsConfigurer {
         }
         // We use the password of the first sentinel host as password and sentinelPassword
         String password = getPassword(hostSupplier.getHosts());
-        if (password != null) {
+        if (properties.getUsername() != null && password != null) {
+            return new JedisSentinel(
+                    new JedisSentinelPool(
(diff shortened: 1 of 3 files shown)


redis-persistence/src/main/java/com/netflix/conductor/redis/config/RedisSentinelConfiguration.java
```diff
@@ -63,16 +63,33 @@ protected JedisCommands createJedisCommands(
         }
         // We use the password of the first sentinel host as password and sentinelPassword
         String password = getPassword(hostSupplier.getHosts());
-        if (password != null) {
+        if (properties.getUsername() != null && password != null) {
```
manish-panwar, line 66: Shall we add null + empty check both? I was also thinking if we should also throw error if password is null/empty, basically make password a mandatory field regardless and rather do that check right after line 65. I can't think of a reason why someone would run Redis without password (and if we must support it, one can look at anonymous flag)... Lets see what community says about this.

Anupchat: The password is optional for various persistence layers that conductor supports, and is the reason why it is obtained from the Host properties. Redis is one of the persistence options and even with Redis it doesn't mandate the use of password. Null check is what existed earlier we can add empty check in addition if needed.

v1r3n (approved):

Issue #3876, opened 2023-12-02
WAIT Task inside DO_WHILE causing infinite task creation which are already completed


appunni-m:
**Describe the bug**
While running the below workflow it goes into infinite loop

**Details**
Conductor version: 3.15.0
Persistence implementation: Postgres and MySQL
Queue implementation:  MySQL and Postgres
Lock: Redis  
Workflow definition:
```
{
  "createTime": 1701489520469,
  "createdBy": "owner@email.com",
  "updatedBy": "owner@email.com",
  "accessPolicy": {},
  "name": "test_do_while",
  "description": "Workflow details",
  "version": 1,
  "tasks": [
    {
      "name": "default__do_while",
      "taskReferenceName": "task_1__loop_databricks",
      "inputParameters": {},
      "type": "DO_WHILE",
      "startDelay": 0,
      "optional": false,
      "asyncComplete": false,
      "loopCondition": "if ($.task_1__loop_databricks['iteration'] < 200) { true; } else { false; }",
      "loopOver": [
        {
          "name": "default__sleep",
          "taskReferenceName": "task_1__wait_databricks",
          "inputParameters": {
            "duration": "20 seconds",
            "tenantId": "csit"
          },
          "type": "WAIT",
          "startDelay": 0,
          "optional": false,
          "asyncComplete": false
        }
      ]
    }
  ],
  "inputParameters": [],
  "outputParameters": {},
  "schemaVersion": 2,
  "restartable": true,
  "workflowStatusListenerEnabled": false,
  "ownerEmail": "owner@email.com",
  "timeoutPolicy": "ALERT_ONLY",
  "timeoutSeconds": 0,
  "variables": {},
  "inputTemplate": {}
}
```

Error in conductor server
```
conductor-server          | 2023-12-02 04:56:48.224 ERROR 13 --- [m-task-worker-8] c.n.c.c.e.AsyncSystemTaskExecutor        : TaskId: 94c1d30a-aef6-4861-be18-4fbfcd03743c could not be found while executing WAIT
conductor-server          | 2023-12-02 04:56:48.227 ERROR 13 --- [-task-worker-11] c.n.c.c.e.AsyncSystemTaskExecutor        : TaskId: 8c683b28-0c10-42dd-894b-2aebead3e3e8 could not be found while executing WAIT
[...]

PR #3878, opened 2023-12-03
Fix: #3876 Update task utils: removeIterationFromTaskRefName

appunni-m:
Parsing name is not considering task ref name with double underscores

- This is not fully fixing the use of this function. There needs to be some kind of validation against the user from setting up the taskRefName with double underscore

Pull Request type
----
- [x] Bugfix
- [ ] Feature
- [ ] WHOSUSING.md

Changes in this PR
----

Issue #3876 
If taskRefName  with double underscore is not considered, then task creation would go into infinite loop. This can be used also as an attack vector on conductor installation as this causes DDos

Alternatives considered
----
Add validation in the create start workflow / create workflow definition api

### DIFF
--- a/common/src/main/java/com/netflix/conductor/common/utils/TaskUtils.java
+++ b/common/src/main/java/com/netflix/conductor/common/utils/TaskUtils.java
@@ -26,6 +26,10 @@ public class TaskUtils {
 
     public static String removeIterationFromTaskRefName(String referenceTaskName) {
         String[] tokens = referenceTaskName.split(TaskUtils.LOOP_TASK_DELIMITER);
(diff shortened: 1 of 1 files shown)


v1r3n:
Hi @appunni-m can you fix the build?

appunni-m:
@v1r3n hey I have fixed the build.

manan164:
Hi @appunni-m , Is there a particular reason that you are using double underscore in the task reference name?

appunni-m:
@manan164 i take it as input from user, thus I don't want to take chances with it. There are no validations against it either.

appunni-m:
@manan164 can we merge this ? Any blocker ?

PR #3884, opened 2023-12-06
Resolve open vulnerabilities in Conductor

scottcarter87:
Pull Request type
----
[*] Build related changes (Please run ./gradlew generateLock saveLock to refresh dependencies)

Changes in this PR
----

Updates to resolve the following CVE's in Conductor

Screenshot 2023-12-06 at 1 11 22 PM

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -29,7 +29,7 @@ ext {
     revEurekaClient = '1.10.10'
     revFasterXml = '2.15.3'
     revGroovy = '3.0.19'
-    revGrpc = '1.57.+'
+    revGrpc = '1.59.1'
     revGuava = '32.1.2-jre'
     revHamcrestAllMatchers = '1.8'
     revHealth = '1.1.+'
--- a/grpc-client/dependencies.lock
+++ b/grpc-client/dependencies.lock
@@ -51,13 +51,13 @@
             "project": true
         },
         "io.grpc:grpc-netty": {
-            "locked": "1.57.2"
+            "locked": "1.59.1"
         },
         "io.grpc:grpc-protobuf": {
-            "locked": "1.57.2"
+            "locked": "1.59.1"
         },
         "io.grpc:grpc-stub": {
-            "locked": "1.57.2"
+            "locked": "1.59.1"
         },
         "org.apache.commons:commons-lang3": {
             "locked": "3.12.0"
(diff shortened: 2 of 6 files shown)


annotations-processor/dependencies.lock
```diff
@@ -1,42 +1,42 @@
 {
     "annotationProcessor": {
         "org.springframework.boot:spring-boot-configuration-processor": {
-            "locked": "2.7.16"
+            "locked": "2.7.18"
         }
```
LuisLainez, line 5: Great job on these btw - but I have already been targetting  these in https://github.com/Netflix/conductor/pull/3855 
As well as the entire upgrade to Spring 3.
We had to have a fork of Netflix Conductor - with Spring 3 ( literally the PR above - copy paste ) - and we are already using that in prod with no problem at all.
I think we should probably push for the change that will give us more bang for our buck, and will move Conductor further in the security space (as Spring 2 is EOL and new versions of libraries are no longer compatible with Spring 2)

scottcarter87: Thank you!  Im definitely interested in those changes as well.  My usage of Conductor requires that we maintain 0 CVEs at all times to remain compliant with very stringent security standards where we deploy it in production.  For that reason I have had to make these fixes in our local mirror of the Conductor codebase and in the interest of trying to not diverge from the main branch I opened this PR.  I’ll leave it up to you guys to decide what to do with it.

LuisLainez: I can't agree more with you - as we have the same requirements.
On saying that - Spring 2 has announced their end of life on the 24th of November, many libraries have started causing problems with newer versions ( all javax vs jakarta, etc ) - which means that we will have to migrate regardless, and that new versions with vulnerabilities will not be able to patched - making conductor unsafe software - in less than 1 year.

scottcarter87: We’re also in the process of moving everything over to Spring 3 for the same reason.  One of the security requirements for us is we cannot use any EOL libraries.  This also meant we had to completely delete the elastic search module from our mirror and turn off ES indexing.  Even having the jar packaged but not used is unacceptable for some reason.  The 6.x elasticsearch versions were EOL a year ago at this point.  So really this PR was just a stop gap to allow us to continue to run Conductor in production.  You have my full support for the Spring 3 work you have started on so thanks for that.

meggarr: IMO, we can make Conductor 3.15.x live on SpringBoot 2.x for a period of time, while making 3.16.x for SpringBoot 3..

Per my understanding, the SB 3 has breakage changes, so it need some time to stabilize the changes.

wildMythicWest (approved):
LGTM

annotations-processor/dependencies.lock
LuisLainez: @meggarr Have you identified any issues with the changes in https://github.com/Netflix/conductor/pull/3855 ? If so, what are they and what is the plan to tackle them?

LuisLainez: @meggarr  The PR has been open for some weeks now and it passes all tests - we have also had to fork from this branch an add to Atlassian services, because of vulnerabilities that require fixing (like Spring2 End Of Life) - and it's working fine in production for us. IMO I don't think we'll get a lot more signal on things that may not go as we expect until we release.

